PR #35, opened 2017-06-07
windows compatibility

sebasv:
On windows machines, the file mode is less relevant than on unix machines. On top of that the `std::os::unix::fs::PermissionsExt` crate has no windows alternative as of this moment, so the highlighting of executable files is disabled with a conditional compilation directive that verifies the target os.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,6 +12,7 @@ use std::error::Error;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::Write;
+#[cfg(target_os = "linux")]
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, Component};
 use std::process;
@@ -77,12 +78,16 @@ fn print_entry(base: &Path, entry: &Path, config: &FdOptions) {
         None    => return
     };
 
+    #[cfg(target_os = "linux")]
     let is_executable = |p: &std::path::PathBuf| {
         p.metadata()
          .ok()
          .map(|f| f.permissions().mode() & 0o111 != 0)
          .unwrap_or(false)
     };
+
+    #[cfg(not(target_os = "linux"))]
+    let is_executable =  |p: &std::path::PathBuf| {false};
 
     if let Some(ref ls_colors) = config.ls_colors {
         let mut component_path = base.to_path_buf();
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,7 @@ use std::error::Error;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::Write;
-#[cfg(target_os = "linux")]
+#[cfg(target_family = "unix")]
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, Component};
 use std::process;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -12,6 +12,7 @@ use std::error::Error;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::Write;
+#[cfg(target_os = "linux")]
```
sharkdp: Will this still work for other Unix-like operating systems (OSX, ..)?

sharkdp:
Thank you very much!

I'm fine with the dummy-behavior for `isExecutable` on windows.

src/main.rs
BatmanAoD: [Doesn't look like it](https://stackoverflow.com/a/41743950/1858225); this should probably be something like `target_family = "unix"`.

sharkdp: Yes, seems like we can use
```rust
#[cfg(unix)]
```
and
```
#[cfg(not(unix))]
```

BatmanAoD:
@Boscop As a point of comparison, it appears that Git-Bash treats `.exe` and `.dll` files as executable, but *not* `.bat` files. But a Windows system on which `.bat` files aren't executable is essentially broken.

To my mind, special-case logic trying to guess how file extensions should be treated is almost certainly the wrong approach. It is a can of worms that gets increasingly complicated; for instance, script-language files (`.py`, `.pl`, `.rb`, etc) may or may not be executable depending on how file extension handling is set up, which depends on registry settings. My opinion (for what it's worth) is that it's better to be simple and consistent.

sebasv:
I second @BatmanAoD , executability is determined by file-wise global settings on windows but they are user-specific. Trying to infer executability based on extension is confusing, and I don't think there currently is a clear and simple method to extract this information from the registry.

sharkdp:
In the end, `is_executable` only decides about the output color, so it's probably not too bad if there are some executable files that are not highlighted.

I'm okay with either of:
* always return false => nothing get's highlighted
* check the extension for `exe` and `bat` and only highlight those


@sebasv What do you think about the `#[cfg(unix)]` change? Do you want to change that?

sebasv:
@sharkdp `#[cfg(unix)]` is indeed more appropriate since the crate in question seems to target all unix systems. I initially chose for `#[cfg(target_os="linux")]` out of ignorance, I didn't know of the `target_family` option until now!

src/main.rs
```diff
@@ -77,13 +78,17 @@ fn print_entry(base: &Path, entry: &Path, config: &FdOptions) {
         None    => return
     };
 
+    #[cfg(target_os = "unix")]
```
sharkdp: It should be `target_family` here, too

src/main.rs
```diff
@@ -77,13 +78,17 @@ fn print_entry(base: &Path, entry: &Path, config: &FdOptions) {
...
     let is_executable = |p: &std::path::PathBuf| {
         p.metadata()
          .ok()
          .map(|f| f.permissions().mode() & 0o111 != 0)
          .unwrap_or(false)
     };
 
+    #[cfg(not(target_os = "unix"))]
```
sharkdp: It should be `target_family` here, too

src/main.rs
sebasv: Oops, well spotted. I appreciate the keen eye!

sharkdp:
Thank you very much!

BatmanAoD:
Awesome! Can't wait to try this out now!

PR #50, opened 2017-07-19
Add --print0 option

michaelmior:
This adds `--print0` and `-0` as possible options which do the same thing as `find -print0`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -53,6 +53,9 @@ struct FdOptions {
     /// Whether to follow symlinks or not.
     follow_links: bool,
 
+    /// Whether elements of output should be separated by a null character
+    zero: bool,
+
     /// The maximum search depth, or `None` if no maximum search depth should be set.
     ///
     /// A depth of `1` includes all files under the current directory, a depth of `2` also includes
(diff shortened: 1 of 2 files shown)


michaelmior:
Thanks :)

src/main.rs
```diff
@@ -53,6 +53,9 @@ struct FdOptions {
     /// Whether to follow symlinks or not.
     follow_links: bool,
 
+    /// Whether elements of output should be separated by a null character
+    zero: bool,
```
sharkdp: I find the name `zero` a little bit too generic here. What do you think about just naming this `null_separator` or `print0`? Another option would be to call this `output_separator` and use an enum `OutputSeparator { Newline, NullCharacter }`

tests/test.sh
```diff
@@ -32,7 +32,7 @@ expect() {
 
     echo "$expected_output" > "$tmp_expected"
 
-    "$fd" "$@" | sort -f > "$tmp_output"
+    "$fd" "$@" | xargs -0 echo -n | sort -f > "$tmp_output"
```
sharkdp: Instead of modifying `expect()` here, couldn't we somehow check for the string `"result1\0result2\0result3"` below?

michaelmior: That does work. The reason I didn't do that is that including `\0` causes git to treat `test.sh` as binary and then it's no longer diffable. Perhaps there's another way to resolve this, but using `xargs` seemed like the easiest solution.

src/main.rs
michaelmior: I think `null_separator` sounds fine.

tests/test.sh
sharkdp: Oh, I see. I thought about somehow adding the two characters `\0` in the shell script (not the null-character itself), like in `echo -e "a\0b"`.

sharkdp:
Do you want to incorporate these changes? Alternatively, I can merge this and update it myself?

michaelmior:
I was working on the changes, but didn't get the chance to make much progress. If you're willing to finish these, that would be great!

sharkdp:
Ok, thanks again!

I see why you chose to add `xargs` now. Unfortunately, it did not work for me because the `sort` would be useless for strings on a single line. Replaced by a `sed`-substitution now.

michaelmior:
Thanks for taking the time to clean things up :)

PR #53, opened 2017-09-15
Highlight matches

sharkdp:
First working prototype with underlining to highlight the matched part of the file path.

image

TODO:
* Only works for `fd -p`, so far.
* The code is horrible.
* Doesn't seem to work on Rust 1.16 (but could be fixed easily)

I'm not sure that I like this better than the non-highlighted version.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -112,7 +112,11 @@ static ROOT_DIR : &'static str = "/";
 static PARENT_DIR : &'static str = "..";
 
 /// Print a search result to the console.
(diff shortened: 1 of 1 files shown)


sharkdp:
I'm going to close this, as I don't really like how (a) the output looks like and (b) the code looks like :smile:

epitron:
This patch is also missing the ability to highlight multiple matched segments (eg: matching "a" in "foo b**a**r b**a**z").

If you don't like how the underline looks, another option would be to change the background colour -- that can also be done independently of the foreground colour...

Issue #54, opened 2017-09-15
Filter on files or directories


ramn:
I'd like a new option ```--type d``` for matching only directories and ```--type f``` for matching only files.

sharkdp:
Absolutely! This was on my TODO list, but I haven't created an issue, yet.

sharkdp:
Closed by @exitium via #55 (Thanks!)

PR #55, opened 2017-09-17
Add file type filtering

exitium:
Add file type filtering as mentioned in #54.


sharkdp:
Very cool!

Thank you very much (especially for also thinking about the `README` and the tests)!

Issue #56, opened 2017-09-29
Add option to only search file extensions?


sharkdp:
For example:
``` bash
> fd -e rs
src/fshelper/mod.rs
src/lscolors/mod.rs
src/main.rs
```

Right now, this is only possible by searching for `\.rs$`.

jszym:
I could take this :)

sharkdp:
> I could take this :)

Thank you very much, but there are already two open PR's :smile: 

If you are interested in working on a different feature, let me know.

jszym:
noob is noob, thanks for the heads up!

sharkdp:
Fixed via #58

sharkdp:
@deg4uss3r @jszym - If you are still interested, there are a few new open tickets now: https://github.com/sharkdp/fd/issues?q=is%3Aopen+is%3Aissue+label%3AHacktoberfest

PR #58, opened 2017-09-29
Add extension filtering

reima:
Add extension filtering as suggested in #56.

I went for the following behavior:
* File extension comparison is always case insensitive.
* Leading dots from the argument are ignored. So both `-e rs` and `-e .rs` have the same effect.

Let me know what you think!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -93,6 +93,11 @@ struct FdOptions {
 
     /// The type of file to search for. All files other than the specified type will be ignored.
     file_type: FileType,
+
+    /// The extension to search for. Only entries matching the extension will be included.
+    ///
+    /// The value (if present) will be a lowercase string without leading dots.
+    extension: Option<String>,
 }
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -301,6 +306,17 @@ fn scan(root: &Path, pattern: Arc<Regex>, base: &Path, config: Arc<FdOptions>) {
                 },
             }
 
+            // Filter out unwanted extensions.
+            match (&config.extension, entry.path().extension()) {
```
exitium: `entry.path().extension()` dynamically extracts the extension. I imagine it would be better to match/map on `config.extension` alone first, so as to avoid calling it on searches that don't need it.

sharkdp:
@reima That looks great, thank you very much! Unfortunately, I won't have time to properly review this before Tue/Wed. I'll get back to you

sharkdp:
> Leading dots from the argument are ignored. So both -e rs and -e .rs have the same effect.

Sounds great!

> File extension comparison is always case insensitive.

It might seem strange that something like `fd --case-sensitive -e TXT` performs a case-*in*sensitive search, but I see your point. And I think you are right in assuming that file-extension searches are probably mostly meant to be case-insensitive.

src/main.rs
sharkdp: Yes, I also think this would be better.

tests/test.sh
```diff
@@ -238,5 +238,17 @@ one/two/three
 one/two/three/directory_foo" --type d
 expect "symlink" --type s
 
+
+suite "File extension (--ext)"
```
sharkdp: Should this be `--extension`?

reima:
Thanks the review! I've integrated the changes you requested.

reima:
Thanks for the review! I've integrated the changes you requested.

sharkdp:
Well, thank you very much again. This is great :+1:

PR #59, opened 2017-09-30
Add flag to force color

### LINKED ISSUE #49: flag to force color
Thanks for a great plugin!
`fd` doesn't use color when piping, however it can be useful. My usecase is for use with [lotabout/skim](https://github.com/lotabout/skim).
ripgrep provides this behaviour with: `rg --color=always`

Detegr:
Color can now be forced on even when piping the output. Changed the --no-color flag to --color which uses the same options that ripgrep uses for --color (excluding ansi).

Closes #49.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -376,10 +376,12 @@ fn main() {
                         .long("absolute-path")
                         .short("a")
                         .help("Show absolute instead of relative paths"))
-            .arg(Arg::with_name("no-color")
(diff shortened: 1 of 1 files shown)


sharkdp:
Looks great! I will take a closer look tomorrow.

sharkdp:
Thank you very much, works as expected!

This will go into the next major version, as it is (technically) a breaking change.

PR #61, opened 2017-10-02
File extension feature

deg4uss3r:
Another option to add a file extension filter for issue #58


sharkdp:
@deg4uss3r Thank you very much for your contribution!

As compared to #58, I think this solution has a few drawbacks:

* Since the file-ending is simply attached to the search pattern, `fd -e txt foo` searches for `foo\.txt$`. This means that `barfoo.txt` will be found, but `foobar.txt` will not be found. I would expect both files to be found.
* The input string (the argument to `-e`) is not escaped, so it will be treated as a regex - not as a string.

Do you see any disadvantages in the implementation in #58?

deg4uss3r:
#58  looks great! Wish I had a little more experience with the codebase before I attempted #56. Let me know of any other feature requests!

sharkdp:
> #58 looks great! Wish I had a little more experience with the codebase before I attempted #56.

Okay, I'm going to close this in favor of #58. In any case, thank you very much for your contribution.

> Let me know of any other feature requests!

I will!

PR #62, opened 2017-10-03
Add a comment about the default value of --threads

Detegr:
I was wondering what the default value was and thought it would make sense to add it in the help description.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -419,7 +419,8 @@ fn main() {
                         .long("threads")
                         .short("j")
                         .takes_value(true)
-                        .help("Set number of threads to use for searching"))
+                        .help("Set number of threads to use for searching\n\
+                               (default: number of available CPU cores)"))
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -419,7 +419,8 @@ fn main() {
                         .long("threads")
                         .short("j")
                         .takes_value(true)
-                        .help("Set number of threads to use for searching"))
+                        .help("Set number of threads to use for searching\n\
+                               (default: The number of CPU cores on the system)"))
```
sharkdp: Minor nitpick: I'd like to keep the width of the help text as small as possible. What do you think about `(default: number of available CPU cores)`?

sharkdp:
Good suggestion, thank you!

src/main.rs
Detegr: That's perfect, I'll update the PR.

sharkdp:
Thanks!

Issue #64, opened 2017-10-04
Add shell completion support


sharkdp:
clap is able to [automatically produce shell-completion files](https://docs.rs/clap/2.26.2/clap/struct.App.html#method.gen_completions). It would be great if we could easily generate them and ship them with each `fd` release.

sharkdp:
This has been solved by @ImbaKnugel in #66 (Thanks!).

Leaving this open until I figure out the best way to ship these with each release.

sharkdp:
The PowerShell completions are already included in the Windows tarballs. Closing this in favor of #76.

PR #66, opened 2017-10-04
Generate shell completions.

ImbaKnugel:
Adds the generation of shell completion at compile time using the clap feature mentioned in  #64.

* Moves construction of App struct building to seperate file.
* Adds build.rs to generate the complection files.


sharkdp:
Thank you very much for your contribution, this is great!

Issue #65, opened 2017-10-04
Re-write integration tests in Rust


sharkdp:
If there is a nice way to write the current integration tests (`tests/test.sh`) in Rust, this would be really beneficial:

* the tests would be platform-independent. Currently, the integration tests (naturally) do not run on Windows.
* `cargo test` would run both unit tests and integration tests

reima:
Had the same thought yesterday and just wanted to open an issue for it 😃 

I'll take a stab at this and will report back with a prototype later today.

sharkdp:
Closed via #67

PR #67, opened 2017-10-04
Re-write integration tests in Rust

reima:
This is a proposal for fixing #65.

Things to note:
* The `TestEnv` struct is responsible for creating (and cleaning up) the temporary working directory, filling it with the test files, and running the `fd` executable.
* As the `TestEnv` struct is recreated in each test method, so is the temporary working directory.
* When a test fails, the command line arguments and the output diff are printed (as the test script did).
* Path separators are normalized to the default platform separator before comparison (`testenv::normalize_output`).
* The tests for absolute path and explicit root path do *not* pass on Windows. I've disabled them on Windows for the time being.

I'm looking forward to your thoughts!


reima:
Thank you for your kind words. I've addressed your inline comments and synced with the master branch to resolve the merge conflicts.

Leaving the temporary directory can definitely come in handy in debugging. One could use `TempDir::into_path` to destroy the `TempDir` without deleting the directory, store the resulting `PathBuf` in `TestEnv` and conditionally do the cleanup manually in a `impl Drop for TestEnv`. But I agree that this would warrant another issue/PR.

sharkdp:
Again, thanks!

PR #68, opened 2017-10-04
Add Examples to README redux

deg4uss3r:


sharkdp:
Sorry to be a pain, but now there is still an unrelated change in the README file that should not be included when merging this. Thanks for your patience.

deg4uss3r:
No problem, new commit fixes that weird issues, it must have been a new line/carrage return issue?

sharkdp:
I have thought about this a bit more, and I would like to change a few of the examples (after merging this). In particular, I would like to try to find real-world scenarios for most of the examples (as compared to searches in a demo environment).

If you are still interested in helping, you are more than welcome. In any case, thank you very much!!

deg4uss3r:
@sharkdp No problem! I think the piping to `xargs` is a great example, I did just today in fact. I'll keep my eye out for the real world uses I use fd for :) Thank you for the introduction, I'm also looking forward to writing some rust code for it too ;)

Issue #63, opened 2017-10-04
Provide more detailed '--help' output


sharkdp:
clap has an option to show a more detailed help message if `fd --help` is called (as compared to `fd -h`):

* https://docs.rs/clap/2.26.2/clap/struct.Arg.html#method.long_help
* https://docs.rs/clap/2.26.2/clap/struct.App.html#method.long_about

We could use this to provide a more detailed help text for `fd`.

abaez:
Hey, 

I would like to take a stab at this. Probably get this one done later tonight.

sharkdp:
Closed via #71

PR #71, opened 2017-10-05
initial addition of long_help

abaez:
As requested for #63, here's the initial portion of [long_help][1] for the `--help` flag.

I've only done the first three, since I want to know what you think of the formatting before I continue doing all the rest. I tend to split up clap's expansion help to their own variable for easier readability in argument definitions. 

[1]: https://docs.rs/clap/2.26.2/clap/struct.Arg.html#method.long_help

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,110 +1,150 @@
+use std::collections::HashMap;
+
 use clap::{App, AppSettings, Arg};
 
+struct Help {
+    short: &'static str,
+    long: &'static str,
+}
+
+macro_rules! doc {
+    ($map:expr, $name:expr, $short:expr) => {
+        doc!($map, $name, $short, $short)
+    };
+    ($map:expr, $name:expr, $short:expr, $long:expr) => {
+        $map.insert($name, Help {
+            short: $short,
+            long: concat!($long, "\n ")
+        });
+    };
+}
+
 pub fn build_app() -> App<'static, 'static> {
+    let helps = usage();
+    let arg = |name| {
+        Arg::with_name(name).help(helps[name].short).long_help(
+            helps[name].long,
+        )
+    };
+
     App::new("fd")
         .version(crate_version!())
         .usage("fd [FLAGS/OPTIONS] [<pattern>] [<path>]")
         .setting(AppSettings::ColoredHelp)
         .setting(AppSettings::DeriveDisplayOrder)
+        .arg(arg("hidden").long("hidden").short("H"))
+        .arg(arg("no-ignore").long("no-ignore").short("I"))
+        .arg(arg("case-sensitive").long("case-sensitive").short("s"))
+        .arg(arg("absolute-path").long("absolute-path").short("a"))
+        .arg(arg("follow").long("follow").short("L").alias("dereference"))
+        .arg(arg("full-path").long("full-path").short("p"))
+        .arg(arg("null_separator").long("print0").short("0"))
+        .arg(arg("depth").long("max-depth").short("d").takes_value(true))
         .arg(
-            Arg::with_name("hidden")
-                .long("hidden")
-                .short("H")
-                .help("Search hidden files and directories"),
-        )
-        .arg(
-            Arg::with_name("no-ignore")
-                .long("no-ignore")
-                .short("I")
-                .help("Do not respect .(git)ignore files"),
-        )
-        .arg(
-            Arg::with_name("case-sensitive")
-                .long("case-sensitive")
-                .short("s")
-                .help("Case-sensitive search (default: smart case)"),
-        )
-        .arg(
-            Arg::with_name("absolute-path")
-                .long("absolute-path")
-                .short("a")
-                .help("Show absolute instead of relative paths"),
-        )
-        .arg(
-            Arg::with_name("follow")
-                .long("follow")
-                .short("L")
-                .alias("dereference")
-                .help("Follow symbolic links"),
-        )
-        .arg(
-            Arg::with_name("full-path")
-                .long("full-path")
-                .short("p")
-                .help("Search full path (default: file-/dirname only)"),
-        )
-        .arg(
-            Arg::with_name("null_separator")
-                .long("print0")
-                .short("0")
-                .help("Separate results by the null character"),
-        )
-        .arg(
-            Arg::with_name("depth")
-                .long("max-depth")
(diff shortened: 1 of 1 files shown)


sharkdp:

A couple of comments:
1. I really appreciate what you did with the examples, but I think this is already *too* detailed for these long help messages. In my opinion, these examples would be a great fit for an "Examples" section in a man page (or maybe in the `--help` text appendix).
2. Given that the help texts would be a little bit shorter when excluding the examples, I think I would prefer them to be inline (at the place where the `Arg` is defined)... but I don't have a strong opinion about this. Another option would be to do something [similar to ripgrep](https://github.com/BurntSushi/ripgrep/blob/214f2bef666efd686b1be250fc8e9f54a2cebb0f/src/app.rs#L227)
3. The build is currently failing.. should be easy to fix.

abaez:
Hey,

I took the time to go about and implement the ripgrep style as the option was available. Any issues with formatting now, please let me know. Before accepting this pull request, please inform me, as I do want to squash it before acceptance.

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+        .arg(arg("pattern"))
+        .arg(arg("path"))
+}
+
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn usage() -> HashMap<&'static str, Help> {
+    let mut h = HashMap::new();
+    doc!(h, "hidden"
+        , "Search hidden files and directories"
+        , "Includes, in a search query, any hidden files and directories for the directory of \
```
sharkdp: Nitpick: the hidden files and directories don't really belong to the search query, so I would rather phrase this as: "Also include hidden directories and files in the search results (default: hidden files and directories are skipped)"

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+            the search.");
+    doc!(h, "no-ignore"
+        , "Do not respect .(git)ignore files"
+        , "Searches for a pattern match with files and directories, including all defined \
```
sharkdp: Same as above: "Show search results from files and directories that would otherwise be ignored by '.*ignore' files."

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+    doc!(h, "case-sensitive"
+        , "Case-sensitive search (default: smart case)"
+        , "Define your pattern as case sensitive. By default, fd uses 'smart case' for queries.\
+            This option disables smart case.");
+    doc!(h, "absolute-path"
+        , "Show absolute instead of relative paths"
+        , "Shows the result of a pattern match as an absolute path instead of relative path.");
+    doc!(h, "follow"
+        , "Follow symbolic links"
+        , "Includes a symbolic links destination in the pattern match.");
```
sharkdp: Actually, `--follow` has slightly different semantics: "By default, fd does not descent into symlinked directories. Using this flag, symbolic links are also traversed."

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+    doc!(h, "full-path"
+        , "Search full path (default: file-/dirname only)"
+        , "Searches the full path of directory. By default, fd only navigates file-/dirname.");
```
sharkdp: It's rather: "[..] By default, fd only searches the last part of the path (the file name or the directory name)"

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+    doc!(h, "null_separator"
+        , "Separate results by the null character"
+        , "Separates each pattern match result by the null character.");
```
sharkdp: We could refer to `xargs` here: "Separate search results by the null character (instead of newlines). This is useful for piping results to 'xargs'".

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+    doc!(h, "depth"
+        , "Set maximum search depth (default: none)"
+        , "Set the limit of the maximum search depth in a pattern match query. By default, \
```
sharkdp: Maybe: "By default, there is no limit on the search depth"

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+            none is the max-depth.");
+    doc!(h, "file-type"
+        , "Filter by type: f(ile), d(irectory), s(ymlink)"
+        , "Filter the search by type:\n\
+            f file          for file\n\
+            d directory     for directory\n\
+            s symlink       for symbolic links.");
+    doc!(h, "extension"
+        , "Filter by file extension"
+        , "Filter the search to only match for a filetype extension");
```
sharkdp: `--extension` can be used in addition to a normal search (for example: `fd -e png IMG_`). So I would rather say something like "Only show search results with a specific file extension".

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+        , "Declare when to use color for the pattern match output:\n\
+            auto (default)\n\
+            never\n\
+            always");
+    doc!(h, "threads"
+        , "Set number of threads to use for searching:\n\
+            (default: number of available CPU cores)");
+    doc!(h, "max-buffer-time"
+        , "the time (in ms) to buffer, before streaming to the console"
+        , "Amount of time in milliseconds to buffer, before streaming to the console the results");
```
sharkdp: Minor nitpick "before streaming the search results to the console"

src/app.rs
```diff
@@ -1,110 +1,148 @@
...
+    doc!(h, "pattern"
+        , "the search pattern, a regular expression (optional)"
+        , "The search pattern to use for a match. By default, uses string match. Can use a \
```
sharkdp: Actually, the pattern is always treated as a regular expression. I don't think we need an extended help text here.

abaez:
Hey @sharkdp,

I've gone ahead and applied all the changes you requested. Let me know if there's anything else you may want. Otherwise, I believe I should squash this.

sharkdp:
> Otherwise, I believe I should squash this.

I can also "Squash and merge" in the GitHub PR panel, but if you want to to it yourself, that would also be great.

Thank you very much!

abaez:
Hey @sharkdp,
 
I've gone ahead and done it for you 👍. I'm too used to doing it manually when using different services, that it just became second nature not to use host services for rebase actions.

Thanks for the speedy acceptance! I'll look for other issues to spend some time with in your marvelous find alt.

sharkdp:
Thanks!

PR #72, opened 2017-10-06
(Proper) Windows support

sharkdp:
This fixes the failing tests on Windows

- For now, this is just a workaround: I'm stripping the prefix (`\\?\`) from the absolute paths on Windows. 

For future reference: Windows paths with the `\\?\` prefix are called UNC paths. More information here: https://github.com/rust-lang/rust/issues/42869

### DIFF
--- a/src/fshelper/mod.rs
+++ b/src/fshelper/mod.rs
@@ -1,4 +1,5 @@
 use std::path::{Path, PathBuf};
+use std::io;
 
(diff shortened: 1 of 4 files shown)


sharkdp:
For future reference: the weird `\\?\` prefixes on Windows are called UNC paths. More information here: https://github.com/rust-lang/rust/issues/42869

Issue #73, opened 2017-10-07
Use real-world examples in the tutorial


sharkdp:
The tutorial in the README contains the common use cases of `fd`, but it would be nice if we could find real-world examples for each of these use cases (as opposed to searches in a test-environment).

Some random ideas:
* Find specific log files in `/var/log`
* Find configuration files in `/etc`
* Use the *fd*-repository as an actual real world example
    * Use a plain `fd` call to get an overview of all files
    * `fd -e yml` to show all `*.yml` files. Notice that `.travis.yml` is not present => `fd -H -e yml`.
    * `fd -H sample` to show the `.git/hooks/` files.
    * ...
* Find all Python files in the current directory: `fd -e py`.

TitiAlone:
I take this.

sharkdp:
@TitiAlone Sounds great! Let me know if you have anything to share (even single examples).

TitiAlone:
Not much time for now, unfortunately, beacause of events I can't control...
Here's where I stopped, if someone wants to work on it before:

`
## Tutorial

First, to see all command line options, you can get `fd`'s help text by running:
```
fd --help
```

`fd` can be called with no arguments, it'll then show the content of the actual directory
recursively (like `ls -R`):
```
> cd fd
> fd
appveyor.yml
build.rs
Cargo.lock
Cargo.toml
LICENSE
README.md
src
src/app.rs
src/fshelper
src/fshelper/mod.rs
src/lscolors
src/lscolors/mod.rs
src/main.rs
tests
tests/testenv
tests/testenv/mod.rs
tests/tests.rs
```

If we work in a directory that is a Git repository (or includes several Git repositories), `fd`
does not search folders (and does not show files) that match the `.gitignore` pattern. For instance,
see above example, and note that .gitignore is the following:
```
target/
**/*.rs.bk
```

To disable this behavior, we can use the `-I` (or `--ignore`) option:
```
> fd -I
Cargo.lock
Cargo.toml
LICENSE
README.md
appveyor.yml
build.rs
src
src\app.rs
src\fshelper
src\fshelper\mod.rs
src\lscolors
src\lscolors\mod.rs
src\main.rs
target
target\debug
target\debug\build
target\debug\build\fd-find-11fb469ceafee9ce
target\debug\build\fd-find-11fb469ceafee9ce\out
target\debug\build\fd-find-11fb469ceafee9ce\out\_fd
[...]

Issue #36, opened 2017-06-09
Reproducible benchmarks


sharkdp:
For example:
- Generate a lot of dummy files / directories in a controlled way by a shell script and run specific benchmarks within this test folder.
- Clone a certain (big) repository and run benchmarks within this repository (this is not 100% reproducible).

jrobeson:
As per https://github.com/sharkdp/fd/issues/30 you might also want to bench against network  filesystems like NFS and SMB. And also against sshfs and other FUSE filesystems.

lnicola:
One thing I noticed is that `fd` calls `stat` three times for each file.

sharkdp:
> One thing I noticed is that fd calls stat three times (and lstat once) for each file.

Good catch!! This should be fixed on master (via adaf4f6d73aa6935d381a4eb9be647815aeb46cd), i.e. `fd` should perform only one `lstat` and one `stat` call.

Note: those syscalls are only necessary when colored output is active (in order to check if the entries are symlinks, directories, executables or normal files).

lnicola:
I think it does a `lstat` via `walkdir` and one `stat` per path component in `print_entry`.

But I'm not sure that's a good idea. Don't we care only about the leaf component? Presumably, non-leaf ones are directories and we only want to check the type of the file.

sharkdp:
> Don't we care only about the leaf component?

Well, currently `fd` can do this (i.e. colorize "symlink1" as a symlink):

![image](https://user-images.githubusercontent.com/4209276/30557324-309464c0-9cae-11e7-8124-7cf9a79bf6bf.png)

lnicola:
Yeah, that might be useful, I suppose. I wasn't sure what `fd` is supposed to do with symlinks:

```
 grayshade@ubik  ~/foo  ll
total 1.0K
drwxr-xr-x 2 grayshade grayshade 2 Sep 18 22:32 bar
lrwxrwxrwx 1 grayshade grayshade 3 Sep 18 22:32 baz -> bar
 grayshade@ubik  ~/foo  fd . baz
bar
```

sharkdp:
If somebody has any ideas on how to proceed on this, I'd love to hear them!

Here are the two benchmark scripts that I have so far:

https://gist.github.com/sharkdp/4bc3e5f5ea9df2f29c02ede50634b16a

avently:
On my system fd is slower than find. After multiple starts:
fd -HI -c never zsh         //2.3-2.4 sec 
[...]

PR #74, opened 2017-10-07
Implement `-m / --mount` flag

Detegr:
`-mount` flag in `find` prevents `find` from descending into directories that reside on other filesystems. This flag does the same, and prevents `fd` from descending into directories that are mountpoints of other filesystems.

As this implementation relies on `/proc/mounts`, a check is performed when parsing command line arguments to check whether it exists and is readable. If not, the `-m / --mount` option is disabled.

It looks like that MacOS support would be quite easy to implement too, but I'm not sure how often this would be useful on that platform. Also Windows supports mounting filesystems under directories, but those will be symlinks and not followed by `fd` by default, and my guess is that nearly nobody does that anyway.


sharkdp:
Thank you very much for taking the time to implement this!

I'm kind of torn on this. On the one hand, this looks like a useful feature. On the other hand, I'd also like `fd` to stay true to its goals as stated in the first two lines in the README: "a *simple*, fast, and user-friendly alternative to find" and "While it does not seek to mirror all of find's powerful functionality...".

You could argue that there is no downside in adding an additional feature, and I would tend to agree with you.

However, I believe there are a few possible downsides to keep in mind when we agree to add more features like this (and I'm speaking a bit more in general, not necessarily specific to this PR):
1. Possible performance degradation
2. Added complexity in the code and increased maintenance burden.
3. Added complexity in the user interface, specifically, the `-h` and `--help` texts.

Concerning point (1), this is something I would likely not tolerate. This specific PR seems harmless in this respect (however, there is at least one more `if`-clause, even if the feature is deactivated - but that's probably negligible).

Concerning point (2), I'm not yet decided. So far, *fd*s code doesn't seem too complex and we can potentially handle a few more features..

Concerning point (3), I would really like the help text to stay rather concise and I want the most important command line options to be easily visible. We could experiment with showing some options in the short `-h` text and showing all options in the long `--help` text.

Sorry for the long discussion, but I'm also interested in your view concerning these points.

Coming back to this specific feature: do you happen to have a real-world example where this feature is needed (see #73)?

(Side note: I've added a [CONTRIBUTING.md](https://github.com/sharkdp/fd/blob/master/CONTRIBUTING.md) file to the repo that asks potential future contributors to open an issue first before opening a PR. Not that there is any harm in discussing things here in the PR, but I'd like to avoid people spending their free time on things that might not be merged later on)

Detegr:
I get your points and I won't be offended if you decide to reject this PR. I like `fd` and this feature is useful for me, so I would have used the time to implement it any case.

I do agree that this kind of programs need to be as efficient as possible, the if-clause in the hot path is the best I could think of. I think branch prediction will work quite well here, so as you said, the performance hit is probably negligible. Maybe after #36 we could have some hard data on this.

About the code complexity, I agree that the current code is not very complex, though I think it could be more readable/less complex/easier to maintain by splitting the large main file to a smaller components. I think this feature is quite okay in terms of code complexity vs usefulness of `fd`, but I understand if you want to avoid feature creep and reject the feature.


> We could experiment with showing some options in the short -h text and showing all options in the long --help text.

I like this idea, we just need to make it explicit in `-h` help that `--help` includes more features.


> Sorry for the long discussion, but I'm also interested in your view concerning these points.
> Coming back to this specific feature: do you happen to have a real-world example where this feature is needed (see #73)?

Discussion is fine, no worries :)
On my desktop machine I have a couple of backup drives mounted in my home directory in `~/backup` and `~/backup2` and I don't want those to be searched while searching the home directory. However I think that's not the best real-world scenario for the README, simpler one would be to have an imaginary sd-card mounted under the example directory with few files to keep the file list small.

tmccombs:
Given the use cases, I think it would make more sense to add a more general `--exclude` option (and maybe a matching `--include` option). 

As a workaround with the current fd, you could put a .ignore file in your home directory that ignores the backup and backup2 directories.

Detegr:
> Given the use cases, I think it would make more sense to add a more general --exclude option (and maybe a matching --include option).

Now that you brought this up I think this is indeed a nice way to handle this with more added value than just skipping the mountpoints.

sharkdp:
> Now that you brought this up I think this is indeed a nice way to handle this with more added value than just skipping the mountpoints.

If you like this better, it would be great if someone could write up a proposal as a GitHub issue. I'm not exactly sure how this would look like, yet.

sharkdp:
Just for reference - the proposal by @tmccombs: #89 (thanks!)

Detegr:
Doh, I used master branch for this PR and now I used it again for another. I think I'll close this out as we're discussing about better alternatives already.

Issue #80, opened 2017-10-08
Use "--type l" for symlinks, not "--type s"


sharkdp:
.. to be compatible with `find`.

rozuur:
Can I pick this?

sharkdp:
Thank you for considering to help, but there is an open pull request already.

rozuur:
Ahh.. didn't realize it. No issues will check something else.

PR #83, opened 2017-10-09
Switch symlink arg from 's' to 'l'

jcpetkovich:
Switch the symlink argument for `--type` from `s` to `l`. Fixing #80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ FLAGS:
 
 OPTIONS:
     -d, --max-depth <depth>    Set maximum search depth (default: none)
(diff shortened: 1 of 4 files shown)


sharkdp:
Looks good, thank you very much. We should also change the help text in  `src/app.rs` (for both the short `-h` text as well as the long `--help` text).

Deepak20-1997:
1.

Deepak20-1997:
Hi im now in

jcpetkovich:
Whoops, sorry, missed that! Should be good now.

PR #90, opened 2017-10-10
Break main down

### LINKED ISSUE #88: Split up main.rs into separate modules.
A possible split could look like this:

- `main.rs` just includes the `main` function and the command-line parsing / options.
- Split out `print_entry` into its own module (`output.rs`?). Possibly split up the `print_entry` function as well.
- Split out `scan` into its own module (`traversal.rs`, `walk.rs`, or sth. similar). Possibly split out receiver and transmitter code into their own functions.

gsquire:
Fixes #88 

The first commit here breaks main into separate modules. I can take a shot at refactoring after you review what I have so far. I had to make some types public, reorganize imports, etc.

If you have a preference on how to refactor the larger functions, let me know.


sharkdp:
Looks great, thank you very much!

We should consider splitting out `FdOptions`, `PathDisplay`, `ROOT_DIR`, .. into a separate module as well (something like `config.rs`). Otherwise, the new modules have to import them from `main`, which seems strange (circular) to me.

gsquire:
Sure, I can tackle that this evening.

sharkdp:
> If you have a preference on how to refactor the larger functions, let me know.

Actually, I think it would be best to do this in a (potential) second step after this is merged.

gsquire:
I moved the shared types into a module called `internal`. I also squashed my commits.

Let me know if there is anything else you would like adjusted.

sharkdp:
Perfect, thank you very much for your contribution!

Issue #85, opened 2017-10-09
RFC: Use glob pattern by default & Case-sensitive by default, No smart case


iology:
RegExp is quite complicated if users do not pay attention to it.

Say that you want to delete directories named "tests" but forget to add `^` and `$` (`^tests$`), then some other directories may be deleted too.

Smart case just makes more trouble when you copy-paste some file names with upper-case letters, like ".DS_Store". And for documentation purpose I would like to keep the upper-case (in MacOS's case-insensitive HFS+ volume), and thus have to use `(?i)^\.DS_Store$`. `(?i)` is the only way to enable case-sensitivity in this case.

EDIT: fix typos :P

sharkdp:
Thank you for your contribution.

> **Use glob pattern by default**

I'm certainly against it. Part of the reason I created *fd* was not having to type `'*test*'`, when I just want to search for `test` as a substring.

> Say that you want to delete directories named "tests" but forget to add ^ and $ (^tests$), then some other directories may be deleted too.

I would argue like this: I hope that no-one really calls `any_command | xargs rm` without checking the output of `any_command` first.

> **Case-sensitive by default, No smart case**

Also, I'd rather not.

> Smart case just makes more trouble when you copy-paste some file names with upper-case letters, like ".DS_Store". And for documentation purpose I would like to keep the upper-case (in MacOS's case-insensitive HFS+ volume), and thus have to use (?i)^\.DS_Store$. (?i) is the only way to enable case-sensitivity in this case.

Would it help if *fd* had an `--case-insensitive` flag to complement `--case-sensitive`?

I was thinking of adding a `--literal` flag (or `--glob`?) that would maybe also help with this use case.

iology:
Ah yeah, `--case-insensitive` is useful. Also useful to override the previous `--case-sensitive`. `--glob` is better than `--literal`, since shell users should already know glob patterns. Better to have `--regexp` for switching back to regexp for the alias `fd --case-sensitive --glob`.

> I would argue like this: I hope that no-one really calls `any_command | xargs rm` without checking the output of `any_command` first.

That's not enough. There could be exceptions when searching more files. If you want to write a non-interactive script, you should check your regexp more carefully. Another common pitfall is `.ext$` (`\.ext$`).

Anyway, now I see that you really want a good default for casual search.

sharkdp:
@iology Thank you for your feedback.

[...]

PR #95, opened 2017-10-11
Add --case-insensitive (-i)

iology:
ping #85


iology:
Need --smart-case (-S)?

iology:
Rename this flag as --ignore-case?

sharkdp:
Looks great, thank you very much for your contribution!

`-i`/`--ignore-case` is perfect, as it is analogous to `ripgrep`.

I will properly review this later today.

One thing I was wondering is whether we could hide some of the options from the short `-h` text (maybe not from the long `--help` text). I would somehow like to keep the `-h` text short and concise, with only the most often used options.

iology:
Rebased.

sharkdp:
Thank you very much!

PR #96, opened 2017-10-11
Add --glob and --regex

iology:
WIP:

- ~~[ ] update the "globset" crate in the upstream~~?
- [ ] documentation
- [ ] the behavior of "/" and "\\" on Windows file paths (BurntSushi/ripgrep#526)
- [x] test with `--full-path`
- [ ] more tests
- [ ] todos and fixme in the source code

ping #85


src/main.rs
```diff
@@ -123,9 +125,19 @@ fn main() {
...
 
-    match RegexBuilder::new(pattern)
-              .case_insensitive(!config.case_sensitive)
-              .build() {
+    let mut builder = if !matches.is_present("use-glob") {
+        RegexBuilder::new(pattern)
+    } else {
+        match GlobBuilder::new(pattern).literal_separator(true).build() {
+            Ok(glob) => {
+                // TODO: provide an option to remove (?-u) in the upstream?
```
sharkdp: is this still relevant?

src/main.rs
```diff
@@ -123,9 +125,19 @@ fn main() {
...
+                eprintln!("GLOB:{} | REGEX: {}", glob.glob(), glob.regex());
```
sharkdp: Is this a debug print?

iology: Yes, just to get you familiar with how --glob works.

src/main.rs
iology: If it is unnecessary to match arbitrary bytes in the path, then no need to remove (?-u).

iology:
Added tests for --full-path and rebased.

iology:
Another problem is how to apply `(?s)` (allow `.` to match newlines) to glob patterns.

sharkdp:
> Another problem is how to apply (?s) (allow . to match newlines, "dot-all") to glob patterns.

I think we should do the same as for #111.

iology:
Yes, and no option to disable "dot-all" for glob patterns? I'm ok with this.

sharkdp:
> Yes, and no option to disable "dot-all" for glob patterns? I'm ok with this.

Me too.

iology:
I've added some todos in the top comment.

iology:
Closing. I plan to do this in my private fork. Thanks for making `fd`. It really helps to get me familiar with Rust.

PR #100, opened 2017-10-11
Format the source code using `rustfmt`

Detegr:
Just a `rustfmt` pass in this PR. I'll do Travis setup separately. #99

PR #101, opened 2017-10-11
Make Travis check that the code is correctly formatted

### LINKED ISSUE #99: Use rustfmt to format the source code
[Rustfmt](https://github.com/rust-lang-nursery/rustfmt) automatically formats Rust code to conform [Rust style guide](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/guide.md). The current code base of `fd` is hand formatted and I think it would be easier for contributors if we formatted the code with `rustfmt`.

What do you think? If there's some irritating style decisions in `rustfmt`, it is possible to tweak it to some extent.

Detegr:
Stable channel build now runs `rustfmt` and the build is considered as failed if it produces any differences compared to the code that was submitted. The nightly channel runs `rustfmt-nightly` just to see in advance whether there's some changes coming in the nightly version of `rustfmt`. As nightly builds are allowed to fail, it does not affect the result of the build. Closes #99.

Obviously we will want #100 to be merged before this :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,38 @@
 language: rust
-rust:
-  - stable
-  - beta
-  - nightly
+cache: cargo
 matrix:
   allow_failures:
     - rust: nightly
   include:
+    # Stable channel.
+    - os: linux
+      rust: stable
+      before_script:
+          - export PATH="$PATH:$HOME/.cargo/bin"
+          - which rustfmt || cargo install rustfmt
+      script:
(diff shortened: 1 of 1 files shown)


Detegr:
It's getting late, messed up with the titles, sorry about that :)

sharkdp:
This is great, thank you very much.

I hope that we do not shy away future users (whose PRs fail to build) due to this strict enforcement of code style :smile:

PR #102, opened 2017-10-11
Use absolute path to perform search, fixes #81

### LINKED ISSUE #81: Current directory not included when using relative path
```
$ fd --version
fd 4.0.0
```

In directory "a":

```
$ tree -F ..
..
├── a/
│   └── a
├── b/
└── c/
```

```
$ fd . .
a
```

```
$ fd . ..
../b
../c
a
```

In directory "b":

```
$ fd . ..
../a
../a/a
../c
```

sharkdp:
Previously, we were using the relative path to extract the search
string. For the current directory, the returned relative path was equal
to `""`. This is why the current directory did not show in the search
results (see #81).

This commit also changes the way that `--full-path` works, which was
previously working on relative paths. It seems more useful to search the
absolute path, though. Otherwise, search results could change just by
calling fd (with a given search path) from a different directory.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -2,7 +2,7 @@ use internal::{error, FdOptions};
 use fshelper;
 use output;
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::mpsc::channel;
 use std::thread;
@@ -101,12 +101,19 @@ pub fn scan(root: &Path, pattern: Arc<Regex>, base: &Path, config: Arc<FdOptions
         let config = Arc::clone(&config);
         let pattern = Arc::clone(&pattern);
         let tx_thread = tx.clone();
+        let root = root.to_owned();
 
         Box::new(move |entry_o| {
             let entry = match entry_o {
                 Ok(e) => e,
                 Err(_) => return ignore::WalkState::Continue,
             };
+
+            let entry_path = entry.path();
+
+            if entry_path == root {
+                return ignore::WalkState::Continue;
+            }
 
             // Filter out unwanted file types.
             match config.file_type {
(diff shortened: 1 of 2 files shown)


sharkdp:
@iology If you are interested, it would be great if you could review this!

iology:
I'm not familiar with the logic yet and quite a newbie to Rust.

In comparison to `find`, the asymmetry between `fd pattern .` and `fd pattern ..` is confusing.

```
# Tested with bash, in directory b/
$ tree -F ..
..
├── a/
│   └── a-1
└── b -> a/
```

```
$ find .
.
./a-1
```

```
$ find ..
..
../a
../a/a-1
../b
```

```
$ fd "" .
a-1
```

```
$ fd "" ..
.
../b
a-1
```

Problems:

1. "." is not included for `fd "" .` (feature?)
2. "." (instead of "../a") is included for `fd "" ..` (bug)
3. ".." is not included for `fd "" ..` (feature?)
4. "a-1" instead of "../a/a-1" is included for `fd "" ..`
5. "./" is not prepended to "a-1" for `fd "" .` (feature? use `fd` like `ls`?)

iology:
The symlink issue make it more complicated, so I added a normal directory c/ and test again:

```
# In directory c/
$ tree -F ..
..
├── a/
│   └── a-1
├── b -> a/
└── c/
    └── c-1
```

```
$ fd '' .
c-1
```

```
$ fd '' ..
.
../a
../a/a-1
../b
c-1
```

The problems are basically the same.

sharkdp:
Thank you for trying this out!

> "." is not included for fd "" . (feature?)
> ".." is not included for fd "" .. (feature?)

Yes, it's a "feature". I'm hiding the root directory of the search here:
https://github.com/sharkdp/fd/blob/d4313661fe3e62c43d1c3f6c9a9ca7a4f916d027/src/walk.rs#L113-L115

> "./" is not prepended to "a-1" for fd "" . (feature? use fd like ls?)

I don't like the superfluous prefix './' that `find` prints. `ls` also doesn't show this?

> "." (instead of "../a") is included for fd "" .. (bug)
> "a-1" instead of "../a/a-1" is included for fd "" ..

In these cases, aren't "." and "../a" (or "a-1" and "../a/a-1") the exact same path? I agree that it might be more helpful to show "../a", but it's not wrong?

iology:
I also think it is fine to exclude "." and ".." since they are known beforehand.

> >    "./" is not prepended to "a-1" for fd "" . (feature? use fd like ls?)
>
> I don't like the superfluous prefix './' that find prints. ls also doesn't show this?

But a uniform format makes the result tidier.

> > "." (instead of "../a") is included for fd "" .. (bug)
> >
> > "a-1" instead of "../a/a-1" is included for fd "" ..
>
> In these cases, aren't "." and "../a" (or "a-1" and "../a/a-1") the exact same path? I agree that it might be more helpful to show "../a", but it's not wrong?

They have the same real path, but symbolic links and normal directories/files are different.

With "." (instead of "../a") you cannot simply `rmdir ..` or `mv . /new_path`, and etc.

"../b" (symlink) instead of "../a" (directory) is included, so `--type d` cannot find "../a" (bug).

iology:
> But a uniform format makes the result tidier.

To state it clearer (`fd "" ..` in directory a/):

BEFORE:

```
.
../b
../c
../c/c-1
a-1
```

AFTER

```
../a
../a/a-1
../b
../c
../c/c-1
```

sharkdp:
@iology Thank you very much.

I'm going to merge this pull request, as it fixes the weird behavior that you described in #81. At the same time, I'm going to open a new ticket to address the uniform output format.

Issue #29, opened 2017-06-07
Add windows compatibility


sebasv:
This code is cross-platform save for the "executable file" check, which is only relevant on unix systems, and is done for the sole purpose of color-coding output lines. 

I managed to make your code run on windows by adding conditional compilation operators to 
`use std::os::unix::fs::PermissionsExt;` and to redefine the `is_executable` function on non-Linux systems.

sharkdp:
Thank you for trying this. I'll try to implement this as soon as I get access to a Windows system.

sebasv:
I'd be happy to create a pull request!

sharkdp:
That would be great, thank you.

sharkdp:
Apart from the current compilation problems, colored output does not work on Windows (no support for ANSI sequences). We could consider using [termcolor](https://docs.rs/termcolor/0.3.2/termcolor/index.html).

sebasv:
#35 
Powershell does work with colored output. To be honest I haven't touched cmd since the inception of Powershell, but the termcolor crate looks like a nice feature for the hardcore cmd users!

sharkdp:
Oh, I didn't know that, that's cool! I just quickly tried `cmd` today.. thanks!

sharkdp:
Closed via #35

jmcomets:
I'm surprised this was closed considering colored output does not work in `cmd.exe`, I'm getting ansi escapes printed to the console every time I fire up `fd` and have to add the `--no-color` option each time. Maybe we should re-open this/add an issue on Windows coloring? [termcolor](https://github.com/BurntSushi/ripgrep/tree/master/termcolor) crate sounded like a great option.

sebasv:
This issue was regarding compatibility; The code simply didn't compile on windows. Having to flag `no-color` should be the default behavior imho since this is a shortcoming of `cmd`, not Windows. If you want colored output, try Powershell. It's like bash and Python had a baby in Microsoft country. (And it's developed by MSFT)

jmcomets:
I disagree, we have colored output in CMD via the various `color` calls.
Surely if `ripgrep` can do it, `fd` can do it.

sebasv:
Then by all means, start a pull request to add `cmd` support. I'm just saying that `cmd` support is not the same as windows compatibility, and should be opened as another issue.

jmcomets:
[...]

PR #109, opened 2017-10-13
Update ansi_term to 0.10

equal-l2:


sharkdp:
@equal-l2 Hi and thank you for your contribution!

Is there any specific reason why you want to update `ansi_term`?

equal-l2:
No specific reason, actually.

By the way, Travis found that the update requires bump minimum rustc version.
Appvayor fails to compile.
When there's no reason or new feature, this update seems to be just a pain.

Closing now.

sharkdp:
Actually, this brought to my attention that there now seems to be a `ansi_term::enable_ansi_support()` method for Windows 10 in `ansi_term`. Maybe we should upgrade after all, since this could help with the color problems on Windows (#29)  :smile:

PR #110, opened 2017-10-13
Improve smart case to only consider literal uppercase chars

### LINKED ISSUE #103: Smart case is not as smart as ripgrep's
Although I'm not a fan of this feature, you may be interested.

Example: `\Acargo` and `carg\x6F` do not match "Cargo".

[ripgrep uses the regex-syntax crate to check the regex AST](https://github.com/BurntSushi/ripgrep/blob/12ffcb4296e0c2068797daea6ed88885b060d916/grep/src/search.rs#L199-L215).

reima:
Fixes #103.

Uses a slightly simplified version of what ripgrep does.


sharkdp:
This is great, thank you very much!

PR #113, opened 2017-10-14
Use a uniform output format for searching ../

### LINKED ISSUE #82: Path of directory symlink is resolved to real path
```
$ tree -F
.
├── a/
│   └── a
└── b -> a/
```

In directory "b":

```
$ fd -a  # or fd -a . .
/path/a
/path/a/a
$ fd . $PWD
/path/a
/path/a/a
```

(For `fd . $PWD`, with or without option `-L`?) The result should be `/path/b` and `/path/b/a` unless resolving is requested by the user.

### LINKED ISSUE #107: Uniform output format when searching a parent directory
(courtesy of @iology in #102)

Consider this directory tree:
```
> tree
.
├── dir1
│   ├── file1
│   └── file2
└── dir2
    ├── file3
    └── file4

2 directories, 4 files
```

Now, go to `dir1` and perform a search in the parent directory:
```
> cd dir1
> fd '' ..
.
../dir2
../dir2/file3
../dir2/file4
file1
file2
```

The results are correct, but it would be more helpful if the output would look like this (in some order):
```
> cd dir1
> fd '' ..
../dir1
../dir2
../dir2/file3
../dir2/file4
../dir1/file1
../dir1/file2
```

iology:
closes #107, fixes #82 by the way (Yeah, I don't want to split this patch.)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,7 +16,7 @@ mod walk;
 
 use std::env;
 use std::error::Error;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time;
 
@@ -35,26 +35,16 @@ fn main() {
     let pattern = matches.value_of("pattern").unwrap_or(&empty_pattern);
 
     // Get the current working directory
-    let current_dir_buf = match env::current_dir() {
-        Ok(cd) => cd,
-        Err(_) => error("Error: could not get current directory."),
-    };
-    let current_dir = current_dir_buf.as_path();
+    let current_dir = Path::new(".");
+    if !current_dir.is_dir() {
+        error("Error: could not get current directory.");
+    }
 
     // Get the root directory for the search
-    let mut root_dir_is_absolute = false;
-    let root_dir_buf = if let Some(rd) = matches.value_of("path") {
-        let path = Path::new(rd);
-
-        root_dir_is_absolute = path.is_absolute();
-
-        fshelper::absolute_path(path).unwrap_or_else(|_| {
-            error(&format!("Error: could not find directory '{}'.", rd))
-        })
-    } else {
-        current_dir_buf.clone()
+    let mut root_dir_buf = match matches.value_of("path") {
+        Some(path) => PathBuf::from(path),
+        None => current_dir.to_path_buf(),
     };
-
     if !root_dir_buf.is_dir() {
         error(&format!(
             "Error: '{}' is not a directory.",
@@ -62,6 +52,15 @@ fn main() {
         ));
     }
 
+    let path_display = if matches.is_present("absolute-path") || root_dir_buf.is_absolute() {
+        PathDisplay::Absolute
+    } else {
+        PathDisplay::Relative
+    };
+
+    if path_display == PathDisplay::Absolute && root_dir_buf.is_relative() {
+        root_dir_buf = fshelper::absolute_path(root_dir_buf.as_path()).unwrap();
+    }
     let root_dir = root_dir_buf.as_path();
 
     // The search will be case-sensitive if the command line flag is set or
@@ -109,11 +108,7 @@ fn main() {
             .value_of("max-buffer-time")
             .and_then(|n| u64::from_str_radix(n, 10).ok())
             .map(time::Duration::from_millis),
-        path_display: if matches.is_present("absolute-path") || root_dir_is_absolute {
-            PathDisplay::Absolute
-        } else {
-            PathDisplay::Relative
-        },
+        path_display: path_display,
         ls_colors: ls_colors,
         file_type: match matches.value_of("file-type") {
             Some("f") | Some("file") => FileType::RegularFile,
@@ -127,17 +122,21 @@ fn main() {
         }),
     };
 
-    let root = Path::new(ROOT_DIR);
-    let base = match config.path_display {
-        PathDisplay::Relative => current_dir,
-        PathDisplay::Absolute => root,
+    // If base_dir is ROOT_DIR, then root_dir must be absolute.
+    // Otherwise root_dir/entry cannot be turned into an existing relative path from base_dir.
+    //
+    // We utilize ROOT_DIR to avoid resolving parent directories the root_dir.
+    let base_dir_buf = match config.path_display {
+        PathDisplay::Relative => current_dir.to_path_buf(),
+        PathDisplay::Absolute => PathBuf::from(ROOT_DIR),
     };
+    let base_dir = base_dir_buf.as_path();
 
     match RegexBuilder::new(pattern)
         .case_insensitive(!config.case_sensitive)
         .dot_matches_new_line(true)
         .build() {
-        Ok(re) => walk::scan(root_dir, Arc::new(re), base, Arc::new(config)),
+        Ok(re) => walk::scan(root_dir, Arc::new(re), base_dir, Arc::new(config)),
         Err(err) => error(err.description()),
     }
 }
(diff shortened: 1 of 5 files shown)


iology:
Ouch, now rebased after `cargo fmt`.

src/output.rs
```diff
@@ -11,7 +11,11 @@ use std::os::unix::fs::PermissionsExt;
 use ansi_term;
 
 pub fn print_entry(base: &Path, entry: &PathBuf, config: &FdOptions) {
-    let path_full = base.join(entry);
+    let path_full = if *entry != PathBuf::new() {
```
reima: Consider `if !entry.as_os_str().is_empty()` to avoid the allocation.

iology:
@sharkdp Could you figure out what happened here: https://ci.appveyor.com/project/sharkdp/fd/build/1.0.108/job/agyctswxi2ymut4m#L202

tests/tests.rs
```diff
@@ -4,6 +4,21 @@ mod testenv;
 
 use testenv::TestEnv;
 
+fn get_absolute_root_path(env: &TestEnv) -> String {
+    let path = (*env)
```
reima: No need to dereference `env` here. Rust does it automatically.

tests/tests.rs
```diff
@@ -416,3 +423,58 @@ fn test_extension() {
...
+            {abs_path}/one/two/three/d.foo
+            {abs_path}/one/two/three/directory_foo",
+            abs_path = abs_path
+        ),
+    );
+
+    te.assert_output_subdirectory(
+        &abs_path,
+        &[
+            "--absolute-path",
```
iology: This flag is not needed, will be removed.

sharkdp:
> @sharkdp Could you figure out what happened here: https://ci.appveyor.com/project/sharkdp/fd/build/1.0.108/job/agyctswxi2ymut4m#L202

Not sure ... does Windows treat `../..` correctly? Or should this be `..\..`?

reima:
Windows can work with `/` as a path separator just fine.

I think the problem is that symlinks work differently on Windows. On Unix, a symlink `symlink -> one/two` is like an alias for `one/two`, so `symlink/..` is equivalent to `one/two/..`, which is `one`. But on Windows `symlink\..` is whatever directory `symlink` resides in.

I'm not certain if or how we should fix this. In my experience symlinks are still very rarely used in Windows land, so personally I wouldn't bother too much with this (i.e. disable the test on Windows or special case it).

sharkdp:
> I think the problem is that symlinks work differently on Windows. On Unix, a symlink symlink -> one/two is like an alias for one/two, so symlink/.. is equivalent to one/two/.., which is one. But on Windows symlink\.. is whatever directory symlink resides in.

Thanks, I didn't know that.

> I'm not certain if or how we should fix this. In my experience symlinks are still very rarely used in Windows land, so personally I wouldn't bother too much with this (i.e. disable the test on Windows or special case it).

:+1:, in particular if we can add your explanation above as a comment why it's disabled/special-cased on Windows.

sharkdp:
Playing around with this, I've noticed the following (I'm inside a test folder, in the `symlink` directory):
```
> echo "$PWD" 
/home/shark/fd-tests.4ULpgs5vXW/symlink

> fd '' "$PWD"
/home/shark/fd-tests.4ULpgs5vXW/one/two/../../symlink/C.Foo2
/home/shark/fd-tests.4ULpgs5vXW/one/two/../../symlink/c.foo
/home/shark/fd-tests.4ULpgs5vXW/one/two/../../symlink/three
/home/shark/fd-tests.4ULpgs5vXW/one/two/../../symlink/three/d.foo
/home/shark/fd-tests.4ULpgs5vXW/one/two/../../symlink/three/directory_foo
```

iology:
Building on AppVeyor is really slow, can you enable cache for the "target" directory?

iology:
@reima Simply put, symlinks on Windows can refer to symlinks. They are references, not aliases to fixed real paths.

sharkdp:
@iology Nice! Do you think there is anything we can do about https://github.com/sharkdp/fd/pull/113#issuecomment-336639772 ?

iology:
> Nice! Do you think there is anything we can do about #113 (comment) ?

It should be fixed, but you can try to create more complicated examples.

Also please add/amend comments in the code if you see fit. Must have a nice holidy after this tiresome work. :P

src/main.rs
```diff
@@ -35,33 +35,32 @@ fn main() {
     let pattern = matches.value_of("pattern").unwrap_or(&empty_pattern);
 
     // Get the current working directory
-    let current_dir_buf = match env::current_dir() {
-        Ok(cd) => cd,
-        Err(_) => error("Error: could not get current directory."),
-    };
-    let current_dir = current_dir_buf.as_path();
+    let current_dir = Path::new(".");
```
sharkdp: Why did you change this?

The detection of the current directory being missing does not seem to work anymore.

src/main.rs
```diff
@@ -35,33 +35,32 @@ fn main() {
...
     }
 
+    let path_display = if matches.is_present("absolute-path") || root_dir_buf.is_absolute() {
+        PathDisplay::Absolute
+    } else {
+        PathDisplay::Relative
+    };
+
+    if path_display == PathDisplay::Absolute && root_dir_buf.is_relative() {
+        root_dir_buf = fshelper::absolute_path(root_dir_buf.as_path()).unwrap();
```
sharkdp: We should avoid using `unwrap` here:
```
> fd -a non/existing/relative/path
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 2, message: "No such file or directory" } }', /checkout/src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

tests/tests.rs
```diff
@@ -4,6 +4,20 @@ mod testenv;
 
 use testenv::TestEnv;
 
+fn get_absolute_root_path(env: &TestEnv) -> String {
```
sharkdp: :+1:

src/main.rs
iology: So it seems that .is_dir() doesn't really check the existence of path. :(

src/main.rs
iology: I think you forgot to add the pattern.

sharkdp: You are absolutely right :smile:. So `unwrap` was only called because the current directory did not exist (the issue above).

src/main.rs
sharkdp: Maybe `Path::exists` helps? Does this have to be changed?

iology: .exists() doesn't help too, on both stable&nightly branches. Path::new("") is ok, but "." and ".." are not.

iology: Oh, "" doesn't mean current directory. I will ask the rust team.

sharkdp: What was wrong with the old way (`env::current_directory()`)?

iology: env::current_dir() does check the existence. The returned path is resolved, so its value is used only when we don't have an absolute path.

iology: `if let Err(_) = env::current_dir()` can help, but for root_dir, user still can pass in "." or "..". I hope there is an elegant way to check their existence.

iology: Now I'm not sure if we should care the existence of "." and "..". Just let it crash?

sharkdp:
Thank you very much! Very nice work.

PR #114, opened 2017-10-14
Dot "." also matches newline by default (closes #111)

### LINKED ISSUE #111: Allow dot "." to match newlines by default
It may seems dangerous to include newlines in the output, though there is an option `-0, --print0` to help.

But even if the so called "dot-all" feature is disabled, we still can match those files with empty pattern "", `^`,  `[^/]*`, etc.

iology:
Please ignore another patch in issue 96.


sharkdp:
Thanks!

PR #115, opened 2017-10-14
Enable caching on appveyor

sharkdp:
@iology I've seen this in some other appveyor.yml files. Would this work?


iology:
Maybe change `target` to `target\debug\deps -> Cargo.lock`.

PR #118, opened 2017-10-15
AppVeyor: fix path of cache direcotry

iology:
`target/debug/.fingerprint/` and `target/debug/deps` are both important. The file structure may changed in the future, so just cache the whole `target` directory. Only clear cache when Cargo.lock changes, so that it won't exceed the space capacity.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -89,5 +89,5 @@ branches:
     - /^v\d+\.\d+\.\d+.*$/
 
 cache:
-  - target\debug\deps -> Cargo.lock
+  - target -> Cargo.lock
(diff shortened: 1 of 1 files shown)


sharkdp:
Thanks!

sharkdp:
I didn't notice this before merging. Actually, there is this error message:
```
Error uploading cache entry to the cache storage: Remote server returned 500: There is not enough space on the disk.
```

sharkdp:
well.. before this PR, this was also not working on all virtual machines

iology:
What a desperate fact. The cache files have many hash IDs and all tests share the same space. If we can find a way to remove old deps...

PR #120, opened 2017-10-15
Do rework for filter and match process

topelinux:
De rework for filter and match process to make it easy understanding

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -9,7 +9,7 @@ use std::thread;
 use std::time;
 
 use regex::Regex;
-use ignore::{self, WalkBuilder};
+use ignore::{self, WalkBuilder, DirEntry};
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
 enum ReceiverMode {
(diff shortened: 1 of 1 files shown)


sharkdp:
@topelinux Welcome! Thank you very much for your contribution!

There is definitely some room for optimization/refactoring in the `scan` function, but to me, this particular approach doesn't feel quite like it, to be honest. `DirEntryResolveEngine` doesn't seem like a very natural abstraction to me.

That being said, I'm definitely open to discussions about how we could improve the filter/matching code in `scan`. Thanks!

topelinux:
@sharkdp thanks for your reply
Looks like the name `DirEntryResolveEngine` is not a good name for you, what about its responsibility?  Do you agree, the engine/function (without the new name come out, just use`DirEntryResolveEngine` temporally) should focus on `filter`, `match` and `resolve`?

sharkdp:
@topelinux It's not just the name... it's more the concept of creating a new struct that does these two things.

I would rather refactor into separate functions, yes - but I'm not sure how this could look like exactly. Maybe we should discuss it in a new ticket first. Also, a few things have changed now with #123.

PR #124, opened 2017-10-16
Allow overriding of OUT_DIR in build.rs

### LINKED ISSUE #121: Shell completions OUT_DIR
Setting OUT_DIR doesn't seem to have any effect on where the completions files are created. Am I doing something wrong?

ImbaKnugel:
Fixes #121.
* Uses SHELL_COMPLETIONS_DIR to change the output directory of the completions.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,7 +6,7 @@ use clap::Shell;
 include!("src/app.rs");
 
 fn main() {
-    let var = std::env::var_os("OVRD_OUT_DIR").or_else(|| std::env::var_os("OUT_DIR"));
+    let var = std::env::var_os("SHELL_COMPLETIONS_DIR").or(std::env::var_os("OUT_DIR"));
     let outdir = match var {
         None => return,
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -6,7 +6,8 @@ use clap::Shell;
 include!("src/app.rs");
 
 fn main() {
-    let outdir = match std::env::var_os("OUT_DIR") {
+    let var = std::env::var_os("OVRD_OUT_DIR").or_else(|| std::env::var_os("OUT_DIR"));
```
sharkdp: Thank you very much!

What about naming this something like `SHELL_COMPLETIONS_DIR`?

Also, I think you can use `Option::or` here to avoid the lambda expression.

ImbaKnugel:
Removed the lambda and renamed the env variable to SHELL_COMPLETIONS_DIR.

sharkdp:
Thank you very much.

PR #125, opened 2017-10-17
Add fd.1 man page, close #77

### LINKED ISSUE #77: Add man (1) entry
Adding a man entry would be a nice improvement for when `fd` is packaged ([And possibly some day for when installed via `cargo install`?](https://github.com/rust-lang/cargo/issues/2729)).

pickfire:
https://github.com/sharkdp/fd/issues/77

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -3,7 +3,7 @@
 fd \- find files
 .BR find (1)
 .SH SYNOPSIS
-.B spt
+.B fd
 .RB [ \-HIsaLp0hV ]
 .RB [ \- d
 .IR depth ]
@@ -15,6 +15,8 @@ fd \- find files
 .IR when ]
 .RB [ \-j
 .IR num ]
+.RB [ \-x
+.IR exec ]
 .RI [ pattern ]
 .RI [ path ]
 .SH DESCRIPTION
@@ -26,12 +28,12 @@ is a simple, fast and user-friendly alternative to
 .B \-H, \-\-hidden
 Search hidden files and directories.
 .TP
-.B \-I, \-\-no\-gitignore
+.B \-I, \-\-no\-ignore
 Do not respect any
 .BR gitignore (5)
 files.
 .TP
-.B \-s, \-\-case\-insensitive
+.B \-s, \-\-case\-sensitive
 Perform a case-sensitive search (default: smart case).
 .TP
 .B \-a, \-\-absolute\-path
@@ -43,15 +45,16 @@ Dereference all symbolic links encountered.
 .B \-p, \-\-full\-path
 Match
 .I pattern
-against full path.
+against full
+.IR path .
 .TP
 .B \-0, \-\-print0
-Print results followed by a null character (instead of newlines).
+Print results followed by null character instead of newlines.
 .TP
 .B \-h, \-\-help
 Prints help information.
 .TP
-.B \-v, \-\-version
+.B \-V, \-\-version
 Prints version information.
 .TP
 .BI "\-d, \-\-max\-depth " levels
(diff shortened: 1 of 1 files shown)


doc/fd.1
```diff
@@ -0,0 +1,106 @@
+.TH FD 1 fd-4.0.0
+.SH NAME
+fd \- find files
```
sharkdp: Maybe this could be a little bit more detailed: "fd - find entries in the filesystem"?

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.BR find (1)
```
sharkdp: What is this for? is it a link to `find`? Do we need this here?

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.SH SYNOPSIS
+.B spt
```
sharkdp: This should be `fd`, probably :smile:

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.SH DESCRIPTION
+.B fd
+is a simple, fast and user-friendly alternative to
+.BR find (1).
+.SH OPTIONS
+.TP
+.B \-H, \-\-hidden
+Search hidden files and directories.
+.TP
+.B \-I, \-\-no\-gitignore
```
sharkdp: This should be `--no-ignore`.

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+Do not respect any
+.BR gitignore (5)
+files.
+.TP
+.B \-s, \-\-case\-insensitive
```
sharkdp: This should be `--case-sensitive`

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.I pattern
+against full path.
+.TP
+.B \-0, \-\-print0
+Print results followed by a null character (instead of newlines).
+.TP
+.B \-h, \-\-help
+Prints help information.
+.TP
+.B \-v, \-\-version
```
sharkdp: This should be `-V`, not `-v`.

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+Show colors only when standard output is connected to a terminal. (default)
+.IP never
+Do not use colorized output.
+.IP always
+Always use colorized output.
+.RE
+.TP
+.BI "\-j, \-\-threads " num
+Specifies number of threads to use for searching. (default: number of available
+CPU cores)
```
sharkdp: Can we also add the new option `-x`, `--exec`?

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.SH ENVIRONMENT
+.TP
+LS_COLORS
+.BR "fd"'s
+color codes for search results.
```
sharkdp: `LS_COLORS` is also used by `ls` and typically set by `dircolors`. Maybe we should say something like: "determines how to colorize the search results, see DIRCOLORS(1)".

doc/fd.1
```diff
@@ -0,0 +1,106 @@
...
+.SH EXAMPLES
+.TP
+.RI "Find the specific log files in " /var/log .
+$ fd nginx.log /var/log
```
sharkdp: Maybe just use `fd nginx /var/log` here. The `.` can actually mean "any character" in a regex, so this does not perform a literal search for "nginx.log".

doc/fd.1
pickfire: Yes, this is a link to find since it's on the current description.

doc/fd.1
pickfire: Ops, sorry. :P

doc/fd.1
pickfire: I took that from the posix man page and sbase man page. Both uses that.

doc/fd.1
pickfire: Nice suggestion, I don't have better idea to write this environment part.

sharkdp:
Fantastic, thank you very much!

PR #131, opened 2017-10-18
RedoxOS support.

goyox86:
Just small tweaks to make it compile (and run) in RedoxOS!

Works like a charm BTW!

\o/

### DIFF
--- a/src/exec/ticket.rs
+++ b/src/exec/ticket.rs
@@ -30,7 +30,7 @@ impl<'a> CommandTicket<'a> {
 
     /// Executes the command stored within the ticket, and
     /// clearing the command's buffer when finished.'
-    #[cfg(not(all(unix, not(target_os = "redox"))))]
+    #[cfg(not(unix))]
     pub fn then_execute(self) {
         use std::process::Stdio;
         use std::io::Write;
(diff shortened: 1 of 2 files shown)


sharkdp:

Do all test run on RedoxOS (`cargo test`)? I'm wondering because there are a few other places where we have conditionals like `#[cfg(unix)]`.

goyox86:
Thank you for the tool!

Basically we cannot run `cargo test` inside Redox yet.  I basically ran a few manual tests inside Redox and everything seems to be working fine.

Regarding to the `cfg` uses I've checked and `then_execute` was the only one I saw with conditional compilation involving Redox I just removed Redox from the two conditionals and Now I think we are using the `libstd` version of `then_execute` which is fine as redox has full support. But I dunno I maybe wrong about that removal. 

Also in the UNIX `libc` based version you were importing `*` from `libc` which is a lot hehe they thow everything in the root module. 

Let me know if the patch still makes sense!

Regards!

goyox86:
Ah BTW none of the other platforms seems to be broken with the patch I think we are safe merging, and I commit to continue iterating on the upcoming issues regarding Redox integration 😄

sharkdp:
@mmstick Would you mind taking a quick look at this since you initially introduced the `redox` conditionals?

mmstick:
Looks okay to me. Redox does have the syscalls needed to be optimized as the Unix platforms though. You just need to grab the same functions from redox_syscalls.

Issue #76, opened 2017-10-07
Set up Travis to upload binaries for every release


sharkdp:
This already works for Windows builds (via Appveyor, see `appveyor.yml`). It would be great if we could also upload Linux and macOS tarballs, as well.

sharkdp:
ripgrep and tokei seem to do something like this:

* https://github.com/BurntSushi/ripgrep/blob/master/.travis.yml
* https://github.com/Aaronepower/tokei/blob/master/.travis.yml

matematikaadit:
I'm interested in submitting PR for this. Will try to working it out first in a fork and submit it later.

sharkdp:
> I'm interested in submitting PR for this. Will try to working it out first in a fork and submit it later.

Sounds great!

matematikaadit:
Sorry for the slow progress. I've been experimenting and figuring stuff out on building rust project on travis. And I think I found an issue regarding the current `.travis.yml` setting. Opening issue #132 and planning working on that first.

PR #133, opened 2017-10-19
Incorporate TARGET in the install and build step

### LINKED ISSUE #132: Current setting in .travis.yml isn't using the TARGET env var for building the project
So I've been pondering the current [`.travis.yml`](https://github.com/sharkdp/fd/blob/8d85debc12631d47274da8305e710723aae5dc74/.travis.yml) setting. It has multiple item in the build matrix with different TARGET env var. But there's no usage of the TARGET implied anywhere. Reading [the documentation on building rust project on travis-ci.org](https://docs.travis-ci.com/user/languages/rust/) also confirm that the default "install" and "script" stage on Travis CI isn't using the TARGET var in any meaningful way, except setting it out in the build environment. Which of course, wont be picked up by `cargo build` if we not explicitly using it as the `--target` value.

Here some example how they're used in ripgrep repo,

[`ci/install.sh`](https://github.com/BurntSushi/ripgrep/blob/a98156e71cc2aefd2d3e18c76c6d09cc1ef8c5ac/ci/install.sh#L28)
```bash
install_standard_crates() {
    if [ $(host) != "$TARGET" ]; then
        rustup target add $TARGET
    fi
}
```
[`ci/script.sh`](https://github.com/BurntSushi/ripgrep/blob/a98156e71cc2aefd2d3e18c76c6d09cc1ef8c5ac/ci/script.sh#L19-L29)
```bash
run_test_suite() {
    cargo clean --target $TARGET --verbose
    cargo build --target $TARGET --verbose
    cargo test --target $TARGET --verbose
    # ...
}
```

If you're okay with my slow progress, I wanna submit a PR to solve this issue. Tho, if someone else interested in this, I'm okay with that. For reference here some minimal working example of `.travis.yml` setting for building rust for `x86_64` and `i686` target on linux.

https://github.com/matematikaadit/rust-travis-eureka/blob/e0abfbb4bf6bac0466b0549d7aa6b54e7af4b43c/.travis.yml

It targets:

- x86_64-unknown-linux-gnu
- x86_64-unknown-linux-musl
- i686-unknown-linux-gnu
- i686-unknown-linux-musl

on stable, beta, and nightly channels.

matematikaadit:
- Tidying up the build matrix
  - All channels has TARGET env var
  - Build against this targets:
    - x86_64-unknown-linux-gnu
    - x86_64-unknown-linux-musl
    - i686-unknown-linux-gnu
    - i686-unknown-linux-musl
    - x86_64-apple-darwin
  - Code formating check has it's own item in the build matrix
- Prevent target re-add error from rustup
- Incorporate the TARGET env var in the `install` and `script` stage

fix #132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,35 +25,37 @@ matrix:
     - os: linux
       rust: beta
       env: TARGET=x86_64-unknown-linux-gnu
-    - os: linux
-      rust: beta
-      env: TARGET=x86_64-unknown-linux-musl
-    - os: linux
-      rust: beta
-      env: TARGET=i686-unknown-linux-gnu
-    - os: linux
-      rust: beta
-      env: TARGET=i686-unknown-linux-musl
-    - os: osx
-      rust: beta
-      env: TARGET=x86_64-apple-darwin
+    # Disabled to reduce total CI time
+    # - os: linux
+    #   rust: beta
+    #   env: TARGET=x86_64-unknown-linux-musl
+    # - os: linux
+    #   rust: beta
+    #   env: TARGET=i686-unknown-linux-gnu
+    # - os: linux
+    #   rust: beta
+    #   env: TARGET=i686-unknown-linux-musl
+    # - os: osx
+    #   rust: beta
+    #   env: TARGET=x86_64-apple-darwin
 
     # Nightly channel.
     - os: linux
       rust: nightly
       env: TARGET=x86_64-unknown-linux-gnu
-    - os: linux
-      rust: nightly
-      env: TARGET=x86_64-unknown-linux-musl
-    - os: linux
-      rust: nightly
-      env: TARGET=i686-unknown-linux-gnu
-    - os: linux
-      rust: nightly
-      env: TARGET=i686-unknown-linux-musl
-    - os: osx
-      rust: nightly
-      env: TARGET=x86_64-apple-darwin
+    # Disabled to reduce total CI time
+    # - os: linux
+    #   rust: nightly
+    #   env: TARGET=x86_64-unknown-linux-musl
+    # - os: linux
+    #   rust: nightly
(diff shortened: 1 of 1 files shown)


sharkdp:
Very cool, thank you very much for your contribution!

Do you think there is some way to reduce the total CI time a bit? Previously, Travis was running in 20-30min. Now it takes 1h 15min. Maybe we don't have to build nightly and beta for every target?

matematikaadit:
Feel free to exclude any matrix item. I just listed all of them because they're readily available and adding them is no hassle.

Probably only test for `x86_64-unknown-linux-gnu` on nightly and beta and remove the other target? Also, the target that listed under the `# Stable channel.` might become the one for the release deployment  (issue #76). Feel free to drop any item that we don't wanna add to the release.

sharkdp:
> Probably only test for x86_64-unknown-linux-gnu on nightly and beta and remove the other target?

Sounds good (maybe just comment them out for now). Would you like to include this within this PR?

matematikaadit:
With pleasure.

Disabling:
- x86_64-unknown-linux-musl
- i686-unknown-linux-gnu
- i686-unknown-linux-musl
- x86_64-apple-darwin

on beta and nightly channels.

sharkdp:
Thank you very much!!

matematikaadit:
Your welcome

Nice, total CI build time reduced to 16 minutes. 😄 

Also I have some suspicion about what causing [the failing code formatting check on nightly](https://travis-ci.org/sharkdp/fd/jobs/290322690). Currently investigating it.

pickfire:
Maybe we could include `fd.1` along with the compressed tarball? Reference implementation https://github.com/BurntSushi/ripgrep/blob/master/.travis.yml#L52-L53

@sharkdp What do you think?

matematikaadit:
Better place to discuss it I think is #76 .

Issue #134, opened 2017-10-19
Write integration tests for -x/--exec option


sharkdp:
We are still missing a few integration tests for the `-x`/`--exec` option.

The tests could be very simple at first. Like comparing the output of `fd -x echo ...` with the output `fd ...` (where the latter would be hard-coded in the test specification).

sharkdp:
Closed via #142

Issue #135, opened 2017-10-19
Path should be escaped when running '-x cmd'


sharkdp:
This seems like a bug to me

``` bash
> touch 'file with spaces'

> fd -x 'ls "{}"'    # this is fine
file with spaces

> fd -x ls       
ls: cannot access 'file': No such file or directory
ls: cannot access 'with': No such file or directory
ls: cannot access 'spaces': No such file or directory
```

@mmstick - I'm mentioning you here just in case you are interested. I'm not trying to assign this to you.

mmstick:
GNU Parallel also implements these flags for this use case -- quoting inputs. Not really a bug, but a convenience flag so you don't have to manually quote it yourself.

mmstick:
The `--quote` flag, that is

iology:
Why not quote it by default?

mmstick:
@iology You can have an input that provides multiple arguments. In that case, you don't want quoting to happen.

mmstick:
Although I guess when all of your inputs are files, it'd make sense to quote by default.

sharkdp:
> Although I guess when all of your inputs are files, it'd make sense to quote by default.

Yes, I also think so. The flag might be useful for `parallel`, but it shouldn't be needed here.

mmstick:
I'll have this fixed soon. Am going to look for a way of possibly optimizing my solution though. Best way to do this isn't to quote, exactly, but to escape special characters.

mmstick:
Am going to convert that `paths` module into an `input` module that uses a state machine to generate input strings, only allocating as needed. Will ensure that the programmer doesn't forget to escape the input before using it, preventing a possible logic error.

PR #136, opened 2017-10-19
Adding integration test for -x/--exec

DTibbs:
Added some testing of --exec command for Issue #134

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -48,9 +48,9 @@ build: false
 
 # Equivalent to Travis' `script` phase
 test_script:
-  - cargo build --verbose
-  - cargo run
-  - cargo test
+  - cmd: cargo build --verbose
+  - cmd: cargo run
+  - cmd: cargo test
 
 before_deploy:
   # Generate artifacts for release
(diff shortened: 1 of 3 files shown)


DTibbs:
Interesting AppVeyor failed on `x86_64-pc-windows-msvc` as that was the platform and toolchain I tested it locally on. Looks like a powershell path separator issue? I tested in cmd.exe.

DTibbs:
Now I'm just guessing what AppVeyor's issue is. I don't get the same result testing locally in powershell w/ same version of cargo/rustc/etc.
Going to try and escape the string after `--exec` to see if that fixes it, as the following looks like it _might_ be the problem:
```	thread 'test_exec' panicked at '`fd --exec echo found foo` did not exit successfully.```

DTibbs:
Looks like the test on AppVeyor for `x86_64-pc-windows-msvc` is returning error code C0000374. 
No idea what's causing that. As evident in the log, there is nothing on stderr.
This one is a stumper. Something about how the processes are being spawned that powershell x64 doesn't like?
Last try is to specify `cmd` in the appveyor.yml. Not sure if that's desired. I do not use, and haven't seen a lot of powershell use, when developing in Rust on Windows, so maybe it's not that big of a deal to use `cmd` on AppVeyor.

sharkdp:
Welcome and thank you very much for working on this!

#135 has been fixed in the meantime. I haven't looked at the CI failures here, but maybe this could help.

sharkdp:
See #142 by @iology for some progress on this.

sharkdp:
I'm closing this in favor of #142. Thank you very much for your help!

PR #137, opened 2017-10-20
Auto disable colored output for unsupported Windows shells

### LINKED ISSUE #129: Colored output does not work on cmd.exe (Windows)
Need another colorizer for Windows.

iology:
Resolves #129

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,6 +13,7 @@ dependencies = [
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-syntax 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "windows 0.0.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


sharkdp:
Thank you very much for working on this!

I recently found this: https://github.com/ogham/rust-ansi-term/blob/090e0a383d73a43e2f80a7b466e8feeee97c4c76/src/windows.rs#L10

Do you think we could utilize this instead of pulling in all this OS-dependent code?

iology:
Oh yeah, very good. 😄 Rebased and tested on Win7 & Win10.

iology:
Oops, failed. Filed an issue to the upstream: https://github.com/ogham/rust-ansi-term/issues/31

iology:
The previous successful patch is here: https://github.com/jakwings/fd/commit/a44e6775ba4938bc1ecbe3735101a9a50474e740

iology:
While `ansi-term::enable_ansi_support` may take times to stabilize, I move jakwings/fd@a44e677 into a crate in the subdirectory "win", thus slightly reducing the OS macros used.

sharkdp:
Great, thank you very much!

Let's keep an eye on https://github.com/ogham/rust-ansi-term/issues/31

PR #139, opened 2017-10-20
Escape exec Inputs

### LINKED ISSUE #135: Path should be escaped when running '-x cmd'

mmstick:
Paths are now transformed using a state machine, which ensures that the final output is escaped when the inner value is returned. Nothing special beyond inputs now getting escaped.

Closes #135


sharkdp:
Super, thank you very much!

iology:
Other whitespace characters must also be escaped: TAB and LF

iology:
Also there is another special chars in Windows shells: `^`

mmstick:
Do tabs and such need to be escaped? I don't think it's possible to do that in a cross-shell manner. Bash, for example, would have to have it written like `$'path\twith\ttabs'`. In Ion, I've been thinking about adding an `$unescape('path\twith\ttabs')` method.

mmstick:
Maybe if we just replace them with their corresponding ASCII sequences, and let the user handle them accordingly?

iology:
The crate "shell-escape" just quotes the raw tabs/linefeeds/... in single-quoted strings. `$'...'` is also quoting. Anyway, the intermediate process is not important to the final result.

PR #140, opened 2017-10-20
README: Arch Linux has package fd-rs in official repos

cassava:
This is a small update to the README to point toward the officially maintained package rather than the one in AUR.


sharkdp:
Oh, very cool!

Thank you for the update.

Issue #141, opened 2017-10-21
--exec is too flexible


iology:
ping #139 (buggy)

I haven't found out a good way to escape those LFs and control-characters (U+00-1F, U+7F) with traditional posix shell syntax.

Also it seems weird to use cmd.exe if people are used to PowerShell.

I prefer `--exec <program> [arg]...` without escaping.

iology:
Reopen if you are not satisfied with pull #142.

PR #142, opened 2017-10-21
[WIP] Fix --exec

iology:
Problem solved but the tests are in a tangle.

ping #139 #141

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,7 @@ dependencies = [
  "num_cpus 1.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-syntax 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "shell-escape 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "windows 0.0.0",
 ]
@@ -226,6 +227,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "shell-escape"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "strsim"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


iology:
After decoupling the tests about shell_escape from the tests of src/exec/input.rs, it is still not easy to write thorough tests for --exec, unless we fixed the behavior of `shell_escape::escape` by specifying a fixed version string.

sharkdp:
This looks very good to me!

I'd like to hear @mmstick's feedback though.


@iology Do you have any idea what's causing the failure on `i686-unknown-linux-musl`?

iology:
Testing on Windows is a little hard. Always need more tests. Windows is not my daily OS, I just do anything to keep the features complete.

On Windows, A single drive name without slashes (e.g. `C:`, not `C:\`) behaves differently on different drives. So fd.exe still has bugs. ;-) Direct manipulation on path string, instead of using std::path::Path::components, is not DRY and makes this problem unnoticed further.

I don't use musl.

mmstick:
Looks fine to me. The tests pass with musl on my end with your PR, so I'm not sure why Travis errored there.

sharkdp:
Interestingly, the tests with musl pass on x86_64, but not i686.

sharkdp:
I am able to reproduce this locally. Here is what is happening:
```
Starting program: /home/shark/fd-jakwings/target/i686-unknown-linux-musl/debug/fd foo --exec 'echo {}'
[New LWP 20724]
[New LWP 20725]

Thread 2 "fd" received signal SIGILL, Illegal instruction.
[Switching to LWP 20724]
alloc::arc::{{impl}}::clone<fd::exec::TokenizedCommand> (self=0xf79ff510)
    at /checkout/src/liballoc/arc.rs:573
```

sharkdp:
The actual SIGILL seems to come from the `abort()` call here:
https://github.com/rust-lang/rust/blob/0ade339411587887bf01bcfa2e9ae4414c8900d4/src/liballoc/arc.rs#L562-L575

``` rust
        // However we need to guard against massive refcounts in case someone
        // is `mem::forget`ing Arcs. If we don't do this the count can overflow
        // and users will use-after free. We racily saturate to `isize::MAX` on
        // the assumption that there aren't ~2 billion threads incrementing
        // the reference count at once. This branch will never be taken in
        // any realistic program.
        //
        // We abort because such a program is incredibly degenerate, and we
        // don't care to support it.
        if old_size > MAX_REFCOUNT {
            unsafe {
                abort();
            }
        }
```

sharkdp:
So it seems like this is indeed somehow related to that `unsafe` block. The error can be fixed by replacing:
```
let cmd = unsafe { Arc::from_raw(cmd as *const TokenizedCommand) };
```
with
```
let cmd = Arc::new(cmd.clone());
``´

sharkdp:
@iology Could you please try to rebase this on top of master? This should hopefully fix the tests.

iology:
Sorry, I'm going to close this if nobody else wants to take good care of the Windows build. In my unpublished fork the Windows support has been completely removed. (a very backward-incompatible fork ;-)

This patch is not completed for the Windows build even after rebasing. So I can't say it fixes everything. Anyway the code is licensed for merging.

sharkdp:
> Sorry, I'm going to close this if nobody else wants to take good care of the Windows build.

I would really like to integrate the tests you wrote (for Linux). Does this PR make things worse on Windows? Otherwise, someone else can still continue working on the Windows part later.

> In my unpublished fork the Windows support has been completely removed. (a very backward-incompatible fork ;-)

I'm also not a Windows-developer and I think it's definitely not our main target platform, but there has  definitely been interest in a Windows version. However, I think it would be perfectly fine to support only a limited subset of features on Windows (the others might still be supported at a later point in time).

> This patch is not completed for the Windows build even after rebasing. So I can't say it fixes everything. Anyway the code is licensed for merging.

If it's not too much trouble, it would be great if you could still rebase it on master. I currently cannot merge or rebase (unless I would manually cherry-pick your commits on the command line or something similar).

Once you choose to publish your private fork, let me know! It would be a shame to lose you as a contributor. Anyways, thank you very much for your great support in the last few weeks!

iology:
Ok, I'll rebase and add some FIXME notes to the code.

> Once you choose to publish your private fork, let me know!

No problem. I've customized it for my taste. `--exec` will not depend on the system shells and resemble --exec of `find`. `./` is not omitted for relative paths. `<pattern>` comes after the `<root directory>`. `--type executable` is supported for regular files. And changes to flag names, error handling and other trivial things. (Also this is my way to learn Rust without bothering you too much. :P

iology:
Rebased, please check.

sharkdp:
Thank you very much! fd v5.0 coming in 3...2...1...

iology:
@sharkdp FYI: https://github.com/jakwings/ff-find (still a lot of todos)

sharkdp:
@iology Very cool! I will take a closer look soon. Maybe we can 'backport' a few ideas to *fd* :smile:

Issue #105, opened 2017-10-13
Consider relicensing under MIT/Apache-2.0


Detegr:
Some time ago, there were requests that MIT or Apache-2.0 licensed Rust projects would be relicensed under MIT/Apache-2.0 dual license, see [here](https://github.com/sfackler/rust-postgres-macros/issues/19).

Please consider if this would be acceptable for this project as well.

sharkdp:
To be honest, I haven't fully understood dual licensing.

If I understand correctly, it means that users of the software can choose between the two licenses.

MIT is more permissive than Apache-2.0 (due to the absence of patent clauses), so will the software still be protected from patent litigations if it is dual licensed? If not, what is the advantage of dual-licensing with Apache-2.0?

Detegr:
I am no means an expert when it comes to licenses and legal issues, but yes, as you stated it means that the users of the software can choose between the two licenses. If `fd` would be shipped within a project, the project could choose between MIT and Apache 2. This in turn gives more freedom to the user because it can be easier for the user to choose the more strict license (for example to omit copying the copyright statements that MIT requires).

Of course this change would require that every contributor of this project agrees that their contributions can be relicensed and that could be cumbersome.

sharkdp:
Thank you for the clarification. In this case, I'm open to relicensing *fd* under MIT/Apache-2.0. Would you be willing to organize the necessary steps (have all contributors sign, add the new license, modify Cargo file, etc.)?

mmstick:
I'll also add that I support the decision.

Detegr:
> Would you be willing to organize the necessary steps (have all contributors sign, add the new license, modify Cargo file, etc.)?

Absolutely.

## Contributor checkoff
If you agree to relicense, please leave the following comment to this issue:
```
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.
```
- [x] @sharkdp 
- [x] @mmstick 
- [x] @Detegr
- [ ] @reima 
- [ ] @iology 
- [ ] @unsignedint 
- [ ] @gsquire 
- [ ] @jcpetkovich 
- [ ] @olalonde 
- [ ] @abaez 
- [ ] @deg4uss3r 
- [ ] @ImbaKnugel 
- [ ] @exitium 
- [ ] @wezm 
- [ ] @michaelmior 
- [ ] @sebasv

Detegr:
I think it's best to be explicit (even though I did check some checkboxes already), so here's mine:

[...]

PR #143, opened 2017-10-21
Relicense under MIT/Apache-2.0

Detegr:
We're missing one code contributor in #105, but this is now ready to go after we have everybody's agreement of the relicensing.


Detegr:
All code contributors have now agreed to relicense. Two contributors have not signed #105 that have contributed README changes.

sharkdp:
@Detegr Thank you very much!

PR #144, opened 2017-10-21
Add filter trait for walk

topelinux:
Add filter trait for walk to separate filter logical from walk.run

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -9,7 +9,7 @@ use std::sync::mpsc::channel;
 use std::thread;
 use std::time;
 
-use ignore::{self, WalkBuilder};
+use ignore::{self, WalkBuilder, DirEntry};
 use regex::Regex;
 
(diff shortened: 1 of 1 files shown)


sharkdp:
@topelinux Thank you for your contribution.

I'm sorry, but I don't think this is a particularly idiomatic way to refactor this code. A trait is an interface description, a way to describe a particular property for a set of types (that implement the trait). It is useful if there are multiple types that implement the interface because the trait gives us a way to abstract over the concrete type that implements the trait. Here, the `Filter` trait looks like it is designed for the single concrete type that implements it (`FdOptions`).

topelinux:
What i want to do is separate the `filter` and `match` logical from the walker.run main logical, when the filter/match logical become complicated, it easy to main.
`trait` is the the way to meet this, that i could provide. To be frankly, I don't think we should struggle if  "it is designed for the single concrete type that implements it (FdOptions)."

PR #145, opened 2017-10-21
Using cargo-update to update the rustfmt on travis

### LINKED ISSUE #138: The cause of `rustfmt: error while loading shared libraries: libsyntax-6f7d7e35b339f76f.so` on Travis CI
The code formatting test on nightly fail with this error
https://travis-ci.org/sharkdp/fd/jobs/290400026#L602
```
$ cargo fmt -- --write-mode=diff
rustfmt: error while loading shared libraries: libsyntax-6f7d7e35b339f76f.so: cannot open shared object file: No such file or directory
```
The cause of those error is probably because the rustfmt binary being cached from the previous build, but the nightly's  libsyntax is already updated alongside the channel update.

## Solutions
We can chose one of these solution:
1. `--force` rustfmt-nightly reinstall on each build. Example: https://travis-ci.org/matematikaadit/fd/jobs/290417411#L641 . Tho, it's incur an additional build time.
2. Disable the nightly code formatting item from the matrix.
3. Use [cargo-update](https://crates.io/crates/cargo-update) so that `rustfmt-nightly` only updated if there's newer version available.

What do you think? Any other ideas? I'll be glad to submit PR for this.

matematikaadit:
Fix #138

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,12 +78,16 @@ matrix:
     - os: linux
       rust: stable
       # skip the global install step
-      install: which rustfmt || cargo install rustfmt
+      install:
+        - which cargo-install-update || cargo install cargo-update
+        - cargo install-update rustfmt
       script: cargo fmt -- --write-mode=diff
     - os: linux
       rust: nightly
       # skip the global install step
-      install: which rustfmt || cargo install rustfmt-nightly
+      install:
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

@Detegr Thank you for the review.

PR #147, opened 2017-10-22
CI experiment

sharkdp:
Just a small experiment - I'm not going to merge this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -48,9 +48,9 @@ build: false
 
 # Equivalent to Travis' `script` phase
 test_script:
-  - cargo build --verbose
-  - cargo run
-  - cargo test
+  - cmd: cargo build --verbose
+  - cmd: cargo run
+  - cmd: cargo test
 
 before_deploy:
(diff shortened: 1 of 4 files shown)


sharkdp:
Ok, so it seems like the `unsafe` block was (somehow) causing the CI failures in both #136 (on `x86_64-pc-windows-msvc`) and #142 (on `i686-unknown-linux-musl`).

Patching the `unsafe` block fixes the tests in both cases.

/CC @DTibbs @iology @mmstick

mmstick:
Is a bit wasteful to make a second copy of the command template for the sake of wrapping it in an Arc. I wonder why that was failing there.

sharkdp:
Closing in favor of #152

Detegr:
> I wonder why that was failing there.

[The documentation](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.from_raw) states that `The raw pointer must have been previously returned by a call to a Arc::into_raw` which was not the case here.

PR #148, opened 2017-10-22
Add support for exclude-patterns

### LINKED ISSUE #89: Add --exclude option.
Add an --exclude option which takes a .gitignore style pattern to exclude from the search (the option can be specified multiple times to add multiple excludes). 

Maybe also consider a complementary --include option.

# Implementation

The argument(s) would be added to the WalkBuilder with the `overrides` method, with a "!" prepended so they are exclusions, not inclusions. Alternatively, write the arguments to a temporary file and use `add_ignore`. (It may be worth adding a PR to `ignore` to add an easier way to add exclude globs to overrides).

sharkdp:
* Adds the `--exclude`/`-E` option.
* Support for multiple exclude patterns

Example:
``` bash
> fd --exclude 'tests/**/*.rs' mod
src/exec/mod.rs
src/fshelper/mod.rs
src/lscolors/mod.rs
```

I finally decided to call this `--exclude` because it is not exactly the same as ripgrep's `--glob` and I think the name is pretty clear.

Closes #89

/CC @tmccombs @Detegr

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -18,6 +18,7 @@ use std::thread;
 use std::time;
 
 use ignore::{self, WalkBuilder};
+use ignore::overrides::OverrideBuilder;
 use regex::Regex;
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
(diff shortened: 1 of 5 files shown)


src/walk.rs
```diff
@@ -48,13 +49,26 @@ pub fn scan(root: &Path, pattern: Arc<Regex>, config: Arc<FdOptions>) {
...
 
+    let mut override_builder = OverrideBuilder::new(root);
+
+    for pattern in config.exclude_patterns.iter() {
+        let res = override_builder.add(pattern);
+        if res.is_err() {
+            error(&format!("Error: malformed exclude pattern '{}'", pattern));
+        }
+    }
+    let overrides = override_builder.build().expect(
```
Detegr: Any reason we should not use:
```rust
let overrides = override_builder.build().unwrap_or_else(|_| {
    error("Mismatch in exclude patterns")
});
```
instead?

sharkdp: Thank you for the suggestion! That's better.

sharkdp:
Thank you very much for the review, @Detegr

Issue #155, opened 2017-10-30
--exec does not escape cmd.exe metacharacters (Windows)


reima:
On Windows, `--exec` launches the command through `cmd.exe`. This works fine as long as the file names do not contain any of the metacharacters `(`, `)`, `%`, `!`, `^`, `"`, `<`, `>`, `&`, or `|`. These metacharacters are not escaped by `fd`. As a consequence, they are interpreted by `cmd.exe` with their special meaning.

This may even lead to unwanted processes being started by `fd`:
```
E:\Development\Rust\Projects\fd\test>..\target\debug\fd.exe
x&whoami

E:\Development\Rust\Projects\fd\test>..\target\debug\fd.exe --exec "echo {}"
x
reima
```
`reima` is my Windows user name, which is printed by calling `whoami`.

There is also an issue with the escaping of file names with spaces (which  was not fixed with #135, at least on Windows):
```
E:\Development\Rust\Projects\fd\test>..\target\debug\fd.exe
a b c

E:\Development\Rust\Projects\fd\test>..\target\debug\fd.exe --exec "echo {}"
\"a b c\"
```

reima:
This makes we wonder why we take the detour with `cmd.exe`/`sh` at all. It incurs the runtime overhead of launching an additional process and it increases the implementation complexity.

On the other hand this enables features like shell output redirection without explicitly invoking a shell.

I'm not certain this is worth all of the trouble of multiple layers of escaping. `find` and `xargs` don't launch an intermediary shell either.

Looking forward to other thoughts on this.

mmstick:
The main reason for shelling out to a shell is because it's incredibly complex to properly parse shell syntax, and in doing so you will end up restricting yourself to whatever subset of shell functionality that your implementation supports.

> I'm not certain this is worth all of the trouble of multiple layers of escaping

Escaping currently only happens with a single line of code, so there aren't multiple layers -- just one.

https://github.com/sharkdp/fd/blob/614f576c47f2afaff77752d43ef7eb76dfb1131e/src/exec/input.rs#L95

Had a feeling this would happen with that crate. Probably best that we do the escaping ourselves.

mmstick:
We should simply check for any bytes within the range of 0...42, 58...64, 91...96, and 123...127, and then escape them accordingly. On *nix systems, that's the `\` character, and with cmd on Windows it is the `^` character.

mmstick:
Something like:

```rust
fn needs_escape(byte: u8) -> bool {
[...]

PR #160, opened 2017-11-03
Remove shell with --exec

reima:
This is a WIP/PoC for removing the intermediary shell when calling a command with `--exec`. See #155 for the discussion.


src/exec/ticket.rs
```diff
@@ -6,47 +6,39 @@
...
 /// A state that offers access to executing a generated command.
 ///
-/// The ticket holds a mutable reference to a string that contains the command to be executed.
-/// After execution of the the command via the `then_execute()` method, the string will be
-/// cleared so that a new command can be written to the string in the future.
-pub struct CommandTicket<'a> {
-    command: &'a mut String,
+/// The ticket holds the collection of arguments of a command to be executed.
+pub struct CommandTicket {
+    args: Vec<String>,
```
mmstick: Should probably try to eliminate the `Vec` allocations between commands by making this a `&'a mut Vec<String>`, and then implementing `Drop` for `CommandTicket` to clear the args buffer.

sharkdp:
Thank you very much! I'll add my review soon.

Does this change imply that we will have to put `--exec` after the command line arguments (pattern and root directory)?

reima:
With this change, `--exec` takes the subsequent arguments until it either a) runs out of arguments or b) encounters a single `;` as an argument. So these calls are all equivalent:

```fd pattern root --exec command```
```fd pattern --exec command \; root```
```fd --exec command \; pattern root```

I haven't had much time to work on this last week, but I'll try to refine this PR further over the weekend.

File: src/exec/ticket.rs (hunk 1 above)
Comment (reima) [reply src/exec/ticket.rs]: I dropped the allocation by removing `CommandTicket` altogether. There was no need for `CommandTicket` anymore, as it only collected the arguments which were then immediately used to build and execute the `Command`. Now the `TokenizedCommand` builds the `Command` with the generated arguments directly and uses `execute_command` (previously `CommandTicket::then_execute`) to execute it.

**Comment by reima** [comment]
I refactored some things to clean up and updated the documentation to reflect the new `--exec` syntax. I think this PR is now mature enough for a review.

File: src/app.rs [hunk 2]
```diff
@@ -85,7 +85,10 @@ pub fn build_app() -> App<'static, 'static> {
             arg("exec")
                 .long("exec")
                 .short("x")
-                .takes_value(true)
+                .multiple(true)
+                .min_values(1)
+                .allow_hyphen_values(true)
+                .value_terminator(";")
```
sharkdp: We should probably keep `;` as terminator here for consistency with `find` but I was shortly wondering if we should maybe change this to something that doesn't need to be escaped on the shell. The `\;` has always looked really awkward to me.

On the other hand, what ever we would choose as an alternative could not be passed to other programs...

src/app.rs
```diff
@@ -171,15 +174,17 @@ fn usage() -> HashMap<&'static str, Help> {
...
-           The following are valid tokens that can be used within the expression for generating \
-           commands:\n \
-             '{}':   places the input in the location of this token\n \
-             '{.}':  removes the extension from the input\n \
-             '{/}':  places the basename of the input\n \
-             '{//}': places the parent of the input\n \
-             '{/.}': places the basename of the input, without the extension");
+        , "Execute a command for each search result"
+        , "Execute a command for each search result.\n\
+           All arguments following -exec are are taken to be arguments to the command until the \
```
sharkdp: This should be `--exec`, right?

src/exec/command.rs
```diff
@@ -0,0 +1,98 @@
...
+            // to write it's outputs.
+            let _lock = out_perm.lock().unwrap();
+
+            let stdout = io::stdout();
+            let stderr = io::stderr();
+
+            let _ = stdout.lock().write_all(&output.stdout);
+            let _ = stderr.lock().write_all(&output.stderr);
+        }
+        Err(why) => eprintln!("fd: exec error: {}", why),
```
sharkdp: It would be great if we could improve this error message here. I'm not sure what all the possible cases are that could trigger this error, but this message will be shown to the user if the given command is not available:
```
> fd -x foo
fd: exec error: No such file or directory (os error 2)
```
I encountered this by accident and was definitely confused until I realized that the command I specified did not exist.

src/exec/command.rs
```diff
@@ -0,0 +1,98 @@
...
+            // Create a lock to ensure that this thread has exclusive access to writing.
+            let _lock = out_perm.lock().unwrap();
+
+            // And then write the outputs of the process until EOF is sent to each file.
+            let stdout = io::stdout();
+            let stderr = io::stderr();
+            let _ = io::copy(&mut pout, &mut stdout.lock());
+            let _ = io::copy(&mut perr, &mut stderr.lock());
+        }
+        Err(why) => eprintln!("fd: exec error: {}", why),
```
sharkdp: See above.

sharkdp:
@reima Thank you very much (again) - great work!

I added just some minor comments.

src/app.rs
reima: Yeah, I'm not particularly fond of `;` either. I've more than once been puzzled by odd error messages when I forgot to escape it. `:` would be an alternative that doesn't need escaping (but doesn't convey the same "end of statement" feel).

src/app.rs
reima: Right. There's also a duplicated "are".

src/exec/command.rs
reima: This is probably out of scope for this issue. While this code is all marked as new, it's actually what is left from the renamed and modified `ticket.rs`.

mmstick: You'll have to construct the error message yourself, by checking the error kind of the error.

```rust
if why.kind() == io::ErrorKind::NotFound {
    eprintln!("fd: exec error: command not found");
} else {
    eprintln!("fd: exec error: {}", why);
}
```

src/app.rs
sharkdp: Let's keep `;` for now and I'll keep this in mind. I was also thinking of `:`. Would be great if we (clap) could support both.

src/exec/command.rs
sharkdp: Thanks @mmstick. @reima Oh, I missed that. I'll do it after merging this PR.

mmstick:
Maybe also have a benchmark comparison of find versus fd with parallel execution?

sharkdp:
> Maybe also have a benchmark comparison of find versus fd with parallel execution?

Sounds great!

I imagine there are two cases where *fd* could potentially be faster:
- Compared to `find -exec`.. simply because of parallel execution
- Compared to `find .. | parallel` if the limiting factor is the filesystem (??)

Do you have any particular benchmarks in mind?

PR #169, opened 2017-11-15
Create shell completions OUT_DIR

andoriyu:
Right now it panics with OUT_DIR is not exist prior compilation. Which is not nice.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,7 @@ extern crate version_check;
 use clap::Shell;
 use std::io::{self, Write};
 use std::process::exit;
+use std::fs;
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

sharkdp:
5.0.1 was an unfortunate mistake from my side :smile: It's functionally equivalent to 5.0.0 but I had to adapt something for crates.io. I didn't want to tag it in Git because it would have triggered a completely new build with new binaries etc.

Shouldn't happen in the future.. usually all releases will be tagged. Thanks for asking.

PR #171, opened 2017-11-15
Add instructions on how to install under FreeBSD

andoriyu:
As of this morning 5.0.0 is in ports tree.


sharkdp:
Nice!! :+1:

Thank you very much for creating that package (and updating the README).

PR #172, opened 2017-11-16
Add instructions on how to install under Void Linux

maxice8:


sharkdp:
Nice! Thank you very much!

PR #173, opened 2017-11-20
bump ignore to 0.3, ansi_term to 0.10

ignatenkobrain:


sharkdp:
Hi,

Thank you for your contribution! Is there any particular reason why you would like to update these dependencies?

Currently, the build fails on Windows. This is a known bug with `ansi_term` and has already been addressed upstream (https://github.com/ogham/rust-ansi-term/issues/31) by @iology.

ignatenkobrain:
@sharkdp I'm packaging fd-find for Fedora and we try to keep only latest versions of crates as long as it's possible. I can keep those bumps as our downstream patches until someone fixes ansi-term.

sharkdp:
I see. I'm afraid this is the only option for now, as we would like to keep supporting all platforms that are tested via CI.

> I'm packaging fd-find for Fedora

Thank you very much!

sharkdp:
Both ignore as well as ansi_term have now been updated to the latest versions.

Issue #153, opened 2017-10-28
In cygwin, fd output displays backslash instead of forward slash


rajeshduggal:

sharkdp:
Hi, thank you for your feedback.

Could you please give a little bit more details (fd version, operating system, etc.)?

Why do you think *fd* should output `/` as directory separator instead of `\`? Isn't cygwin running on Windows?

rajeshduggal:
Most tools in cygwin output a / separator, but fd uses \ instead...
e.g..
$ uname -a
CYGWIN_NT-6.1 TDL00885482 2.9.0(0.318/5/3) 2017-09-12 10:18 x86_64 Cygwin
$ fd --version
fd 4.0.0
$ ls -1 dax/*
dax/oswald.txt
$ tree -if dax
dax
dax/oswald.txt
0 directories, 1 files
$ find ./dax
./dax
./dax/oswald.txt
$ fd '.*txt' ./dax
dax\oswald.txt

tmccombs:
was fd compiled with cygwin, or as a windows binary?

rajeshduggal:
https://github.com/sharkdp/fd/releases/download/v5.0.0/fd-v5.0.0-x86_64-pc-windows-gnu.zip

So I guess it's understandable that it's output shows windows backslash convention. 
(fyi. There's an interesting discussion thread on a similar cygwin path issue for a different project here.. https://github.com/BurntSushi/ripgrep/issues/269 )

sharkdp:
We could potentially change this, but I'm not sure if `/` would always be preferable over `\`. Do all tools on Windows support `/`? Is there any drawback from using `\` or is this more of an aesthetic thing?

It says `fd v4.0.0` in your output but you downloaded v5.0.0? Could you please make sure that you are on the latest verstion? There have been a few changes to path handling (on Windows) in 5.0.0 - although this will likely not fix this issue.

Doxterpepper:
I was actually curious about this. When compiled on windows fd will always replace forward slashes with back slashes. Doesn't PathBuf already format the slashes depending on the platform?

Doxterpepper:
@sharkdp, it looks like fd is using `Path::from` which will convert the path to the platform specific convention. We could maybe figure out a way to determine if we're running from cygwin when we print to the console and then change the back slashes to forward slashes then.

As a side note from what I can tell `let path = path.replace('/', "\\")` is unnecessary because of PathBuf.

rajeshduggal:
[...]

PR #176, opened 2017-11-22
Cygwin forward slashes

Doxterpepper:
#153 
@sharkdp, I was messing around with this and figured I'd go ahead and open a PR, hope you don't mind. This change adds a check for if the HOME environment variable is present at run time on windows builds. The HOME environment variable is not available in cmd so it servers as a way to differentiate between the two environments and print the appropriate slashes accordingly.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -15,6 +15,8 @@ use std::ops::Deref;
 use std::path::{self, Path, PathBuf, Component};
 #[cfg(any(unix, target_os = "redox"))]
 use std::os::unix::fs::PermissionsExt;
+#[cfg(windows)]
+use std::env;
 
 use ansi_term;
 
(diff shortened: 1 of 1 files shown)


sharkdp:
> I was messing around with this and figured I'd go ahead and open a PR, hope you don't mind.

Absolutely not. Thank you very much for your contribution.

---

Before going forward with this, I'd like to have the questions in #153 answered (also the ones by @ghuls), as it's not clear to me that we actually need this.

Concerning the code changes, I'm not too keen about having so many `#[cfg(...)]` special cases for different OSes, but maybe that's just me. Also, the check for the `$HOME` variable does seem a little bit fragile. Doesn't cygwin provide something like `$CYGWIN_VERSION` or similar?

PR #181, opened 2017-11-29
Refactoring main.rs : remove unused variable (empty_pattern)

LoveTube:
I think this PR is  just small code improvement , so I'd go ahead and open a PR, hope you don't mind. 
In main.rs, empty_pattern is used as a default value for pattern.
I think it's better remove empty_pattern and add "" for default pattern. Because it's more clear.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,8 +48,7 @@ fn main() {
     let matches = app::build_app().get_matches();
 
     // Get the search pattern
-    let empty_pattern = String::new();
-    let pattern = matches.value_of("pattern").unwrap_or(&empty_pattern);
+    let pattern = matches.value_of("pattern").unwrap_or("");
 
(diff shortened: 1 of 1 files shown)


sharkdp:
@LoveTube Welcome and Thank you for your contribution!

I believe this was needed in some older version of Rust (although not sure). Great that we can write this in a clearer way now.

LoveTube:
OK. That's why this code is exist. I close this PR. Thanks

sharkdp:
Sorry, I was unclear: I think we should integrate this now (because we have bumped the minimum required Rust version from 1.16 to 1.19 recently).

LoveTube:
@sharkdp I just read your comment, I think you're right.

sharkdp:
Thanks,

Issue #166, opened 2017-11-12
feature request: allow specifying multiple search paths 


rfourquet:
I.e. `fd [pattern] [path [path ...]]`. I don't foresee a need to specify a 3rd argument to `fd` which would not be a search path, so allowing multiple paths seems harmless and useful. Also, by compatibility with `find`, would there be a problem to not error if the provided paths are files (i.e. not directories)?

rfourquet:
For example, here is my today's use case: I'm looking for the presence of a certain .so library file in  the dropbox versions present on my nixos system; I would like to simply do:
 `fd libGL.so.1 /nix/store/*dropbox*`.
In particular it would be useful that the files in the paths list (expanded from `*dropbox*`) do not generate errors.

for-coursera:
Second that request. Passing a listing from several directories to `fzf` would be extremely handy. `ripgrep` already manages that, but `fd` is faster, isn't it? :)

Doxterpepper:
@sharkdp, mind if I take this on?

sharkdp:
@Doxterpepper That'd be great!

ChengCat:
I'm not sure whether this is a good idea. If `/nix/store/*dropbox*` expands to an empty string, you could be surprised, as fd will be searching inside the current directory. One could assume `fd pattern (empty path list)` to mean no path searched, but it contradicts with the current design.

BTW, NixOS use symlinks extensively, so you might need `-F` sometimes. (#180)

rfourquet:
This is probably a matter of pragmatism. 

> If /nix/store/\*dropbox\* expands to an empty string, you could be surprised,

You are right, but I needed this on the command line, after knowing that the expansion is not empty. Then the alternative I had to resort to was no better in this respect (I don't remember if I used `find` directly, or `fd` with a loop over). `fd` with multiple paths would simply have been more convenient.

ChengCat:
This is a matter of design, and the intention is to reduce potential surprises or pitfalls. It is especially important for a tool intended to replace a fundamental utility like `find`. I understand that it is convenient for this particular use case, but you probably agree that, every time you write `fd pattern pathWith*` in a shell script, you need to consider that `pathWith*` could possibly expand to an empty string.

I'd suggest a new syntax for this: `fd pattern -`. This reads stdin for a path list to be searched.

Doxterpepper:
[...]

PR #182, opened 2017-11-29
Add multiple path support

Doxterpepper:
# Multiple path support
#166 
## Changes

- `fd` allows multiple paths from the command line
- main has been refactored to handle multiple paths
  - No longer makes use of path_display, path_display was never actually used and has been removed
  - Removed redundant logic for absolute paths. `fshelper::absolute_path` does nothing if the path provided is already an absolute path. The only thing that needs to be checked is if the absolute-path flag has been provided.
- `scan` now takes a reference to a vector of `PathBuf`
- Added integration test for multiple files
- Modified `TestEnv` to allow custom directory listings
- Previous hard coded directory structure has been added to tests.rs in the form of static arrays containing each file and directory to be created. This allows for alternative directory structures in integration tests. The symlink is still always created for now.


sharkdp:
Thank you very much. Unfortunately, it will be a few days before I can review this.

Also, not sure what's wrong with Travis. Maybe a caching problem.

Cargo.lock
```diff
@@ -14,6 +14,7 @@ dependencies = [
  "num_cpus 1.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-syntax 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "shell-escape 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
```
sharkdp: Why does this PR re-introduce shell-escape? Is this needed somewhere?

src/main.rs
```diff
@@ -57,28 +57,31 @@ fn main() {
...
-        Some(path) => PathBuf::from(path),
-        None => current_dir.to_path_buf(),
+    //Get one or more root directories to search.
+    let mut dir_vec: Vec<_> = match matches.values_of("path") {
+        Some(paths) => {
+            paths
+                .map(|path| {
+                    let path_buffer = PathBuf::from(path);
+                    if !fshelper::is_dir(&path_buffer) {
+                        error(&format!(
```
sharkdp: There was a request somewhere in the open issue that we also allow for files to be specified here (that would be ignored), but I think it's a good idea to explicitely show errors here. The original problem could be solved by using `drop*/` globs instead of `drop*`, I guess. So :+1:

src/walk.rs
```diff
@@ -48,11 +48,12 @@ pub enum FileType {
 /// If the `--exec` argument was supplied, this will create a thread pool for executing
 /// jobs in parallel from a given command line and the discovered paths. Otherwise, each
 /// path will simply be written to standard output.
-pub fn scan(root: &Path, pattern: Arc<Regex>, config: Arc<FdOptions>) {
+pub fn scan(path_vec: &mut Vec<PathBuf>, pattern: Arc<Regex>, config: Arc<FdOptions>) {
```
sharkdp: Using a mutable reference here seems strange to me. Could `scan` not be implemented without using `pop`?

src/walk.rs
```diff
@@ -48,11 +48,12 @@ pub enum FileType {
...
+    let first_path_buf = path_vec.pop().unwrap();
```
sharkdp: Assuming that there is one element in the vector is fine for now. It would be even better to use the type system for that kind of guarantee (e.g. using a non-empty list type).

src/walk.rs
```diff
@@ -178,7 +181,7 @@ pub fn scan(root: &Path, pattern: Arc<Regex>, config: Arc<FdOptions>) {
 
             let entry_path = entry.path();
 
-            if entry_path == root {
+            if entry.depth() == 0 {
```
sharkdp: I like the change, but does this change something semantically?

tests/tests.rs
```diff
@@ -64,10 +78,32 @@ fn test_simple() {
     );
 }
 
+/// Test multiple directory searches
+#[test]
+fn test_multi_file() {
+    let dirs = &["test1", "test2"];
+    let files = &["test1/a.foo", "test1/b.foo", "test2/a.foo"];
+    let te = TestEnv::new(dirs, files);
+    te.assert_output(
```
sharkdp: Maybe we could also add a test for `["a.foo", "test1"]` here to make sure that `test2/a.foo` is *not* found in this case.

Cargo.lock
Doxterpepper: I don't actually remember adding that. It's not used for anything I'll remove that.

src/walk.rs
Doxterpepper: This shouldn't change anything in practice, this is logically equivalent to the previous if statement. We are just using a different property of root paths to determine this is a root path. The root path is always a depth of zero so we can use that instead checking if entry shares a name with one of the provided paths.

tests/tests.rs
Doxterpepper: Yea that certainly can be done, I'll add that tonight. If you think of any other tests that should be included let me know.

src/walk.rs
Doxterpepper: Is there such a data structure native to rust? A non-empty vector is a precondition to scan so we could just include an assert statement to ensure that condition.

sharkdp: I don't think so. There seem to be some nonempty crates, but if this is the only occurrence, I think your solution is fine. Instead of using an assert, we could replace `unwrap` with `expect` and a custom error message.

tests/tests.rs
sharkdp: No, the rest looks great

src/main.rs
Doxterpepper: Accidentally killing two birds with one stone, glad I could help 😄

sharkdp:
Thank you for the changes!

Could you please run `cargo fmt` in order for the Travis to pass?

Doxterpepper:
Sure thing

rfourquet:
That's great, thanks for your work!

src/main.rs
rfourquet: > The original problem could be solved by using drop*/ globs 

Thanks for this trick! If I come up with a use case for accepting also files, I will open a new issue.

PR #184, opened 2017-11-30
bump lazy_static to 1

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


Cargo.lock
```diff
@@ -254,6 +260,11 @@ dependencies = [
  "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
```
tmccombs: Why does this ad d the shell-escape package?

ignatenkobrain: to be honest, I don't know. I just run `cargo update -p lazy_static`

sharkdp:
Is there any specific reason why you want to update this dependency?

ignatenkobrain:
There are several:

1) It is first stable version (1.0), so no more unexpected breakages after version bump
2) I package fd for Fedora and we try to not keep old versions of crates

sharkdp:
Okay, thanks.

Could you please rebase this on top of master. This should hopefully make Travis pass.

sharkdp:
I've included this change in a recent update to Cargo.toml on master, so this is not needed anymore. Thanks!

PR #185, opened 2017-12-02
The -i flag allows a package to be installed without updating.

Doxterpepper:
The -i flag allows installation of the package when it's not installed. Otherwise cargo-update will just not update the package since it has to be installed to update it.

From the help page for cargo-update:
`    -i, --allow-no-update    Allow for fresh-installing packages `


sharkdp:
Thank you very much!

Issue #178, opened 2017-11-25
hang with exec


aeosynth:
```
sudo touch foo
fd foo -x rm {}
```

attempting to remove a write-protected file hangs

sharkdp:
Thank you for reporting this.

I believe this is due to `rm` waiting for input:
``` bash
▶ sudo touch foo

▶ rm foo
rm: remove write-protected regular empty file 'foo'? 
```

Indeed, if you enter `y`, *fd* terminates:
``` bash
▶ fd foo -x rm {}
y
rm: remove write-protected regular empty file 'foo'? %                                
```

Still, this is a bug because *fd* doesn't show the output of `rm`.

ChengCat:
`rm` is waiting for inputs from stdin. We could resolve this issue, if stdin is closed before executing `rm`.

PR #186, opened 2017-12-03
Show stderr on exec with `rm -i` like `find`

spk:
Hi this fix issue reported in https://github.com/sharkdp/fd/issues/178, cheers

### DIFF
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -45,7 +45,7 @@ pub fn execute_command(mut cmd: Command, out_perm: Arc<Mutex<()>>) {
 /// Executes a command.
 #[cfg(all(unix))]
 pub fn execute_command(mut cmd: Command, out_perm: Arc<Mutex<()>>) {
-    use libc::{close, dup2, pipe, STDERR_FILENO, STDOUT_FILENO};
+    use libc::{close, dup2, pipe, STDOUT_FILENO};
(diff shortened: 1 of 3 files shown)


sharkdp:
Welcome and thank you very much for your contribution!

I somehow don't like that we have to special-case the tests for Travis CI. Could we write a simpler test without using `rm`? We would somehow need a simple program that writes a well-defined message to stderr.

iology:
This patch just solves part of the problem. What if (with another program) the info is printed to stdout?

spk:
Thanks, I made a fix in https://github.com/sharkdp/fd/pull/186/commits/85f9c7f28fc4618331647801b8a51b9671fb07fa (hopefully CI will be green)
I don't know other use case with stdin and asking an input since `read` redirect output to stderr also
Cheers

reima:
This change side steps the issue by printing stderr directly to the console. This will intersperse the outputs from different command invocations instead of keeping the output of each single invocation together like before. (stdin is still captured and only printed after the command exits.)

Another solution was already proposed in #178 by ChengCat: close the command's stdin, so it can never block on input from stdin. I'll create a PR with this change shortly.

sharkdp:
@spk Due to the points raised by @reima, I'm going to close this in favor of #193. Nevertheless, thank you very much for your contribution!

Issue #174, opened 2017-11-20
fd package for Ubuntu


sharkdp:
It would be great if we had an easy way to install *fd* on Ubuntu (or maybe even provide a package for all Debian-based distributions?).

I'm not familiar with creating own `.deb` packages and the ecosystem around PPAs, so any help would be greatly appreciated!

danielpclark:
`.deb` packages are very involved.  Ubuntu supports a much easier to make lightweight package installation through [snap](https://docs.snapcraft.io/build-snaps/).

tmccombs:
I could help out with this

sharkdp:
@danielpclark Never heard of snap, but if that's something that's easily available (also on older Ubuntu versions?), that might be an alternative option.

@tmccombs Sounds great!

sharkdp:
Just for reference, thanks to @tmccombs help, we have `*.deb` packages for the v6.1.0 release (and every forthcoming release).

What would be logical next steps to make this even more convenient for Ubuntu users?

Set up a PPA?

As a long term goal, [get into the official repositories](https://askubuntu.com/questions/16446/how-to-get-my-software-into-ubuntu)? :smile:

In any case, I should add a new "Ubuntu" section in the README install instructions.

wsandin:
Would you be interested in a RPM, FreeBSD package or similar too?

sharkdp:
> Would you be interested in a RPM, FreeBSD package or similar too?

- On FreeBSD, *fd* can be installed via portmaster already (see README) - is there something else that is needed?
- An RPM package would be great!

matrixik:
Snaps are build automatically
https://docs.snapcraft.io/build-snaps/ci-integration
`ripgrep` config for it: https://github.com/BurntSushi/ripgrep/blob/master/snapcraft.yaml

Kristinita:
**Type: Reply** 💬

It would be nice to be able to simply and without compilation install on Ubuntu the latest fd version like this:

```text
snap install fd-find
```

1. For Ubuntu Jammy Jellyfish ([**the latest Travis CI environment as of January 2024**](https://docs.travis-ci.com/user/reference/linux#overview)) [**the latest apt fd-find version is 8.3.1**](https://packages.ubuntu.com/jammy/fd-find), not 9.0.0.
1. When installing from binaries, the [**user must manually specify the version number and architecture**](https://github.com/sharkdp/fd#on-ubuntu).
1. Compilation from sources isn’t a “simply installation”.

Thanks.

PR #187, opened 2017-12-05
Deb

tmccombs:
This works. But I'm wondering if I should move the installation to the target directory into a `make install` make target that could be reused by other package formats (such as archlinux)


tmccombs:
This is  a partial solution for #174

sharkdp:
Thank you very much!

I think this approach is good for now. We have pre-built packages for Arch anyways.

PR #193, opened 2017-12-09
Unify execute_command across platforms

### LINKED ISSUE #178: hang with exec

reima:
Fixes #178 by proving an empty stdin, which leads to the command not blocking on input.

To achieve this, I removed the `#[cfg(all(unix))]` implementation of `execute_command` in favor of the simpler `#[cfg(not(unix))]` implementation (which is now used unconditionally). This implementation uses `Command::output`, which by default captures stdout and stderr and provides an empty stdin.

Incidentally, this also fixes an (unreported) bug in the removed `execute_command` implementation: when the command output grew larger than the size of a pipe, a deadlock occured:
* The command was blocked on writing to the filled pipe.
* _fd_ was blocked on waiting for the command to finish.

I'm not entirely sure why we even had the removed `execute_command` to begin with, as using the `Command`'s built-in output capture features seem to work just fine on Unix configurations.

### DIFF
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -9,21 +9,18 @@
 use std::process::Command;
 use std::sync::{Arc, Mutex};
 use std::io;
+use std::io::Write;
 
 /// Executes a command.
-#[cfg(not(unix))]
 pub fn execute_command(mut cmd: Command, out_perm: Arc<Mutex<()>>) {
-    use std::process::Stdio;
-    use std::io::Write;
-
     // Spawn the supplied command.
-    let output = cmd.stdout(Stdio::piped()).stderr(Stdio::piped()).output();
+    let output = cmd.output();
 
     // Then wait for the command to exit, if it was spawned.
     match output {
         Ok(output) => {
             // While this lock is active, this thread will be the only thread allowed
-            // to write it's outputs.
+            // to write its outputs.
             let _lock = out_perm.lock().unwrap();
 
             let stdout = io::stdout();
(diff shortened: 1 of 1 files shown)


iology:
@mmstick /cc

I also tried using `Command::output`, but it doen't capture anything on my mac. So, that separate implementation is a fix for unixes, not RedoxOS.

iology:
Ah, sorry, I said it wrong. It does capture output but there seemed to be something wrong with stdin. I don't remember.

mmstick:
Is `STDIN_FILENO` being closed in the child before executing the command?

iology:
Yes, the stdin is closed. (checked with libc::isatty in `.before_exec()`)

But the doc [said] Command::output: By default, **stdin**, stdout and stderr are captured (and used to provide the resulting output).

[said]: https://doc.rust-lang.org/src/std/process.rs.html#636-637

reima:
@mmstick Yes. Technically [std::process::Command::output](https://github.com/rust-lang/rust/blob/1.22.1/src/libstd/process.rs#L655) excutes the command first with the command's stdin set to `/dev/null` (`null:` on Redox, `NUL` on Windows), and then closes the command's stdin in [std::process::Child::wait_with_output](https://github.com/rust-lang/rust/blob/1.22.1/src/libstd/process.rs#L1229).

sharkdp:
Thank you very much!

iology:
This issue confirm the behavior of `Command::output`: https://github.com/rust-lang/rust/pull/46220

PR #197, opened 2017-12-11
Fix bash_completion path in deb.

### LINKED ISSUE #195: Bash completion not working with deb package
I installed the new Debian package (fd-musl_6.1.0_amd64.deb) on Ubuntu 17.10 and bash completion is not working.

I'm *definitely* no expert in this, but based on other files on my system, what I tried and seems to work was to move the `/usr/share/bash_completion/completions/fd/fd.bash-completion` file to `/usr/share/bash-completion/completions/fd`.

(It also used to work for me when I manually moved the `fd.bash-completion` file to `/etc/bash_completion.d`.)

tmccombs:
Fixes #195

### DIFF
--- a/ci/before_deploy.bash
+++ b/ci/before_deploy.bash
@@ -85,7 +85,7 @@ make_deb() {
     install -Dm644 LICENSE-APACHE "$tempdir/usr/share/doc/$PROJECT_NAME/LICENSE-APACHE"
 
     # completions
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you for fixing this. There was a related change that was already included in 6.1.0, but somehow I managed to forget to update the master branch. It would be great if you could resolve the conflicts and rebase this (tests should also be fine then).

Issue #201, opened 2017-12-17
ctrl-c when using colored output gives exit code 0 instead of 130


Doxterpepper:
When `fd` recieves `SIGINT` its exit code is 0 instead of the standard 130. Changing the exit code will fix it.

sharkdp:
Closed via #202

PR #202, opened 2017-12-17
Correct exit codes

Doxterpepper:
#201 

Changing the exit code for ctrl-c to 130 and the code when printing to stdout to 1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,24 +1,3 @@
-[root]
-name = "fd-find"
-version = "6.1.0"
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you very much!

Could we maybe introduce constants here to avoid those "magic numbers"? Like `EXITCODE_ERROR = 1` and `EXITCODE_TERMINATED_BY_SIGINT = 130`?

sharkdp:
Thanks!

Issue #199, opened 2017-12-13
Search multiple extensions


erf:
Would it be an idea to add support for searching multiple extensions?

E.g. if i want to find both `.h` and `.c` files in a folder: `fd -e h -e c`

Or should i use regex patterns?

sharkdp:
Interesting idea. This could actually have a similar implementation as #177.

sergeevabc:
So how do I find _audio files_ (*.mp3|m4a|ogg|opus|flac) with the word _love_ in their name?

tavianator:
> So how do I find _audio files_ (*.mp3|m4a|ogg|opus|flac) with the word _love_ in their name?

```console
$ fd 'love.*\.(mp3|m4a|ogg|opus|flac)$'
media/audio/music/Al_Green/Love_And_Happiness.mp3
...
```

user-1-2-3-4-5:
Is using a regex pattern the only supported method? Being able to repeat the extension would help the more novice users:

`fd --extension flac --extension mp3 --extension aac --extension m4a --extension ogg --extension wav --extension wave`

 … would be truncated to:

`fd --extension flac mp3 aac m4a ogg wav wave`

---

I prefer using the long-form flags with most CLI software.

tmccombs:
you can do `fd --extension flac --extension mp3 --extension aac --extension m4a --extension ogg --extension wav --extension wave`

Issue #177, opened 2017-11-22
Improve filtering by filetype (exclude a filetype, or allow multiple filetypes to be specified)


rlue:
Currently, `fd` offers the `-t` command line option to view only files of a single type (regular files, directories, symbolic links). 

However, there is no way (that I can see) to view all files _except_ those of a single type (in other words, to exclude a given filetype from the results). Perhaps a syntax like `+td` or `-Td` to exclude directories, or `-tfl` to do the same?

Hopefully, I've missed something, and this feature is already available!

### Workaround

The best I can think to do is

```
$ rg -FNxvf <(fd -td) <(fd)
```

* `rg <(fd)` searches the output of fd
* `-f <(fd -td)` searches against patterns contained in the "file" <(fd -td) (the output of that command), rather than patterns given on the command line
* `-F` treats search patterns as fixed strings (not regular expressions)
* `-x` requires a whole like match (like wrapping each line in ^$)
* `-v` inverts the results (gives non-matches, rather than matches).
* `-N` suppresses line numbers

sharkdp:
Thank you very much for your feedback.

> Hopefully, I've missed something, and this feature is already available!

Unfortunately - no. However, consider the following:

```
▶ tree
.
├── dir
├── file
├── symlink_to_dir -> dir
└── symlink_to_file -> file
```

* We can search for `dir` only (`-td`)
* We can search for `file only (`-tf`)
* We can search for both symlinks (`-tl`)
* We can also search for `dir` and `symlink_to_dir` (`-L -td`)
* as well as for `file` and `symlink_to_file` (`-L -tf`)

What is missing is:
1. search for `dir` and both symlinks
2. search for `file` and both symlinks
3. search for `dir` and `file`

For now, I would propose to allow for multiple occurrences of the `--type`/`-t` option in a union-like fashion ("or" instead of "and"). Since there are only three types of entries, I think this would still be acceptable (having to specify `--type file --type directory` instead of `--exclude-type symlink`).

I'm open for other ideas, though. What do you think?

septatrix:
> For now, I would propose to allow for multiple occurrences of the `--type`/`-t` option in a union-like fashion ("or" instead of "and"). Since there are only three types of entries, I think this would still be acceptable (having to specify `--type file --type directory` instead of `--exclude-type symlink`).

Now there are 7+2 file types (the latter two being the executable and empty dir). A negated filetype seems like a reasonable addition now. Should I create a new issue for that?

[...]

PR #205, opened 2017-12-30
Added support for filtering by multiple filetypes and extensions

tkadur:
#177 #199

### Changes 
You can now specify multiple extension and file type filters by simply adding more `-e` or `-t` options, respectively, as was suggested in the above issues. 

So, to find all files and symlinks with the extension `.c` or `.h`, you could run

```
fd -ec -eh -tf -tl
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -175,13 +175,14 @@ fn usage() -> HashMap<&'static str, Help> {
            on the search depth.");
     doc!(h, "file-type"
         , "Filter by type: f(ile), d(irectory), (sym)l(ink)"
-        , "Filter the search by type:\n  \
+        , "Filter the search by type (multiple allowable filetypes can be specified):\n  \
              'f' or 'file':         regular files\n  \
              'd' or 'directory':    directories\n  \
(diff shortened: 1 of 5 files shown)


sharkdp:
@tkadur Welcome and thank you very much for your contribution!

It would be great if we could also:

- Update the tests (add a tests for multiple file types and for multiple file extensions)
- Update the long `--help` text (describe what happens if multiple options are specified)
- Update the man page in the same way.

sharkdp:
Also, there seems to be a problem with additional pattern matching.

With this PR:
```
▶ touch test.txt
▶ fd -e txt foo
test.txt
```

Previously:
```
▶ fd -e txt foo
```
(no match)

tkadur:
I updated the help text and tests and also fixed the issue you mentioned. I'm not quite sure what you mean by updating the man page.

sharkdp:
Thank you very much!

The [man page](https://en.wikipedia.org/wiki/Man_page) is the file in `doc/fd.1` - I'll simply update this later before releasing the next version.

PR #206, opened 2017-12-30
Routine dependencies update

kirushik:
It's a trivial task, unless obsoleteness accumulates.

The only actual change I required here was `winapi` 0.2 → 0.3 upgrade.
They've split the API into different namespaces and hid them behind feature flags. As a side-effect, explicit dependency on `kernel32-sys` is no longer required.

This PR is tested and verified to work on both Windows and Linux.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,23 +8,17 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
-name = "ansi_term"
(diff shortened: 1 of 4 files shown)


kirushik:
Hmmm, looks like there's a build issue on msvc stable toolchain in Appveyor. Will investigate.

kirushik:
Yep, can reproduce on the same same toolchain on my local machine.
Will reopen this when the failure will be fixed.

sharkdp:

I believe the build issues are due to a known bug with ansi_term (see ogham/rust-ansi-term#31). This has popped up in a few recent PRs too.

PR #207, opened 2017-12-30
Routine dependency update

kirushik:
This is #206 but without the `winapi` part.

Looks like linker errors in [Appveyor](https://ci.appveyor.com/project/sharkdp/fd/build/1.0.330/job/y7av4ojwrjdsf9fu) are related to retep998/winapi-rs#411, but I'm not proficient enough in Windows cross-compiling to solve the issue on my own.


kirushik:
OK, so it's not a `winapi`-specific issue, but more general breakage in cross-compiling pipeline on windows.

I'm not sure then that the time spent on narrowing it down will be a time well spent; sorry for bothering you with all this version-update nonsence.

PR #209, opened 2018-01-02
Fix shell completion directories for Ubuntu

sharkdp:
@tmccombs Just fyi.

Where did you get the original paths from? None of them seemed to work for me on Ubuntu (14.04).


tmccombs:
I have them on my 16.04 box, but looking at it again, /usr/share/fish/vendor_completions.d is empty and /usr/share/zsh/site-functions/ only has a single file.  This looks good to me.

Issue #210, opened 2018-01-02
Killing 'fd' during long-running search takes long


sharkdp:
If there are only very little search results and I press `Ctrl-C`, *fd* does not stop. The process only seems to exit when the next search result is found (and new output is generated).

This can be tested with
``` bash
> fd --follow non-existing-pattern /
# try to kill the process with Ctrl-C
```

Doxterpepper:
Yea this is kinda a big bug, if fd searches a large enough directory and only one item is to be found it will never exit via ctrl-c until it finds that one entry. That could be a really long time and may be a pretty common use case. To fix this we need to exit from the sender thread gracefully as well when we handle ctrl-c.

PR #211, opened 2018-01-03
Using wants_to_quit to exit the sender as well as the reciever.

Doxterpepper:
#210 
Adding a conditional for wants_to_quit to the sender thread allows us to exit the sender thread as well as the receiver thread when ctrl-c is received.


sharkdp:
That was quick :smile:

Thank you so much!

Doxterpepper:
No problem! I sorta broke it so I figured I should probably fix it :upside_down:

Issue #214, opened 2018-01-08
(Feature request) Support extensions with multiple suffixes


althonos:
Hi there,
currently, `fd` does not support extensions with multiple suffixes in them, like `tar.gz` and the like. Of course, one could simply run `fd tar.gz`, but unfortunately this will also find some other files (such as `*.tar.gz.sig`). The correct workaround would be `fd tar.gz\$`, which starts to get cumbersome !

I could probably have a look at the code myself, since I don't think this should be too complicated to implement, but I'd like to have your opinion about it.

sharkdp:
I would have actually needed this just yesterday. Completely agreed, this would be nice to have.

I think it's not *completely* trivial to implement this because we currently extract the extension via 
[`Path::extension`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.extension) - which just extracts everything after the final `.` - and match that against the pattern. I guess we would have to resort to something like [`str::ends_with`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.ends_with).

sharkdp:
Closed via #215.

PR #215, opened 2018-01-08
Implement multiple suffixes extension support with tests (#214)

althonos:
Instead of relying on `Path.extension()`, the path is checked to see if it ends with any of the given extensions. The only quirk was to compare `.ext` to the end of the path instead of `ext`, or else we could have matched wrong files.

I had several ways of doing so in mind, and while not being the *quicket* this one should be the most obvious, and I don't think it's really necessary to micro-optimize here.

I also added a few test cases that should prevent regressions.


sharkdp:
Welcome and thank you very much for your contribution!

> I had several ways of doing so in mind, and while not being the quicket this one should be the most obvious, and I don't think it's really necessary to micro-optimize here.

Unfortunately, I think it will be. I have been running some benchmarks and it looks like matching on file extensions is about a factor of two slower with this PR:

Before:
```
benchmarking bench/fd -HI -e jpg '' '/home/shark'
time                 834.3 ms   (737.0 ms .. 878.1 ms, ci 0.990)
                     0.998 R²   (0.996 R² .. 1.000 R², ci 0.990)
mean                 861.9 ms   (839.3 ms .. 873.2 ms, ci 0.990)
std dev              19.55 ms   (0.0 s .. 19.68 ms, ci 0.990)
variance introduced by outliers: 19% (moderately inflated)
```

After (this PR):
```
benchmarking bench/fd -HI -e jpg '' '/home/shark'
time                 1.777 s    (1.704 s .. 1.840 s, ci 0.990)
                     1.000 R²   (1.000 R² .. 1.000 R², ci 0.990)
mean                 1.757 s    (1.740 s .. 1.776 s, ci 0.990)
std dev              17.99 ms   (0.0 s .. 20.65 ms, ci 0.990)
variance introduced by outliers: 19% (moderately inflated)
```

althonos:
@sharkdp : I just thought of something that should work better. Don't merge yet.

sharkdp:
Thanks!

Unfortunately, this is still much slower than the original version:
```
benchmarking bench/fd -HI -e jpg '' '/home/shark'
time                 1.681 s    (1.667 s .. 1.701 s, ci 0.990)
                     1.000 R²   (1.000 R² .. 1.000 R², ci 0.990)
mean                 1.683 s    (1.679 s .. 1.685 s, ci 0.990)
std dev              3.412 ms   (0.0 s .. 3.603 ms, ci 0.990)
variance introduced by outliers: 19% (moderately inflated)
```

It seems like the expensive part is the `.to_string_lossy().to_lowercase()` conversion that needs to be done for each and every path(!). I found a way to make it faster by only selecting the file name first:
``` rust
if let Some(os_path) = entry_path
    .file_name()
    .map(|s| s.to_string_lossy().to_lowercase())
{
    if !filter_exts.iter().any(|x| os_path.ends_with(x)) {
        return ignore::WalkState::Continue;
    }
}
```
.. but it's still not ideal:
```
benchmarking bench/fd -HI -e jpg '' '/home/shark'
time                 926.8 ms   (NaN s .. 945.7 ms, ci 0.990)
                     1.000 R²   (1.000 R² .. 1.000 R², ci 0.990)
mean                 934.5 ms   (932.4 ms .. 936.8 ms, ci 0.990)
std dev              2.233 ms   (0.0 s .. 2.568 ms, ci 0.990)
variance introduced by outliers: 19% (moderately inflated)
```

Any more ideas? :smile:

althonos:
@sharkdp : I may have an idea that avoids using `to_string_lossy` and uses only `extension`...

src/utils/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>
+// or the MIT license <LICENSE-MIT or http://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+use std::path::Path;
+use std::iter::Iterator;
+use std::ascii::AsciiExt;
```
sharkdp: This import seems to be unused.

src/utils/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+            let utf16_vec: Vec<u16> = os_str.encode_wide().collect();
+            exts.into_iter().any(|x| {
+                decode_utf16(utf16::reverse_iter(utf16_vec.iter()))
+                    .zip(x.chars().rev())
+                    .all(|(a, b)| a.map(|c| c.eq_ignore_ascii_case(&b)).unwrap_or(false))
+            })
+        } else {
+            false
+        }
+    }
```
sharkdp: What if we have neither `unix` nor `windows` (like `redox`*)? Could we have a fallback solution for this?

Or would the `unix` solution work for `redox`?

sharkdp:
Wow, impressive work! :fireworks: 

This is indeed faster than before. My only concern is that this increases the code complexity quite a bit :smile:.

I have added a few inline comments.

Also, sorry for the late reply. I have been working on a way to perform benchmarks for `fd` in an easier way. I wrote a new tool - [hyperfine](https://github.com/sharkdp/hyperfine) - that can perform benchmarks with disk-cache-warmups (or diskcache-reset commands between the benchmarks). If anyone is interested, I've uploaded my benchmarking scripts for `fd` here: https://github.com/sharkdp/fd-benchmarks

src/utils/mod.rs
althonos: Unfortunately, this is needed for Rust `1.1.20` (see [this build](https://travis-ci.org/sharkdp/fd/jobs/328591730)) because `u8` and `u16` are missing the `eq_ignore_ascii_case` method.

And when it rains, it pours: there is no `cfg` attribute depending on Rust version.

src/utils/mod.rs
althonos: `unix` / `windows` belong to the `target_family` attribute, and `unix` includes `redox` as well. 

But I guess there could be a default implementation that still uses `to_string_lossy().to_lowercase().ends_with(ext)`, just in case.

althonos:
> Wow, impressive work! 🎆

I'm the one who should be saying that ! I'm messing up a bit with a PR, you end writing a complete tool from scratch 😅 

> My only concern is that this increases the code complexity quite a bit 😄.

I agree, the Windows part in particular was not so easy because of [`EncodeWide`](https://doc.rust-lang.org/std/os/windows/ffi/struct.EncodeWide.html) not being a `DoubleEndedIterator` (which it should be IMO). So maybe this could be reported to the Rust team.

In the end, this is just a dirty way of implementing `Into<&OsStr>.ends_with(Into<&OsStr>)` (now that I'm writing it, I realize it could be a lot cleaner to do that with a custom trait).

src/utils/mod.rs
sharkdp: Do you know when this was changed? I'd be okay with bumping the minimum-required Rust version.

src/utils/utf16.rs
```diff
@@ -0,0 +1,67 @@
...
+    type Item = u16;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(next) = self.nx {
+            self.nx = None;
+            Some(next)
+        } else {
+            if let Some(chr) = self.it.next() {
+                if let Some(next_chr) = self.it.next() {
+                    if (next_chr >= &&0xD800u16) && (next_chr <= &&0xDFFFu16) {
```
sharkdp: Could you add a short comment here (I have no idea what is going on here :smile:)?

tests/tests.rs
```diff
@@ -525,6 +525,18 @@ fn test_extension() {
...
+
+    te2.assert_output(
+        &["--extension", ".bar.baz"],
+        "spam.bar.baz
+        egg.bar.baz",
+    );
+
+    te2.assert_output(&["--extension", "sig"], "yolk.bar.baz.sig");
+
+    te2.assert_output(&["--extension", "bar.baz.sig"], "yolk.bar.baz.sig");
```
sharkdp: Could we add a test to check if everything is still Unicode-compatible? Maybe a file with a non-ASCII file-extension.

src/utils/mod.rs
sharkdp: Oh, I see. Thank you for the clarification.

In this case, I guess we can leave it as is.

src/utils/mod.rs
althonos: The import is unneeded since `1.1.23`.

tests/tests.rs
althonos: Done with acute `e` and a smiley face.

src/utils/utf16.rs
althonos: This is an implementation of a reverse iterator over a valid UTF-16 sequence. Since UTF-16 encodes some characters on 2 words, it's not possible to simply reverse the sequence, as some word pairs must be kept in the same order. I added a small comment about that.

src/utils/mod.rs
sharkdp: Thanks for tracking this down. In this case, could we please change this to
``` rust

**Comment by sharkdp** [comment]
Fantastic, thank you very much.

I'm just waiting for the Mac builds to finish (which might take some - there seems to be a maintenance downtime for Mac builds).

**Comment by sharkdp** [comment]
Looks like the new Unicode test actually fails on macOS.

**Comment by sharkdp** [comment]
Also, I just noticed that there is a change in behavior concerning hidden files / dotfiles:

Before:
```  bash
> touch .hidden
> fd -He hidden
# no match
```

After:
```
> touch .hidden
> fd -He hidden
.hidden
```

sharkdp:
I think the Unicode test failing on macOS could be due to https://stackoverflow.com/a/6153713/704831

althonos:
@sharkdp : in the end, I think we could simply:
* use the current implementation on Linux (+ non MacOS UNIXes ?)
* use a fallback with `to_string_lossy` on the filename for MacOS + Windows (I don't think the reverse UTF-16 iterator is that efficient actually)

Maybe we could also use [`to_str`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.to_str) when possible, to avoid doing a copy with valid Unicode path names.

src/utils/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use std::path::Path;
+use std::iter::Iterator;
+
+/// Determine if an os string ends with any of the given extensions (case insensitive).
+pub fn path_has_any_extension<'a, I>(path: &Path, exts: I) -> bool
+where
+    I: 'a + IntoIterator<Item = &'a String>,
+{
+    _path_has_any_extension(path, exts)
```
Doxterpepper: Couldn't we just define two `path_has_any_extension` methods but where one has the `#[cfg(all(unix, not(target_os = "macos"))]` and the other has `#[cfg(any(target_os = "macos", windows))]` rather than wrapping a single platform specific method that does everything?

althonos: This avoids duplicating the documentation comment for the two methods I guess :smile:

Doxterpepper: That's good enough a reason for me

sharkdp:
@althonos Thank you for the updates. It still seems to fail on macOS, though.

althonos:
I'm having trouble even understanding what's the issue in the MacOS build on Travis. Since I don't have a Mac to code on, this is getting more and more complicated for me to test.

At the beginning this PR was just supposed to bring multiple extensions comparison to `fd`, but now I end up porting the whole code to support UTF-8, even though I don't think it even works as intended for the program on MacOS in its current state.

sharkdp:
> At the beginning this PR was just supposed to bring multiple extensions comparison to fd, but now I end up porting the whole code to support UTF-8, even though I don't think it even works as intended for the program on MacOS in its current state.

I'm sorry - you are right (I have not seen your "Edit" comment above until now). The normalization issue seems to be a general problem on macOS. 

I just wanted to have a Unicode test because I was worried that the specialized iterators would do something wrong with Unicode characters.

I suggest the following for now:
- We replace replace `é` by `é` (the normalized version) in the tests for now. This should work on macOS as well.
- I will create a new ticket to solve the normalization issue on macOS in general.

In general, I'm still not completely happy with the split-up version... couldn't we write something like this for all operating systems:
``` rust
/// Determine if an os string ends with any of the given extensions (case insensitive).
pub fn path_has_any_extension<'a, I>(path: &Path, exts: I) -> bool
where
    I: 'a + IntoIterator<Item = &'a String>,
{
    // TODO: remove these two lines when we drop support for Rust version < 1.23.
    #[allow(unused_imports)]
    use std::ascii::AsciiExt;

    if let Some(path_str) = path.to_str() {
        exts.into_iter().any(|x| {
            let mut it = path_str.chars().rev();

            if x.chars()
                .rev()
                .zip(&mut it)
                .all(|(a, b)| a.eq_ignore_ascii_case(&b))
            {
                match it.next() {
                    Some('/') | None => false,
                    _ => true,
                }
            } else {
                false
            }
        })
    } else {
        false
    }
}
```
(yes, this will not work if a filename has invalid unicode characters)

althonos:
No need to be sorry, I didn't mean to be rude ! I think what you came up with should work as well, so I'll edit the tests to reflect that. 

> In general, I'm still not completely happy with the split-up version...

The advantage of the split-up version is that it's zero-copy on UNIXes, so I think it will end up being quicker. As a side-note, the current implementation could maybe work for both Linux and MacOS if instead of normalising the paths, we normalized the extensions in a way HFS+ likes beforehand. So only Windows would end up using the default slower way.

althonos:
@sharkdp : I think you're going to love that bug (which is why the tests are not working when they should with the normalisation): `String::from_utf8_lossy("e\u{301}xt")` does not put the accute on the `e` but on the `x`!

sharkdp:
> @sharkdp : I think you're going to love that bug (which is why the tests are not working when they should with the normalisation): String::from_utf8_lossy("e\u{301}xt") does not put the accute on the e but on the x!

I can't really reproduce this (`String::from_utf8_lossy("e\u{301}xt")` does not compile).

I think that everything should work if you now remove the normalization stuff in the macOS branch.

sharkdp:
@althonos Thank you very much!

One final question/comment from my side: I still think we could use your fantastic approach (zipping two reversed iterators) but apply it not on a byte level, but on a character level (like in this https://github.com/sharkdp/fd/pull/215#issuecomment-361374634).

Here are some timings:

First, the fallback approach:
```
  Time (mean ± σ):     362.6 ms ±   3.8 ms    [User: 1.765 s, System: 1.034 s]
 
  Range (min … max):   356.2 ms … 375.5 ms
```

Your unix-specialized approach with `.bytes()`:
```
  Time (mean ± σ):     298.5 ms ±   2.1 ms    [User: 1.276 s, System: 1.022 s]
 
  Range (min … max):   294.6 ms … 303.2 ms
```

Your approach, but with `.chars()`:
```
  Time (mean ± σ):     300.9 ms ±   2.8 ms    [User: 1.279 s, System: 1.034 s]
 
  Range (min … max):   296.4 ms … 309.1 ms
```

I would draw the following conclusions from this:
- We should *at least* replace the fallback approach by the `.chars()` version. It is significantly slower.
- We could drop the Unix-specialized version because it's only 0.8% faster (however, this 0.8% speedup is not by chance, it's significant). This would decrease code complexity and simplify maintenance in the future.

What do you think? (sorry this has been dragging on for so long)

althonos:
@sharkdp : You're right, I was against using `to_str` because in my head it needed a copying operation but it does not seem so. It looks equivalent on both mixed-case and lowercase folder (see the [benchmark](https://asciinema.org/a/csWdwlMm9bVjUGbx3TfLvkWuI)).

iology:
Curious why you don't make use of RegexSetBuilder to build an efficient matcher, rather than hand-rolling one.

sharkdp:
Ha! Great idea, @iology. Thank you for the implementation, @althonos. Did you already get a chance to run the benchmarks?

althonos:
It's a little bit slower but still in the same order of speed: [see here](https://asciinema.org/a/1LZOPKmC1GwyYPj7pkwmTb56S). It may be because I didn't use `regex` to its full power, since I'm not really familiar with it (I'm a Pythonista, we don't have fancy `RegexSet` !).

I tried to add a small optimization to only compile the regex in Unicode mode when the extension contains Unicode, but for some reason it became invalid and crashed at runtime.

Main perk is that the code is really reduced now, and there's no need for platform or version-specific tweaks.

sharkdp:
Thank you so much!

PR #218, opened 2018-01-10
Fix typos

vnbhattarai:
This fixes few typos: "distributions", "want to" instead of "want so".


sharkdp:
Oh, nice. Thank you very much!

Let me know if there are other things in the README that you think could be restructured.

vnbhattarai:
You are very welcome 😊

PR #223, opened 2018-01-20
Add svg screencast

marionebl:
* replace `*.gif` screencast with `.svg` version
* add simple script to reproduce [asciinema screencast](https://asciinema.org/a/8ppHiyXyiDgRbKJT1sYH4QsCj) 

```
asciinema rec --command="sh ./screencast.sh"
```

* generated svg version via [svg-term](https://github.com/marionebl/svg-term-cli) 

```
svg-term --cast 8ppHiyXyiDgRbKJT1sYH4QsCj --out screencast.svg --window
```

### Yeah, but why?

* smaller file size `300kb` => `75kb`
* better quality


sharkdp:
This looks great (fantastic project)! Thank you very much.

I love how the screencast is completely scripted.

There are a few differences with respect to the old GIF that I would like to change before merging this:
- I personally love the λ - prompt, but it might confuse some people. I'd prefer something like a simple `> ` or `▶`.
- There is a noticeable delay from the `sleep 0.75` statement when the `fd` command is executed. It looks like `fd` is much slower than it actually is :smile:
- Is it possible to decrease the overall width and height? Most of the terminal is empty on the right hand side.
- I think I'd prefer a gif without the window decoration (although it looks quite cool). 
- As for the executed commands, there are a few things that we could improve
    - It'd be great if we could showcase `-e`. Maybe `fd -e md` to see all Markdown files.
    - Maybe we could also add `fd -e md --exec wc -l` after executing `fd -e md` to show the `--exec` feature.
    - `fd -h` is executed, but the result is not shown?

marionebl:
@sharkdp Updated the screencast as requested, some notes:


> There is a noticeable delay from the sleep 0.75 statement when the fd command is executed. It looks like fd is much slower than it actually is 😄

This was intended to simulate the small delay between tying a command and hitting return. Without it some of the commands where not readable before the output took over the entire screenbuffer. I changed this to sleep **before** creating a newline. Maybe this does not create the impression of `fd` being slower 

> Is it possible to decrease the overall width and height? Most of the terminal is empty on the right hand side.

Decreased the width to 40 columns, this impacts the readability of `-h` output thought. There are some bugs in `svg-term-cli` concerning forced `--height`, so this can't be changed ftb.

sharkdp:
It looks great, thank you very much for the updates!

> I changed this to sleep before creating a newline. Maybe this does not create the impression of fd being slower

Yes, I like it much better.

Sorry for being picky, but I have a few more comments:
- The overall playback speed is quite fast. I think it might be hard to follow. Could we add another delay before typing the first character for each command (to give some time to understand the output from the last command)?
- if we can't decrease the height, I think I'd actually prefer the wider version (you are right about the `-h` text).
- And super-nitpicky: there seems to be a small issue with the Unicode prompt character. It's only visible for a split-second. Maybe we could simply try `>`.

I definitely appreciate your help but if you would like me to merge this as-is, I could also work on these updates myself. In any case, thank you!

marionebl:
Updated again.


> *    The overall playback speed is quite fast. I think it might be hard to follow. Could we add another delay before typing the first character for each command (to give some time to understand the output from the last command)?

I added a delay of 1 second for each new command

> *    if we can't decrease the height, I think I'd actually prefer the wider version (you are right about the -h text).

Removed the dimension bounds


> *    And super-nitpicky: there seems to be a small issue with the Unicode prompt character. It's only visible for a split-second. If there is no fix for that, maybe we could simply try >.

Replaced it for the time being with `>`

> I definitely appreciate your help but if you would like me to merge this as-is, I could also work on these updates myself. In any case, thank you!

No worries, I like to lend a hand :)

sharkdp:
Thanks!

It looks like you accidentally replaced the wrong character with `>`. Now every line ends with `>` and the prompt is still the old one :-)

marionebl:
Got you wrong there I guess - I can‘t see the issue with the arrow disappearing on my machine, could you provide your OS/Browser and a recording of the bug?

sharkdp:
Google Chrome 64 on Linux:

image

- The `>` prompt appears at the end of each line (I would have expected `▶` to be replaced by `>`)
- For a very short time, I see the cursor being right next to the `▶` prompt. My assumption was that this was caused by some Unicode hickup.

sharkdp:
Thank you very much!

Issue #221, opened 2018-01-15
Misleading error when typing -exec


bagage:
I'm so much used to `find` syntax, and so when replacing it with `fd` (which is awesome), I continued to type `-exec` instead of the expected `--exec` (or `-x`). 
But it's only now after few minutes of investigation that I found the reason, the error message being, in my opinion, misleading/unhelpful:

```
$ fd foo -exec rm
Error: 'rm' is not a directory.
$ fd foo -exec rm {} \;
error: Found argument '{}' which wasn't expected, or isn't valid in this context

USAGE:
    fd [FLAGS/OPTIONS] [<pattern>] [<path>]

For more information try --help
```

I know that `-exec` is thus interpreted as `-e=xec` or, more clearly, `--extension xec` which explains why `{}` is not expected here. But since `fd` and `find` share similar syntax, might it make sense to add some userhint `warning: -exec found, did you mean to use --exec instead?` for instance? 
I believe I won't be the last one to fall into this trap :).

sharkdp:
Thank you for the feedback!

I guess we could try to detect that, yes -- but what about the poor people that want to search for `.xec` files? :smile:

sharkdp:
I'm not sure we can distinguish `-e xec` from `-exec` (since we use `clap` to parse command line arguments, and it probably doesn't make that transparent). So we would have to check for that before running `clap`...... or we simply ignore the minuscule probability that someone wants to search for `.xec` files :smile:

Let's implement this.

Doxterpepper:
The problem seems to be that a space is not required between a short arg and the short args value. Clap then doesn't know the difference between -exec as a argument and -e[Value]. I think we could add a validator method to the extension argument to check if there is a space between -e and the value and throw an error appropriately if not.

bagage:
@Doxterpepper I don't think that's a solution -- this is the standard way of handling arguments and `-e value` should behave exactly as `-evalue`. 
At least I suppose that's what people expect to have so I would not break that rule; particularly since this is required only for a very specific use case of `-exec` value. 
If it's impossible to warn user in this single case of `-exec (did you mean --exec?)`, I think it's better to not do anything instead.

Doxterpepper:
[...]

PR #226, opened 2018-01-27
Allow -exec to work as --exec 

stevepentland:
This will check for the presence of -exec and change it to --exec prior to passing the args to clap to support the request of #221 

There may be some merit to updating the help docs for -e option to ensure it is known that an .xec file extension will require the --extension option instead. But that may not really be needed as .xec files don't really seem to be a common type.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -195,19 +195,19 @@ where
             }
             args.push(curr);
             return args;
-        } else {
-            if curr == target || curr == long_start || curr == short_start {
-                args.push(if curr == target {
-                    OsString::from("--exec")
-                } else {
-                    curr
-                });
-                in_exec_opt = true;
-            } else {
-                args.push(curr);
-            }
-            args
         }
+
+        if curr == target || curr == long_start || curr == short_start {
+            args.push(if curr == target {
+                OsString::from("--exec")
+            } else {
+                curr
+            });
+            in_exec_opt = true;
+        } else {
+            args.push(curr);
+        }
+        args
     })
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,6 +32,7 @@ mod windows;
 
 use std::env;
 use std::error::Error;
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time;
@@ -45,7 +46,8 @@ use lscolors::LsColors;
 use walk::FileType;
 
 fn main() {
-    let matches = app::build_app().get_matches();
+    let checked_args = transform_args_with_exec();
+    let matches = app::build_app().get_matches_from(checked_args);
 
     // Get the search pattern
     let pattern = matches.value_of("pattern").unwrap_or("");
@@ -169,4 +171,26 @@ fn main() {
         Ok(re) => walk::scan(&dir_vec, Arc::new(re), Arc::new(config)),
         Err(err) => error(err.description()),
     }
+}
+
+fn transform_args_with_exec() -> Vec<OsString> {
+    let target = {
+        let mut ex = OsString::new();
+        ex.push("-exec");
+        ex
+    };
+
+    if !std::env::args_os().any(|v| v == target) {
+        return env::args_os().collect();
+    }
+
+    let replacement = {
+        let mut exr = OsString::new();
+        exr.push("--exec");
+        exr
+    };
+
+    env::args_os()
+        .map(|v| if v == target { replacement.clone() } else { v })
+        .collect()
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -46,7 +46,7 @@ use lscolors::LsColors;
 use walk::FileType;
 
 fn main() {
-    let checked_args = transform_args_with_exec(env::args_os().collect());
+    let checked_args = transform_args_with_exec(env::args_os());
     let matches = app::build_app().get_matches_from(checked_args);
 
     // Get the search pattern
@@ -178,7 +178,10 @@ fn main() {
 /// # Returns
 ///
 /// * The args, with substitution if required
-fn transform_args_with_exec(original: Vec<OsString>) -> Vec<OsString> {
+fn transform_args_with_exec<I>(original: I) -> Vec<OsString>
+where
+    I: Iterator<Item = OsString>,
+{
     let target = OsString::from("-exec");
     original
         .into_iter()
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -170,3 +172,25 @@ fn main() {
         Err(err) => error(err.description()),
     }
 }
+
+fn transform_args_with_exec() -> Vec<OsString> {
+    let target = {
```
sharkdp: You can use `OsString::from("-exec")`.

src/main.rs
```diff
@@ -170,3 +172,25 @@ fn main() {
...
+        let mut ex = OsString::new();
+        ex.push("-exec");
+        ex
+    };
+
+    if !std::env::args_os().any(|v| v == target) {
```
sharkdp: If we are going to traverse the arguments anyway, I think we can leave out this check. Just try to replace `-exec` with `--exec` below.

src/main.rs
```diff
@@ -170,3 +172,25 @@ fn main() {
...
+    }
+
+    let replacement = {
+        let mut exr = OsString::new();
+        exr.push("--exec");
+        exr
+    };
+
+    env::args_os()
+        .map(|v| if v == target { replacement.clone() } else { v })
```
sharkdp: Instead of `replacement.clone()`, you can use `OsString::from("--exec")` here.

src/main.rs
```diff
@@ -170,3 +172,25 @@ fn main() {
         Err(err) => error(err.description()),
     }
 }
+
+fn transform_args_with_exec() -> Vec<OsString> {
```
sharkdp: Could we add a doc-string here to describe what the function does?

stevepentland:
It is true that -exec passed as an argument to a previous --exec option of fd would result in an unexpected change of the given parameter. My initial thought was to only change the first instance of -exec, but as exec is a parameter that can be repeated this would not work. 

I suppose it comes down to how much weight we want to give to -exec appearing as an arg to another command? Should it be similar to .xec where we simply don't worry about the possibility? Or  would limiting the number of --exec arguments to one be a potential solution for better consistency? If multiple is desired, and we are worried about changing -exec given to the --exec target, then a more complex solution such as a full-on pre-parser seem to be the only option left.

src/main.rs
stevepentland: For sure, no problem

src/main.rs
stevepentland: Thanks! I had missed that

src/main.rs
stevepentland: I've removed this, but please see my comment below

sharkdp:
> My initial thought was to only change the first instance of -exec, but as exec is a parameter that can be repeated this would not work.

Oh, actually `--exec` should only be allowed once. I think you discovered a bug here. It seems like currently `fd -x cmd1 -x cmd2` is actually equal to `fd -x cmd1 cmd2`. We might restrict it to just one usage of `--exec`/`-x`.

> I suppose it comes down to how much weight we want to give to -exec appearing as an arg to another command? Should it be similar to .xec where we simply don't worry about the possibility?

I'd be okay with that.

> Or would limiting the number of --exec arguments to one be a potential solution for better consistency?

I think that would be even better.

stevepentland:
Ok, so I've tried to reach a middle ground with the handling. I have limited --exec to a single occurrence (as opt for fd). This means I am not able to only transform the first instance of -exec as it will no longer cause an error in clap. I've added some tests, both to the handling of the incoming args, and the overall operation. An option of -exec passed to a secondary command will also be transformed, but at least this allows the downstream to provide it's own error explanation.

sharkdp:
> This means I am not able to only transform the first instance of -exec

Why not? Clap will not throw an error because it knows that the other `-exec` or `--exec` is just a value given to the exec-argument.

A possible implementation could look like this:
``` rust
/// Traverse args_os, looking for -exec and replacing it with --exec.
fn transform_args_with_exec<I>(original: I) -> Vec<OsString>
where
    I: Iterator<Item = OsString>,
{
    let mut new_args: Vec<OsString> = original.collect();

    let target = OsString::from("-exec");
    let result = new_args.iter().position(|arg| *arg == target);

    if let Some(pos) = result {
        new_args[pos] = OsString::from("--exec");
    }

    return new_args;
}
```

It works as expected:
```
▶ fd -exec echo -exec --exec    
-exec --exec README.md
-exec --exec src
-exec --exec LICENSE-MIT
...
```


This will actually still leave us with a problem if someone passes the actual `--exec`/`-x` to `fd` but `-exec` to the command (as it will now replace the wrong `-exec`):
```
▶ fd --exec echo -exec --exec    
--exec --exec README.md
--exec --exec src
--exec --exec LICENSE-MIT
...
```

sharkdp:
To be 99.9% foolproof, we could actually only replace the first `-exec` by `--exec` if neither `-x` nor `--exec` is in the arguments vector. This would only fail if someone (wrongly) uses `-exec` for *fd* but `-x` or `--exec` for another program.

stevepentland:
When only the first is changed, clap will later on see the second -exec passed to fd as --extension instead. As I removed the .multiple from the exec arg, it is now relying on the validation from clap. Currently this means the only error that will occur is when someone tries to pass -exec to the given executable and not fd

sharkdp:
Oh, you mean if someone passes `-exec` to fd twice? Like to execute two commands for each search result?

stevepentland:
> Oh, you mean if someone passes -exec to fd twice? Like to execute two commands for each search result?

Yes, this causes a strange hybrid behavior in which we have new action for the first -exec but then the same error from the original bug for any further -exec

stevepentland:
I figured that leaving any errors to the downstream makes at least our behavior consistent, and since -exec isn't the most common option (I only know of find) then this is probably the least impactful

stevepentland:
@sharkdp  I've created a small state machine that will keep track of it's current position (in exec option or out) and properly convert -exec as expected. It will need some cleanup and some code should be moved around, but it could do the trick.

sharkdp:
Oh wow, nice! :+1:

stevepentland:
> Oh wow, nice! 👍
Thanks! I've actually gone over it again and reduced it to a single fold that does the same thing without all the extra junk. It's pretty concise and works as expected, we change any -exec to --exec, but not when inside of a previous --exec, -exec, or -x block, delimited with the terminator ;

Doxterpepper:
Have you considered using the validator method in clap to solve this problem? https://docs.rs/clap/2.29.2/clap/struct.Arg.html#method.validator

stevepentland:
> Have you considered using the validator method in clap to solve this problem?

I had looked at it, but it does not seem to provide a way to change one argument into another. It's not really validation that is being looked for, rather to keep some level of similarity to the standard find utility.

stevepentland:
@sharkdp This is pretty much done for what we wanted to do. Any more changes you'd like me to make?

sharkdp:
No, it looks great.

Thank you very much!

PR #231, opened 2018-02-06
Install `rustfmt-nightly` in debug mode in Travis

althonos:
This should quicken the build time of the `rustfmt-nightly` binary by a factor of two without impacting the actual runtime, hence reducing the total CI job time by about 3 minutes.


sharkdp:
Nice! Thank you very much.

PR #236, opened 2018-02-14
Update dependencies

srijs:
This patch bumps the `ansi_term` and `ignore` dependencies to the next minor version.

As a suggestion, it also includes a dependency status badge in the README.md. Let me know if you want to keep it!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "crossbeam"
-version = "0.2.10"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


srijs:
Looks like we're running into this issue (again?): https://github.com/ogham/rust-ansi-term/issues/31

sharkdp:
Welcome, Thank you very much for your contribution!

https://deps.rs looks like a really cool and helpful project! I have added it to my release to-do list.

I'm hesitant to add a status badge for the dependency status, though. Mainly because I don't think that an "outdated" dependency is necessarily a bad thing. The `Cargo.toml` manifest already allows me [to specify](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) dependencies in such a way that (compatible) updates are automatically pulled in on `cargo update`. I might choose to use the `0.6.x` versions of a specific library (including the latest patches `x`) because `0.7` breaks some things for me (see `ansi_term`). It might also be because the `0.6` version of a dependency works just fine and I don't want to risk updating to `0.7` just yet.

srijs:
That's fair enough, will remove the badge for now. It was just a suggestion after all :)

What do you think about the `ansi_term` issue? Would you rather stick with it for now, or should I try to `termcolor` instead?

sharkdp:
For now, I don't see any reason to switch to another crate. If, however, `ansi_term` will continue to be unmaintained, we can explore other options (`termcolor`, `colored`, ..).

srijs:
Fair enough, closing this!

PR #238, opened 2018-02-18
Fixed #234: Suggest --fixed-strings on invalid regexps

### LINKED ISSUE #234: Suggest --fixed-strings on invalid regexps
Inspired by https://github.com/BurntSushi/ripgrep/issues/709

PramodBisht:
Fixed #234 
Giving suggestion to use `--fixed-strings` in case user passes invalid metacharacters as a regex.

@sharkdp it may be a very dumb solution, I have started giving suggestion to use `--fixed-string` for any value of `ErrorKind` of `regex` library. Please let me know if I need to exclude any specific set of `ErrorKind`. All test cases are passing over this changeset.

Let me know if any test case needs to be added for this.

Plus I have added intellij-idea based editor config directory in .gitignore, so it doesn't annoy other devs in git's unstaging area.

Thanks

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -182,6 +182,12 @@ fn main() {
         .build()
     {
         Ok(re) => walk::scan(&dir_vec, Arc::new(re), Arc::new(config)),
-        Err(err) => error(err.description()),
+        Err(err) => error(
+            format!(
+                "{}\nHint: You can use the '--fixed-strings' option to search for a \
+                            literal string instead of a regular expression",
+                err.description()
+            ).as_str(),
+        ),
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -185,7 +185,7 @@ fn main() {
         Err(err) => error(
             format!(
                 "{}\nHint: You can use the '--fixed-strings' option to search for a \
-                            literal string instead of a regular expression",
+                 literal string instead of a regular expression",
                 err.description()
             ).as_str(),
         ),


.gitignore
```diff
@@ -1,2 +1,5 @@
 target/
 **/*.rs.bk
+
+# IDEs
+.idea/*
```
sharkdp: I think this something that rather belongs in a personal, global `.gitignore` file.

src/main.rs
```diff
@@ -182,6 +182,12 @@ fn main() {
         .build()
     {
         Ok(re) => walk::scan(&dir_vec, Arc::new(re), Arc::new(config)),
-        Err(err) => error(err.description()),
+        Err(err) => error(
+            format!(
+                "{} \n(Hint: Try the --fixed-strings \
+                 flag to search for a literal string.)",
```
sharkdp: Maybe: "Hint: You can use the '--fixed-strings' option to search for a literal string instead of a regular expression"?

src/main.rs
```diff
@@ -182,6 +182,12 @@ fn main() {
         .build()
     {
         Ok(re) => walk::scan(&dir_vec, Arc::new(re), Arc::new(config)),
-        Err(err) => error(err.description()),
+        Err(err) => error(
+            format!(
+                "{} \n(Hint: Try the --fixed-strings \
```
sharkdp: I guess the space after `{}` can be removed.

sharkdp:
Thank you very much!

> I have started giving suggestion to use --fixed-string for any value of ErrorKind of regex library. Please let me know if I need to exclude any specific set of ErrorKind. All test cases are passing over this changeset.

I think your solution is perfectly fine.

> Let me know if any test case needs to be added for this.

That would be great, but I'm not sure we currently have a good way to test for this. I'd be okay with merging this without a test case.

> Plus I have added intellij-idea based editor config directory in .gitignore, so it doesn't annoy other devs in git's unstaging area.

See my inline comment.

src/main.rs
PramodBisht: @sharkdp right, it's more meaningful, will push this code change.

src/main.rs
PramodBisht: @sharkdp I will remove this useless space :)

PramodBisht:
@sharkdp done all code changes and rebased multiple commits to one.

PR #239, opened 2018-02-19
README.md: gentoo now has fd ebuild, document it

gyakovlev:


sharkdp:
Very cool! Thank you very much.

PR #254, opened 2018-03-08
Update to regex-syntax 0.5

cuviper:
The new regex 0.2.7 has updated to regex-syntax 0.5, so it seems prudent
to update fd's dependency too, if only to avoid duplication.


cuviper:
The CI failure is a format check, on code I didn't touch, but I can update that if you want.
https://travis-ci.org/sharkdp/fd/jobs/350683392#L787

sharkdp:
Welcome, Thank you very much!

Could you please rebase this on master? This should fix the CI failure.

cuviper:
Sure, here's the rebase.

sharkdp:
Thank you very much!

PR #259, opened 2018-03-14
Print error if user supplied path instead of search pattern

### LINKED ISSUE #235: Detect when user supplied path instead of pattern
From time to time it happens to me that I forget to supply a regex-pattern when using an explicit search path. In particular, this happens when I use `--extension EXT`/`-e EXT` because it feels like `EXT` is a pattern by itself:
``` bash
> fd -e jpg /share/pictures    # this will not do what you expect
```

The right way would be to use `fd -e jpg . /share/pictures` or `fd -e jpg "" /share/pictures`.

Instead, what the command above will try to do is to search for files in the current directory that match the regular expression `/share/pictures` (there will be none). I was wondering whether we could detect such cases and warn the user.

A possible implementation could look like this:

If all of the following conditions are fulfilled:
- only a single argument is given (which fd normally interprets as a search pattern)
- the pattern contains a path-separation character (`/` or `\`)
- (the *pattern* is an existing path in the filesystem)
- `--full-path`/`-p` has *not* been enabled (otherwise, we could potentially find entries that match the pattern)

... print a warning:
```
The search pattern "/share/pictures" looks like a path. If you want to specify a
root directory, use an empty search pattern: `fd [options] "" /share/pictures`.
```

sharkdp:
Fixes #235

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,6 +77,22 @@ fn main() {
             .collect();
     }
 
+    // Detect if the user accidentally supplied a path instead of a search pattern
+    if !matches.is_present("full-path") && pattern.contains(std::path::MAIN_SEPARATOR)
+        && fshelper::is_dir(Path::new(pattern))
+    {
+        error(&format!(
+            "Error: The search pattern '{pattern}' contains a path-separation character ('{sep}') \
+             and will not lead to any search results.\n\n\
+             If you want to search for all files inside the '{pattern}' directory, use a match-all pattern:\n\n  \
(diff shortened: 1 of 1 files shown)


cup:
https://github.com/sharkdp/fd/issues/235
https://github.com/sharkdp/fd/issues/277
https://github.com/sharkdp/fd/issues/312
https://github.com/sharkdp/fd/issues/385
https://github.com/sharkdp/fd/pull/259
https://github.com/sharkdp/fd/pull/349

PR #275, opened 2018-03-31
README: openSUSE package available

avindra:
add software.opensuse.org link and `zypper in fd`


sharkdp:
Thank you very much!!

Issue #281, opened 2018-04-12
Could you add `using fd with Emacs` section into README


redguardtoo:
See
https://github.com/technomancy/find-file-in-project#use-fd-a-simple-fast-and-user-friendly-alternative-to-find

Usage is simple,
Install `find-file-in-project`, add one line `(setq ffip-use-rust-fd t)` into `~/.emacs`.

Tested on Windows/Linux/Mac.

sharkdp:
Thank you for your feedback.

Sounds good! Maybe we should add this (and the part about `fzf`) into a "Integration with other programs" section?

Would you like to open a PR for this?

PR #282, opened 2018-04-12
Re-org fzf part. Doc on using fd with Emacs

redguardtoo:
- re-organize `fzf` part into new section `Intergration with other programs`
- add `using fd with Emacs` part

See https://github.com/sharkdp/fd/issues/281 for details


redguardtoo:
Can't figure out why one build failed.

sharkdp:
Thank you very much!

> Can't figure out why one build failed.

Don't worry. This is not related to your PR.

Issue #276, opened 2018-04-01
support size filter option


yifeikong:
`find` supports the `-size` filter, could `fd` do that too?

sharkdp:
Thank you for your feedback!

Filtering by file size is something that I would consider not to be part of the "80% of the use cases" (see README). If we decide to include this (and related tickets #165 and #273), I guess we have to re-consider that statement at some point :smile:.

Before we move forward with this, I would like to discuss a clear strategy on how to implement this. In particular, I would like to know which kind of syntax we would use for file-size filtering.

LNJ2:
For the syntax I'd suggest something as:
`-S >9M` (larger than 9 GiB) or `-S <5G` (smaller than 5 MiB)

And for ranges:
`-S 100k-2M` (larger than 100 KiB, but smaller than 2 MiB)

It should accept both capital and non-capital k for kilobytes (`1k` / `1K`), even if the capital variant is actually not accurate (captial letters only start from Mega). The only problem this might cause is that the `>` character might interfere with the bash syntax.

sharkdp:
Thank you for the feedback!

> The only problem this might cause is that the > character might interfere with the bash syntax.

Exactly. `fd -S >9M` would be parsed as "pipe the output of `fd -S` into the file named `9M` by the shell. This would mean that users would have to remember to quote the arguments (`fd -S '>9M'`), which makes this less appealing.

Maybe we could introduce two arguments (`--larger-than`, `--smaller-than` or `--size-larger`, `--size-smaller`)? They could act in a *AND*-fashion, such that ranges could be specified via `--larger-than 100k --smaller-than 2M`.

Susurrus:
> Maybe we could introduce two arguments (--larger-than, --smaller-than or --size-larger, --size-smaller)? They could act in a AND-fashion, such that ranges could be specified via --larger-than 100k --smaller-than 2M.

That seems pretty reasonable. I was going to suggest following the same syntax as `find` has, where you use a `+` or `-` prefix to indicate larger or smaller, but I like that this reads better and is composable. My question here is that does this include the equal-to case? I'd expect that there is little need for both a `--larger-than` and a `--larger-or-equal-to` argument,

sharkdp:
Actually, I'd be interested in some *real world* use cases for the usage of the size option. I don't think I have ever used `find -size` for anything other than searching for large files.

Susurrus:
[...]

PR #286, opened 2018-04-23
Initial impl of size constraints.

stevepentland:
This adds find-style size constraints in relation to #276.

Sizes are indicated using the same notation for find:
`+`: Greater than or equal to this value
`-`: Less than or equal to this value

Thus, greater than 10KB is `+10K` and less than 1MB is `-1M`.  Values can have trailing B if desired and are not case-sensitive.

As a side note, on current master `test_fixed_strings` fails for some reason on Linux (Fedora 27, up-to-date),

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -264,16 +264,20 @@ fn usage() -> HashMap<&'static str, Help> {
         , "Alias for no-ignore ('u') and no-ignore and hidden ('uu')");
     doc!(h, "size"
         , "Limit results based on the size of files."
-        , "Limit results based on the size of files using the format <+-><NUM><UN>.\n   \
-            '+': file size must be greater than this\n   \
-            '-': file size must be less than this\n   \
-            'NUM':      The numeric size (e.g. 500)\n   \
-            'UN':       The units for NUM.\n\
+        , "Limit results based on the size of files using the format <+-><NUM><UNIT>.\n   \
+            '+': file size must be greater than or equal to this\n   \
+            '-': file size must be less than or equal to this\n   \
+            'NUM':  The numeric size (e.g. 500)\n   \
+            'UNIT': The units for NUM. They are not case-sensitive.\n\
             Allowed unit values:\n   \
-                'b' or 'B': bytes\n   \
-                'k' or 'K': kilobytes\n   \
-                'm' or 'M': megabytes\n   \
-                'g' or 'G': gigabytes\n   \
-                't' or 'T': terabytes");
+                'b':  bytes\n   \
+                'k':  kilobytes\n   \
+                'm':  megabytes\n   \
+                'g':  gigabytes\n   \
+                't':  terabytes\n   \
+                'ki': kibibytes\n   \
+                'mi': mebibytes\n   \
+                'gi': gibibytes\n   \
+                'ti': tebibytes");
     h
 }
--- a/src/app.rs
+++ b/src/app.rs
@@ -5,7 +5,6 @@
 // at your option. All files in the project carrying such
 // notice may not be copied, modified, or distributed except
 // according to those terms.
-
 use std::collections::HashMap;
 
 use clap::{App, AppSettings, Arg};
@@ -16,10 +15,10 @@ struct Help {
 }
 
 macro_rules! doc {
-    ($map:expr, $name:expr, $short:expr) => {
+    ($map: expr, $name: expr, $short: expr) => {
         doc!($map, $name, $short, $short)
     };
-    ($map:expr, $name:expr, $short:expr, $long:expr) => {
+    ($map: expr, $name: expr, $short: expr, $long: expr) => {
         $map.insert(
             $name,
             Help {
@@ -147,6 +146,15 @@ pub fn build_app() -> App<'static, 'static> {
                 .value_name("num"),
         )
         .arg(
+            arg("size")
+                .long("size")
+                .short("S")
+                .takes_value(true)
+                .number_of_values(1)
+                .allow_hyphen_values(true)
+                .multiple(true),
+        )
+        .arg(
             arg("max-buffer-time")
                 .long("max-buffer-time")
                 .takes_value(true)
@@ -254,6 +262,18 @@ fn usage() -> HashMap<&'static str, Help> {
     doc!(h, "rg-alias-hidden-ignore"
         , "Alias for no-ignore and/or hidden"
         , "Alias for no-ignore ('u') and no-ignore and hidden ('uu')");
-
+    doc!(h, "size"
+        , "Limit results based on the size of files."
+        , "Limit results based on the size of files using the format <+-><NUM><UN>.\n   \
+            '+': file size must be greater than this\n   \
+            '-': file size must be less than this\n   \
+            'NUM':      The numeric size (e.g. 500)\n   \
+            'UN':       The units for NUM.\n\
+            Allowed unit values:\n   \
+                'b' or 'B': bytes\n   \
+                'k' or 'K': kilobytes\n   \
+                'm' or 'M': megabytes\n   \
+                'g' or 'G': gigabytes\n   \
+                't' or 'T': terabytes");
     h
 }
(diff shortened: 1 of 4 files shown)


src/walk.rs
```diff
@@ -248,6 +248,16 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) {
                 }
             }
 
+            // Filter out unwanted sizes if it is a file and we have been given size constraints.
+            if entry_path.is_file() && config.size_constraints.len() > 0 {
```
sharkdp: It might be premature optimization, but I think we should probably swap these two conditions (and verify by a benchmark). The way it is written right now would lead to a syscall even when there are no size constraints.

src/internal.rs
```diff
@@ -14,10 +14,14 @@ use std::time;
 
 use exec::CommandTemplate;
 use lscolors::LsColors;
-use regex::RegexSet;
+use regex::{Regex, RegexSet};
 use regex_syntax::hir::Hir;
 use regex_syntax::Parser;
 
+lazy_static! {
+    static ref SIZE_CAPTURES: Regex = { Regex::new(r"^(\+|-)(\d+)([a-zA-Z]{1,2})$").unwrap() };
```
sharkdp: Why do we need two regex patterns for capturing and validation? Wouldn't one be enough?

src/main.rs
```diff
@@ -34,12 +34,17 @@ use std::sync::Arc;
...
+use regex::{RegexBuilder, RegexSetBuilder, Regex};
 
 use exec::CommandTemplate;
-use internal::{error, pattern_has_uppercase_char, transform_args_with_exec, FdOptions, FileTypes};
+use internal::{error, pattern_has_uppercase_char, transform_args_with_exec, FdOptions, FileTypes,
+               SizeFilter};
 use lscolors::LsColors;
 
+lazy_static! {
+    static ref VALIDATE_SIZE: Regex = { Regex::new(r"^[\+-]{1}\d+[bBkKmMgGTt]{1,2}$").unwrap() };
```
sharkdp: I'm going to review this regex piece by piece :smile: 

- `[\+-]{1}` - unneccesary `{1}`?

- `\d+` - Just an idea... do we want to allow for floating point numbers as well?

- `[bBkKmMgGTt]{1,2}` - This allows for a lot of wrong patterns like `bb`, `mm` or `bk`. I think we should separate the unit-part from the optional `byte` annotation and write something like `[kKmMgGtT]?[bB]?`.

Also, see my other comment regarding the units.

src/internal.rs
```diff
@@ -37,6 +41,58 @@ impl Default for FileTypes {
...
+impl SizeFilter {
+    pub fn is_within(&self, size: u64) -> bool {
+        match self.limit_type {
+            SizeLimitType::Max => size <= self.size,
+            SizeLimitType::Min => size >= self.size,
+        }
+    }
+}
+
+const KILO: u64 = 1024;
```
sharkdp: Concerning units, I would like to follow the [official SI and EIC standards](https://en.wikipedia.org/wiki/Kibibyte). The "k", "M", "G", "T"... prefixes are for powers of 10 while the "Ki", "Mi", "Gi", "Ti"... prefixes are for powers of 2. For example:

- 1 MB = 10^6 byte = 1000² byte
- 1 MiB = 2^20 byte = 1024² byte

We can add the "Ki", "Mi".. prefixes later, but these numbers here should be `1000`, not `1024`.

src/app.rs
```diff
@@ -254,6 +262,18 @@ fn usage() -> HashMap<&'static str, Help> {
...
         , "Alias for no-ignore and/or hidden"
         , "Alias for no-ignore ('u') and no-ignore and hidden ('uu')");
-
+    doc!(h, "size"
+        , "Limit results based on the size of files."
+        , "Limit results based on the size of files using the format <+-><NUM><UN>.\n   \
+            '+': file size must be greater than this\n   \
+            '-': file size must be less than this\n   \
+            'NUM':      The numeric size (e.g. 500)\n   \
+            'UN':       The units for NUM.\n\
```
sharkdp: Great help text!

Can we replace `<UN>` by `<UNIT>`? I think we can be verbose in the "long" help text.

src/internal.rs
```diff
@@ -37,6 +41,58 @@ impl Default for FileTypes {
...
+const MEGA: u64 = KILO * 1024;
+const GIGA: u64 = MEGA * 1024;
+const TERA: u64 = GIGA * 1024;
+
+impl<'a> From<&'a str> for SizeFilter {
+    /// Create the `SizeFilter` from the given `&str`.
+    /// It is imperative that the incoming value has been validated for
```
sharkdp: Could we move the verification of the format here in this function (`from`) and return `Option<Self>`?

src/internal.rs
```diff
@@ -37,6 +41,58 @@ impl Default for FileTypes {
...
+    /// proper format.
+    fn from(s: &str) -> Self {
+        let captures = SIZE_CAPTURES.captures(s).unwrap();
+        let limit = match captures.get(1).map_or("+", |m| m.as_str()) {
+            "+" => SizeLimitType::Min,
+            _ => SizeLimitType::Max,
+        };
+
+        let quantity = captures.get(2).unwrap().as_str().parse::<u64>().unwrap();
```
sharkdp: If we move the verification to this function, we could replace all the `.unwrap()` calls by `?`.

src/walk.rs
stevepentland: Yeah, it would be better that way

src/main.rs
stevepentland: - Yes, the `{1}` is not required to qualify this match. 
- I do not believe that adding floating point support would increase functionality at all
- This is true, I will switch it to your suggested pattern.

src/internal.rs
stevepentland: Well, we could verify the pattern and call `error` if it was not correct, since it is something we fail for anyway. According to the docs `From<T>` must not fail. I could switch out the implementation for something returning `Option<Self>` also, but it comes down to whether we want to fail & report for incorrect parameters or continue silently

sharkdp: Maybe we just provide a `SizeFilter::from_string` method instead of implementing `From`? This method could then return `Option<Self>` and we could move the error handling / error output to the call site (the main function).

stevepentland: Sure, sounds good

stevepentland:
Still could use some integration tests, just need to see how to use `TestEnv` and create sized files.

stevepentland:
@sharkdp It seems builds are failing due to rust version. It seems the `?` operator was not available for `Option<T>` in rust 1.20. I can change the code to get around this if you really want to keep the CI pinned at that version, or if you want to update to a newer stable then it should start passing.

src/internal.rs
```diff
@@ -14,10 +14,14 @@ use std::time;
 
 use exec::CommandTemplate;
 use lscolors::LsColors;
-use regex::RegexSet;
+use regex::{Regex, RegexSet};
 use regex_syntax::hir::Hir;
 use regex_syntax::Parser;
 
+lazy_static! {
+    static ref SIZE_CAPTURES: Regex = { Regex::new(r"(?i)^([+-])(\d+)([bkmgt]i?)b?$").unwrap() };
```
sharkdp: This still allows for `bb` and `bib`, right? The `b` in the first character clause should be removed, I think. In order to make the whole unit optional, just add `?` after the parens:

`([kmgt]i?)?b?`

src/internal.rs
```diff
@@ -37,6 +41,79 @@ impl Default for FileTypes {
...
+    Min,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct SizeFilter {
+    size: u64,
+    limit_type: SizeLimitType,
+}
+
+// SI units for now
```
sharkdp: Can this be removed? or maybe `// SI prefixes (powers of ten)` and `// Binary prefixes (powers of two)`?

stevepentland: Yeah, forgot to change the comments when I updated

PR #290, opened 2018-04-25
Do not include non-files when using --size

sharkdp:
Right now, directories (and other non-files) are included, when using the `--size` option. With this change, we exclude everything except files.

/cc @stevepentland


sharkdp:
> Isn't there already a way to include only files in the search?

Yes, there is `--type file`. However, the current `--size` implementation includes *all* directories (in fact, everything that is not a file) without checking the size.

> It may be that a user would want to see these directories and only limit the returned files by size no?

I think directories will always be reported with 4096 bytes (we do not want to calculate the size recursively), so `--size` will not be very helpful in combination with directories.

Issue #289, opened 2018-04-25
Write integration tests for new '--size' option


sharkdp:
Test various combinations of `--size +X`, `--size -Y`.

PR #292, opened 2018-04-29
Add tests for SizeFilter (--size) fixes sharkdp/fd#289

JonathanxD:
This adds the integration test for `--size` option.

This is my first pull request, I'd love to hear your opinion. :D

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1005,9 +1005,6 @@ fn test_invalid_utf8() {
 /// (--size)
 #[test]
 fn test_size() {
-    use std::ffi::OsStr;
-    use std::os::unix::ffi::OsStrExt;
-
     let dirs = &["test1", "test2"];
     let files = &["test1/a.foo", "test1/b.foo", "test1/c.foo"];
     let te = TestEnv::new(dirs, files);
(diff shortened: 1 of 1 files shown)


tests/tests.rs
```diff
@@ -1001,3 +1001,77 @@ fn test_invalid_utf8() {
...
+/// (--size)
+#[test]
+fn test_size() {
+    let dirs = &["test1", "test2"];
+    let files = &["test1/a.foo", "test1/b.foo", "test1/c.foo"];
+    let te = TestEnv::new(dirs, files);
+
+    let mut f = fs::File::create(te.test_root().join("test1/tiny.foo")).unwrap();
+    f.write("Hello world".as_bytes()).unwrap(); // 11 bytes
+    f.flush().unwrap();
```
sharkdp: Maybe it'd be worth to put these lines into a little helper function:

``` rust
fn create_file_with_size(filename: &str, size_in_bytes: usize) {
    let content = "#".repeat(size_in_bytes);
    let mut f = File::create(filename).unwrap();
    f.write(content.as_bytes()).unwrap();
}
```

The `flush()` call would not be necessary because the files would be closed properly at the end of that helper function when `f` goes out of scope.

Also, I think the tests would be a little bit easier to read if the files would be called `11_byte.foo` and `30_byte.foo`.

sharkdp:
Thank you very much!

PR #308, opened 2018-07-01
simplify SizeFilter as enum with value instead of struct

alexmaco:
Also move the touched cfg(test) code into cfg(test) module and trim test names


sharkdp:
Looks great! Thank you very much!

PR #309, opened 2018-07-01
Preserve non utf8 names

alexmaco:
Hopefully this fixes #295 .
Right now it's linux-only, since i don't have access to other unices.
The test harness changes are also a little fragile, and it will work reliably only for single-line output.
Let me know if anything needs adjusting, or if there's anything better to be done about `TestEnv`.

### DIFF
--- a/src/exec/input.rs
+++ b/src/exec/input.rs
@@ -71,90 +71,46 @@ pub fn dirname(path: &str) -> &str {
 }
 
 #[cfg(test)]
-mod tests {
+mod path_tests {
     use super::{basename, dirname, remove_extension, MAIN_SEPARATOR};
 
     fn correct(input: &str) -> String {
         input.replace('/', &MAIN_SEPARATOR.to_string())
     }
 
-    #[test]
-    fn path_remove_ext_simple() {
-        assert_eq!(remove_extension("foo.txt"), "foo");
+    macro_rules! func_tests {
+        ($($name:ident: $func:ident for $input:expr => $output:expr)+) => {
+            $(
+                #[test]
+                fn $name() {
+                    assert_eq!($func(&correct($input)), correct($output));
+                }
+            )+
+        }
+    }
+
+    func_tests! {
+        remove_ext_simple:  remove_extension  for  "foo.txt"      =>  "foo"
+        remove_ext_dir:     remove_extension  for  "dir/foo.txt"  =>  "dir/foo"
+        hidden:             remove_extension  for  ".foo"         =>  ".foo"
+        remove_ext_utf8:    remove_extension  for  "💖.txt"       =>  "💖"
+        remove_ext_empty:   remove_extension  for  ""             =>  ""
+
+        basename_simple:  basename  for  "foo.txt"      =>  "foo.txt"
+        basename_dir:     basename  for  "dir/foo.txt"  =>  "foo.txt"
+        basename_empty:   basename  for  ""             =>  ""
+        basename_utf8_0:  basename  for  "💖/foo.txt"   =>  "foo.txt"
+        basename_utf8_1:  basename  for  "dir/💖.txt"   =>  "💖.txt"
+
+        dirname_simple:  dirname  for  "foo.txt"      =>  "."
+        dirname_dir:     dirname  for  "dir/foo.txt"  =>  "dir"
+        dirname_utf8_0:  dirname  for  "💖/foo.txt"   =>  "💖"
+        dirname_utf8_1:  dirname  for  "dir/💖.txt"   =>  "dir"
+        dirname_empty:   dirname  for  ""             =>  "."
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


alexmaco:
Actually, I've now realized this is incomplete. The `--exec` issue is still there.
Will update soon.

sharkdp:
Thank you very much for working on this!

Let me know when it's ready to review.

alexmaco:
While i initially thought it should be a quick fix, it turns out to be much more involved. Essentially, it looks like multiple operations from exec/* need to be ported from working over `str` to `OsStr`/`Path`. Some work well, but things like `remove_extension` are really tricky to do in a cross-platform way.

I'm thinking of having a `cfg(unix)` version done first (since we can use OsStr as bytes there), and leaving windows for later (and someone who can actually test on windows). Alternatively maybe there's a lib out there hat can do it, but i haven't found one.

sharkdp:
Your changes look great, thank you!

I'd be okay with a unix-only/linux-only version for now. Do you think the `--exec` issue could be fixed using a similar approach?

alexmaco:
Yes, the exec issues also stem from a `to_string_lossy()`, it's just more lines to change :smile: 

Yes, some benchmarks are definitely in order. It has the potential to be a little faster since it won't construct new strings for non-utf8 cases, but we'd also want to avoid surprise regressions.

I'm almost done with the unix-specific version, i still have to get a reasonably-efficient way to get the dirname.

alexmaco:
Now this is most of it, but there's still a few kinks.

The BIG One: for some reason the test fails on darwin, so it's linux only, but the code is essentially broken on non-linux unices, see [https://travis-ci.org/sharkdp/fd/jobs/400208971](). I have no idea why right now.

The smaller one: the tests in `input` have been shortened, but then copied over to `input_unix`, which used them as reference. The intent was to have a single separate file with the tests, and then import the testing macro either from `input` or `input_unix`. Not sure whether to go thru with that idea or try something else.

Comments/suggestions welcome.

sharkdp:
> The BIG One: for some reason the test fails on darwin, so it's linux only, but the code is essentially broken on non-linux unices, see https://travis-ci.org/sharkdp/fd/jobs/400208971. I have no idea why right now.

Not that I'm an expert, but I think the Mac OS HFS filesystem does some kind of sanitizing before storing a file with a particular name. We recently had an issue where HFS would normalize Unicode strings in a particular way (see https://stackoverflow.com/a/6153713/704831).

Apparently something similar happens when we create a file named `test_\xFEinvalid.txt`: it shows up in the filesystem as `test_%FEinvalid.txt` and the `%` sign actually seems to be there:
```
> ls test_%*
test_%FEinvalid.txt

> find . -name "*%*"
./test_%FEinvalid.txt
```

sharkdp:
Just a quick note to let you know that I am still planning to review this, but simply cannot find the time right now. It will probably take me a few more days.

sharkdp:
Sorry for not responding for such a long time. I still haven't fully reviewed every part of this.

My main issue with this is that this has somehow turned into a rather large amount of changes to the code base. Most of these are platform-dependent, which will make it harder to maintain in the future. And all of this just to fix a small, weird corner-case (I don't know if I have ever experience a non-UTF8 filename).

Don't get me wrong. Your code looks great and I really, really appreciate the effort - but I also have the overall maintainability/quality of the code base in mind (which even presently is not in the state that I would like it to be in).

alexmaco:
@sharkdp Similarly, I have to apologize for leaving this hanging for such a long time :).

I can see the argument however for not including as many changes for a small case. The last time i dealt with non-utf8 file names myself was many years ago, due to a filesystem corruption.

This leaves 2 questions that I can see:
1. What does this mean for the original issue ? I can keep trying to make the changes smaller and less intrusive, especially if you have any other approaches to suggest. Otherwise, should the original issue be closed ?
2. Since you mentioned maintainability, do you have any specifics in mind ? Perhaps some issues can be logged for cleanup. I noticed myself quite a few things to clean up, but it would be nice to know my observations agree with the general project direction before spamming refactoring PRs :).

sharkdp:
Thank you for the feedback.

There are many things I have in mind that would hopefully help to improve the code quality:
- Group related functions into `impl` blocks of appropriate `structs` such that we don't have to pass around so many parameters.
- Think about a unified interface for filters (file type, file size, modification data, file extension, ..). Maybe a `trait`-based interface where the filters would be stored in a `Vec<Box<dyn Filter>>`.
- Extract some of the functionality like the `OsStr`-based path handling into separate libraries / crates.

alexmaco:
Perhaps we should take these design discussions on a different thread. I've created #382 for this purpose, and quoted the bullet list above as a starting point.

sharkdp:
Closing this in favor of #558, which is based on this PR. @alexmaco Thank you so much for your work!

PR #316, opened 2018-08-09
Remove unused windows dependencies

kpcyrd:
It seems those dependencies are not actually in use:
```
cargo check --target=x86_64-pc-windows-gnu
```
finished successfully with this patch applied.


sharkdp:
Thank you very much!

PR #323, opened 2018-09-13
Add `--maxdepth` as hidden alias for `--max-depth`

mqudsi:
For muscle memory compatibility with `rg`. This variant will not show in
the help or in the program options, and is only checked if `--maxdepth`
is not specified.


sharkdp:
Looks great! Thank you very much.

sharkdp:
Released in [fd-7.2.0](https://github.com/sharkdp/fd/releases/tag/v7.2.0).

Issue #307, opened 2018-06-27
Finding by owner username/group/uid/gid


alexmaco:
Hello,

I've been using `fd` for quite a long time, and loving it! Recently I've found myself needing to search for files owned by one user or another. I'm willing to do any implementation work required for it, and i'd like to start off a discussion and see if anyone else here has had a need for this feature.

Interface-wise, my first proposal would be to add a `--user` (as well as a `--group` switch) in the style of find, and maybe they could handle a uid/gid by default when passed a number.

Another idea, which is more flexible and less backwards-looking would be passing an `--owner` flag that can parse a `[(uid|username)][:(gid|groupname)]` input pair. As a precedent, docker's `--user` does this. Maybe it would make sense to call the flag `--user` in fd also, but i'm personally worried that it would make finding the "search by gid" option a little more obscure or unintuitive.

Lastly, I really don't know what to say right now about negative reasoning. I haven't encountered the need per se, but it feels "natural" in a way to be able to search for files "not owned by my user in this directory". Any ideas of how (or if) to do negation/exclusion by owner are welcome.

I'll start to hack a prototype together, and I'm really curious as what people think about this option.

sharkdp:
Thank you very much for your feedback!

I think this request goes into a similar direction as #273 and #301 in the sense that I don't think that this is really part of the *"80% of the use cases"* (as mentioned in the README). That being said, I'm happy to consider adding this, if there is a strong support for this.

genevera:
I just ran into this needing to find and delete directories owned by a certain UID inside of a huge dir tree

sharkdp:
It looks like there *is* a strong support for this. Let's see if we can get #328 ready (some things need to be discussed first, see PR comments).

alexmaco:
@sharkdp Good to hear that you are maybe willing to merge such a feature. I think now is a good point to discuss how it should look.

#328 currently up adds 4 cli flags, with no short forms. Since I filed the issue, I have been using fd from my fork (https://github.com/sharkdp/fd/compare/v7.5.0...alexmaco:find_by_owner) with the single `-o | --owner` option (which also supports negation) (that branch still has the `-U | --user` name, we can bikeshed here). I have come to find it quite usable, so i would suggest that the feature should be exposed like that:
* using a single option
* the argument should be `chown`-like (`(user|uid):(group|gid)`), with either side of the `:` being optional
* either side of `:` can be preceded with `!` to invert the match

[...]

PR #328, opened 2018-09-20
Add `--uid`, `--gid`, `--user`, `--group`

bennyyip:
#307

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,6 +89,7 @@ dependencies = [
  "regex 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-syntax 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "users 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "version_check 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -319,6 +320,14 @@ version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "users"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "utf8-ranges"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


sharkdp:

I didn't have time to look at this in detail, so far. I still stand by my statement in #307, so I will first have to think about this. Thank you for your patience.

sharkdp:
I think I'd be in favor of this if we can somehow avoid adding four new command line options. Something like what has been suggested in #307:
```
--user [<username>][:<group>]
```

which could be used as
```
--owner sharkdp
--owner sharkdp:somegroup
--owner :somegroup
```

I wouldn't care about UID/GID personally (unless there is a good reason why this is needed).

neersighted:
Files can often be owned by a non-existent user or group (meaning only UID/GID is available). I think that using the syntax you described, but allowing numeric IDs in addition to names makes the most sense. This is basically identical to the semantics of `chown`, and very intuitive in my opinion.

tmccombs:
> I think that using the syntax you described, but allowing numeric IDs in addition to names makes the most sense. 

that makes sense to me

bennyyip:
I like the neat `--owner` syntax. As @neersighted , sometimes we need UID/GID.
So I think `<username>` should be parsed as username by default, unless it is a number, which should be parsed as a uid. Same for `<group`>, just like how find treat username and group.

sharkdp:
Sounds good to me. Thank you all for your feedback!

alexmaco:
To add a little to the discussion, I also implemented a similar feature for my own use (which you can see here https://github.com/sharkdp/fd/compare/master...alexmaco:find_by_owner).

Although i initially though that `--owner` was a good idea in principle, especially since the argument can also take a group, i found that calling it `--user` still seems more natural, in the rule-of-least-astonishment way. It could be just my preference however, and owner still seems more logical. The issue with `--user` is the collision with `-u | --unrestricted`, and having to use `-U` for the short form, which seems a little ugly, whereas `-o` was available for owner.

Second, I found that negation was easy to implement and surprisingly useful in my personal use cases, mostly as a "files not owned by me", or "files not owned by root" kind of way.

Lastly, what was surprisingly difficult to write in a simple and easy to parse manner was the help text for the `[ [!] <uid|username> ] [ : [!] <gid|groupname> ]` grammar. I still don't think i have it concise enough.

To be clear, I'm just sharing a few impressions (also from a users perspective), but this is not intended to be read as pressure for adding this feature to master (it's enough for me to update my fork sometimes). Hopefully these are useful in some way.

sharkdp:
@alexmaco Thank you for the feedback. One thing I noticed while looking through your changes is the now-obligatory call to `fs::metadata` (even when no filtering for the user is present). Note that this is an additional syscall (`lstat`) on Unix platforms which incurs some measureable overhead.

sharkdp:
Closing this as #581 has been merged. @bennyyip Thank you very much for your initial work on this. Everyone else: thank you for the discussion.

PR #331, opened 2018-09-30
enable error messages

### LINKED ISSUE #311: Issue error messages (eg permission denied)
If `fd` can't search a directory, there is no notification:

```
% mkdir -m000 noperms
% fd -HI moo noperms
%
```

I'd expect a `permission denied` by default.

If not default (why not? that's what STDERR is for), perhaps add a switch to enable them.

psinghal20:
Fixes #311.
This PR enables the display of error messages such as permission denied. It adds the `WorkerResult` enum representing the results of the worker threads which are passed through the `mpsc::channel` to the main thread.

### DIFF
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -7,7 +7,7 @@
 // according to those terms.
 
 use super::CommandTemplate;
-use internal::print_error_and_exit;
+use internal::print_error;
 use std::path::PathBuf;
 use std::sync::mpsc::Receiver;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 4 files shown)


sharkdp:
This looks great, thank you very much!

Two things:

* To be sure, I ran some benchmarks. This doesn't seem to have any measurable performance impact.
* I prefer if we just print the error but continue execution. I was just experimenting with this locally, so I will just push a commit to this PR (hope this is fine).

HaleTom:
Brilliant! Thank you! :)

PR #334, opened 2018-10-03
Add exec tests and cfg(test) compiler guards

joshleeb:
Some tests throughout the codebase were just functions marked with the `#[test]` macro. It is convention to have these test functions in a `test` module with a compiler guard `cfg(test)`. This PR updates the tests that aren't already setup like this to be in the `test` module.

Additionally, this PR also adds tests to the `exec` module to check the remaining `Token` enum variations.


sharkdp:
Thank you very much!

PR #335, opened 2018-10-03
Move exit code consts into enum

joshleeb:
Previously, the constants defined in `src/exit_codes` weren't being used, and the constants for exit codes were being redefined in the `internals` module.

This PR removes the exit code consts and instead uses an enum defined in `src/exit_codes`. This centralizes the definitions of exit codes making them easier to modify and keep track of.

### DIFF
--- a/src/exit_codes.rs
+++ b/src/exit_codes.rs
@@ -1,13 +1,13 @@
 pub enum ExitCode {
-    Error,
-    Sigint,
+    GeneralError,
+    KilledBySigint,
 }
 
 impl Into<i32> for ExitCode {
     fn into(self) -> i32 {
         match self {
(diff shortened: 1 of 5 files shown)


src/exit_codes.rs
```diff
@@ -1,4 +1,13 @@
-/// exit code 1 represents a general error
-pub const ERROR: i32 = 1;
-/// exit code 130 represents a process killed by signal SIGINT
-pub const SIGINT: i32 = 130;
+pub enum ExitCode {
```
sharkdp: Maybe while we're at it, we could make the names a little bit more descriptive. Maybe `ExitCode::GeneralError` and `ExitCode::KilledBySigint`?

PR #336, opened 2018-10-03
Add print_error macros to replace functions

joshleeb:
This patch removes the `print_error` function which was just calling `eprintln!`. Instead of making the function call we use the macro instead.

The patch also replaces the `print_error_and_exit` function with a macro `eprintln_and_die!`. This is more consistent with the print macros provided by stdlib.

### DIFF
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -30,10 +30,10 @@ pub fn execute_command(mut cmd: Command, out_perm: Arc<Mutex<()>>) {
             let _ = stderr.lock().write_all(&output.stderr);
         }
         Err(ref why) if why.kind() == io::ErrorKind::NotFound => {
-            eprintln!("fd: execution error: command not found");
+            print_error!("fd: execution error: command not found");
         }
         Err(why) => {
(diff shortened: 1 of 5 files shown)


joshleeb:
@sharkdp Bump on this.

Happy to stick with the functions rather than use the macros if you'd prefer, but thought this was more idiomatic Rust 😄

sharkdp:
Thank you very much for this contribution. Sorry for the late reply.

I like the macro idea, but I would prefer to keep the name `print_error_and_exit!(..)`.

I would also like to keep the `print_error!()` function/macro for now because it would be nice to print a (possibly red-colored) prefix "[fd error]: $msg". Especially if `--exec` is involved, a prefix like this can help to figure out where the error message came from.

What do you think?

joshleeb:
Yep that sounds good to me :) I'll update the names of these macros for this PR and add an issue to track modifying the `print_error!` macro to include colors

sharkdp:
Thank you. Could you please change the `eprintln!` statements back to `print_error!` statements?

sharkdp:
Thank you!

Issue #165, opened 2017-11-11
Search based on file timestamps (atime, ctime, mtime)


guihkx:
`find` has this very useful feature that I use often, which is timestamp-based file search. You're probably familiar with it already, but here's the documentation for those flags:

```
-atime n
              File  was  last  accessed n*24 hours ago.  When find figures out
              how many 24-hour periods ago the file  was  last  accessed,  any
              fractional part is ignored, so to match -atime +1, a file has to
              have been accessed at least two days ago.

-ctime n
              File's status was last changed n*24 hours ago.  See the comments
              for -atime to understand how rounding affects the interpretation
              of file status change times.

-mtime n
              File's  data was last modified n*24 hours ago.  See the comments
              for -atime to understand how rounding affects the interpretation
              of file modification times.
```

What do you think? Thank you.

sharkdp:
Sounds good. Do we need all three? I would guess that `-mtime` is used most of the time? Also, not all of them will be available cross-platform.w

GUiHKX:
I just use `-mtime` as well, but I suggested all of them just in case. :+1:

ChengCat:
I agree with adding support for only file modification time. And I think a more powerful syntax for describing the time duration may help, e.g. 20min, 1h, 5d. Time moment is helpful too, but that could be too ambiguous.

sharkdp:
Just an idea, but I think it would be cool if we could parse relative timespans (as well as absolute date/times) and allow for things like:
```
fd --older-than "2 weeks"
fd --changed-within "3 days"
```

Related: https://crates.io/crates/chrono-humanize

Dowwie:
an mtime feature would be a valuable addition to 'fd'

mvives-broadsign:
atime is a nice feature when you want to clean a file server :)
in GNU find, there is also ```amin, cmin and mmin``` that that does the same thing as the ```?time ``` variants but based on minutes instead of hours, (very) useful when you want smaller granularity :)

Manozco:
atime is a nice feature when you want to clean a file server :)
in GNU find, there is also ```amin, cmin and mmin``` that that does the same thing as the ```?time``` variants but based on minutes instead of hours, (very) useful when you want smaller granularity :)

sharkdp:
[...]

PR #339, opened 2018-10-09
Add support for --changed-before and --changed-with for modification …

kimsnj:
This PR aims at implementing #165 

It adds two options `changed-before` and `changed-with` that can accept either a duration (such as `10min`, `2d`, …) or a date and time (such as `2018-10-09 10:00:00`).

Some bikeshedding might be useful on the name on the options, as some of the combination reads quite well with:
- `fd --changed-within 10d`
- `fd --changed-before "2018-10-09 10:00:00"`

But the other are a bit less self-explanatory:
- `fd --changed-before 10d`
- `fd --changed-within "2018-10-09 10:00:00"`

Any thoughts ?

On a slightly different topic, as i was running `clippy` before committing my changes, I noticed some lints triggered. Would you welcome a PR that applies those?

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -108,23 +108,23 @@ pub enum TimeFilter {
 }
 
 impl TimeFilter {
-    fn from_str(s: &str) -> Option<SystemTime> {
+    fn from_str(ref_time: &SystemTime, s: &str) -> Option<SystemTime> {
         use humantime;
         humantime::parse_duration(s)
-            .map(|duration| SystemTime::now() - duration)
+            .map(|duration| *ref_time - duration)
             .or_else(|_| humantime::parse_rfc3339_weak(s))
             .ok()
     }
 
-    pub fn before(s: &str) -> Option<TimeFilter> {
-        Some(TimeFilter::Before(TimeFilter::from_str(s)?))
+    pub fn before(ref_time: &SystemTime, s: &str) -> Option<TimeFilter> {
+        TimeFilter::from_str(ref_time, s).map(TimeFilter::Before)
     }
 
-    pub fn after(s: &str) -> Option<TimeFilter> {
-        Some(TimeFilter::After(TimeFilter::from_str(s)?))
+    pub fn after(ref_time: &SystemTime, s: &str) -> Option<TimeFilter> {
+        TimeFilter::from_str(ref_time, s).map(TimeFilter::After)
     }
 
-    pub fn is_within(&self, t: &SystemTime) -> bool {
+    pub fn applies_to(&self, t: &SystemTime) -> bool {
         match self {
             TimeFilter::Before(limit) => t <= limit,
             TimeFilter::After(limit) => t >= limit,
@@ -196,7 +196,7 @@ pub struct FdOptions {
     pub size_constraints: Vec<SizeFilter>,
 
     /// Constraints on last modification time of files
-    pub modification_constraints: Vec<TimeFilter>,
+    pub time_constraints: Vec<TimeFilter>,
 }
 
 /// Print error message to stderr.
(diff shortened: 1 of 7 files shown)


src/walk.rs
```diff
@@ -289,6 +289,21 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) {
                 }
             }
 
+            // Filter out unwanted modification times
+            if !config.modification_constraints.is_empty() {
+                if_chain!{
+                    if entry_path.is_file();
+                    if let Ok(metadata) = entry_path.metadata();
+                    if let Ok(modified) = metadata.modified();
+                    if config.modification_constraints.iter().all(|tf| tf.is_within(&modified));
```
sharkdp: If we would invert the logic here by using `.iter().any(|tf| !tf.is_within(&modified)` we could profit from short-circuiting and also prevent the empty `if`/`then` clause.

tests/tests.rs
```diff
@@ -1110,3 +1113,65 @@ fn test_size() {
     // Files with size equal 4 kibibytes.
     te.assert_output(&["", "--size", "+4ki", "--size", "-4ki"], "4_kibibytes.foo");
 }
+
+#[cfg(test)]
+fn create_file_with_modified<P: AsRef<Path>>(path: P, duration_in_secs: u64) {
```
sharkdp: Awesome!

tests/tests.rs
```diff
@@ -1110,3 +1113,65 @@ fn test_size() {
...
+
+    te.assert_output(
+        &["", "--changed-before", "15min"],
+        "1_h
+        2_h
+        1_day",
+    );
+
+    te.assert_output(
+        &["min", "--changed-within", "12h"],
```
sharkdp: This test really confused me for a while because I didn't see the first `"min"` argument. Maybe you could name some of the files `foo_1_h` and others `bar_10_min` and then additionally search for `foo` here?

src/internal.rs
```diff
@@ -100,6 +100,38 @@ impl SizeFilter {
...
+    fn from_str(s: &str) -> Option<SystemTime> {
+        use humantime;
+        humantime::parse_duration(s)
+            .map(|duration| SystemTime::now() - duration)
+            .or_else(|_| humantime::parse_rfc3339_weak(s))
+            .ok()
+    }
+
+    pub fn before(s: &str) -> Option<TimeFilter> {
+        Some(TimeFilter::Before(TimeFilter::from_str(s)?))
```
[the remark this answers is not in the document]
kimsnj: I indeed find it more readable with `map` :)

src/internal.rs
```diff
@@ -100,6 +100,38 @@ impl SizeFilter {
...
+pub enum TimeFilter {
+    Before(SystemTime),
+    After(SystemTime),
+}
+
+impl TimeFilter {
+    fn from_str(s: &str) -> Option<SystemTime> {
+        use humantime;
+        humantime::parse_duration(s)
+            .map(|duration| SystemTime::now() - duration)
```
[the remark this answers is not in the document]
kimsnj: :+1:

src/walk.rs
kimsnj: `all` also benefits from the same short-circuiting (as-soon as a predicate fails, `all` returns).
in this case, it allows to have all the `return WalkState::Continue` as the else branch, and thus put them in common.

src/walk.rs
```diff
@@ -289,6 +289,21 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) {
                 }
             }
 
+            // Filter out unwanted modification times
+            if !config.modification_constraints.is_empty() {
+                if_chain!{
```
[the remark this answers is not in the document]
kimsnj: I wanted to avoid the 4 * `else { return Continue }`.
I removed the dependency and added a boolean flag instead.
I could put back the else { return } if you prefer.

src/walk.rs
sharkdp: You are right, of course :+1:

sharkdp:
Thank you very much for the updates! We can think about the command-line option names before the next release, but I'd like to get this merged :+1: 

(one option would be to add aliases `changed-after = --changed-within` and `??? == changed_before`)

PR #340, opened 2018-10-10
Split internals.rs into module

joshleeb:
This PR splits `internals.rs` into the `internal` module with multiple files, and moves `FdOptions` into `opts.rs`.

The main motivation behind this is to move logic for constructing `FdOptions` out of the main function and more readable and easier to understand in the `opts` module. The goal will eventually to be able to write `FdOptions::from(matches)` and have the options constructed.


src/output.rs
```diff
@@ -6,19 +6,22 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
+use ansi_term;
+use std::{
```
sharkdp: Wow, I didn't know this was possibly :-)

sharkdp:
Unfortunately, there are probably horrible merge conflicts now due to the merge of #339.

joshleeb:
All good about the merge conflicts :smile:

Yep there are no major code changes in this PR it's mainly just moving stuff around. 

One change which might be worth mentioning though is in `internals/mod.rs` tests where it is changed to use the macro `oss_vec!` instead of `vec!(oss(a), ...)`.

joshleeb:
Closing in favor of #341

PR #344, opened 2018-10-13
Use oss_vec fn in internal/mod.rs tests

joshleeb:
This PR creates and uses the `oss_vec!` macro instead of manually calling the `oss` function for each entry of the vecs used for testing.

### DIFF
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -99,79 +99,74 @@ where
 mod tests {
     use super::*;
 
-    fn oss(v: &str) -> OsString {
-        OsString::from(v)
+    fn oss_vec(strs: &[&str]) -> Vec<OsString> {
+        strs.into_iter().map(OsString::from).collect()
     }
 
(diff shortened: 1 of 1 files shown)


sharkdp:

Maybe that's just me, but I would generally always prefer a non-macro solution if it isn't too verbose. For example:
```rust
fn oss_vec(strs: &[&str]) -> Vec<OsString> {
    strs.into_iter().map(OsString::from).collect()
}
```

which could be used like this:
```rust
oss_vec(&["foo", "bar", "baz"])
```

What do you think?

sharkdp:
Any update on this? I'd like to either close or merge this (after updates or discussion).

joshleeb:
@sharkdp updated - apologies for the delay.

I much prefer using that function. I tend to lean towards macros when they're not totally needed (something I'm working on snapping out of 😝)

sharkdp:
Thank you!

Issue #317, opened 2018-08-14
Dependencies update: Cargo.toml not updated


paride:
I see that 59c9901cde03bfde2715a07e0c4a8ffc7b2da9c6 updates some dependencies, but Cargo.toml does not reflect these changes: it still lists the old dependencies.

Once this is fixed, may I ask you if you intend to tag a new release? I'm planning to maintain the fd-find Debian package, and with the updated dependencies it will be way easier. Thank you!

sharkdp:
Thank you for the feedback.

Yes, `Cargo.lock` was updated via `cargo update`, but no changes were made to the version boundaries in `Cargo.toml` (which is perfectly fine, I believe).

As far as I can see, all [dependencies are up-to-date](https://deps.rs/repo/github/sharkdp/fd)? Are there any particular version-updates that you have in mind?

paride:
@sharkdp You are right, and you can expect to see fd Debian packaged shortly. Unfortunately I will have to rename the binary (and the manpage) to `fdfind`, as `/usr/bin/fd` is alredy taken by another package.

About the dependencies: there is a thing I have to ask you. It doesn't seem that you are using the `wrap_help` feature of `clap`, but you still depend on it. This is a bit troublesome, as `wrap_help` depends on `term_size`, which *does* have some outdated dependencies. Could you consider droppping the dependency on `wrap_help`, if you are not planning to use it?

Thanks!

paride:
Once again I think I have misunderstood how the dependency system works. You have to excuse me: I don't actually program in Rust, I'm just trying to do the packaging work.

So `wrap_help` *does* get used. I managed to compile fd-find without it, but I'm not sure I want to release the package like this. Let's see if I can get wrap_help packaged.

sharkdp:
> Once again I think I have misunderstood how the dependency system works. You have to excuse me: I don't actually program in Rust, I'm just trying to do the packaging work.

In cargo, features are a way to change/configure the behavior of a library at compile time (here, `wrap_help` enables some code in `clap` that formats the `--help` text of `fd` in a different way). Features may [pull in other dependencies](https://github.com/clap-rs/clap/blob/3ea14b8de2cfe56d24bb2b14dd955c7daee3d87d/Cargo.toml#L49) (like `term_size` in this case).

> So `wrap_help` does get used. I managed to compile fd-find without it, but I'm not sure I want to release the package like this.

[...]

Issue #42, opened 2017-06-12
Upload to crates.io


dead10ck:
Title is self-explanatory.

sharkdp:
Thank you for the suggestion.

Unfortunately, the name `fd` is already taken by @stemjail's https://github.com/stemjail/fd-rs - https://crates.io/crates/fd

Any suggestions on how to proceed? 

Are there any other advantages, apart from being able to run `cargo install <crate-name>`?

tmccombs:
does the cargo/crate name need to be the same as the executable? If not you could name the crate something like `fd-command` or `fd-finder` or even `better-find`

dead10ck:
The crate name really doesn't matter. Any of @tmccombs ' suggestions would work. @sharkdp there's that, and also storing every old version that gets published without having to resort to something like git tags. Also, it's just a pain for users to have to look up the github URL every time.

sharkdp:
I quite like the `fd-finder` idea, but also +1 for the remark about finding an expanded name in order for this to be more memorable (and search-engine friendly).

I tried my best in expanding the `fd` acronym. Here are a few ideas (not to be taken too seriously):
* `find-deluxe` - maybe a little bit too pretentious :smile: 
* `file-detector`
* `find-data`
* `finding-dory`

Something like *ag - the silver searcher* would be cool, but *Fd* is not a chemical element :frowning_face:.

erf:
Suggestion: 
'find direct' or 'file direct(or)' or 'fd - find with rust'

sharkdp:
`fd` is now on crates.io: https://crates.io/crates/fd-find and therefore can be installed via
```
cargo install fd-find
```

Thanks!

PR #345, opened 2018-10-14
fd is now officially in Debian

sylvestre:
Work by @paride and the debian rust team!

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ Download the latest `.deb` package from the [release page](https://github.com/sh
 sudo dpkg -i fd_7.0.0_amd64.deb  # adapt version number and architecture
 ```
 
-### On Ubuntu
+### On Debian
 
 If you run Debian sid (and soon Debian Buster), fd is
 [officially maintained by Debian](https://tracker.debian.org/pkg/rust-fd-find).


README.md
```diff
@@ -132,6 +132,15 @@ Download the latest `.deb` package from the [release page](https://github.com/sh
...
 ```
 
+### On Ubuntu
+
+If you run Debian sid (and soon Debian Buster), fd is
+[officially maintained by Debian](https://tracker.debian.org/pkg/rust-fd-find).
+```
+$ sudo apt-get install fd-find
+```
+The binary is called `fdfind` as the binary name `fd` is already used by another package.
```
sharkdp: I wonder if we should recommend `alias fd=fdfind` here to make the documentation (README and man page) consistent?

sylvestre: yeah, probably. 
we are also discussing about other options like install fd in a specific path and recommending to update PATH

sylvestre:
by the way @sharkdp how would you fill about renaming fd to fdfind?
it would help on debian :)

sharkdp:
> by the way @sharkdp how would you fill about renaming fd to fdfind?
> it would help on debian :)

I would rather not. I don't really like `fdfind`/`fd-find` because it duplicates the word "find" and its abbreviation.

We have had a similar discussion in [this ticket](https://github.com/sharkdp/fd/issues/42). However, this discussion is now over a year old and I believe that the name "fd" has now made it into a few online discussions and blog posts. "fd" is also available as a package named "fd" (with a binary named "fd") in a lot of different distributions and package managers. See [this ticket](https://github.com/sharkdp/fd/issues/317#issuecomment-419555117) for a similar discussion with @paride. I would therefore like to avoid any renaming.

See also [my comment in this discussion](https://github.com/sharkdp/fd/issues/104#issuecomment-336244635):

> I knew that there was a price to pay (name clashes, harder to find) when naming this `fd`, but I wanted to have something short like `rg`, `ag`, `ls`, etc. Of course, I could have asked all users to set an `alias fd`, but somehow that doesn't seem very user-friendly to me.

sharkdp:
Also, sorry for the trouble that this causes with Debian packaging. I really appreciate your effort.

Does Ubuntu have a different policy?

paride:
@sharkdp I'm not sure, but Ubuntu is likely to inherit the Debian package at some point...

Issue #322, opened 2018-08-31
About the chinese readme.md 😊


chinanf-boy:
## repo

https://github.com/chinanf-boy/fd-zh

sharkdp:
Cool! :+1: :smile: 

Thank you very much!

sharkdp:
I'm going to close this, as I don't really know what do do here. If the chinese README is up to date and you want to add a (small, discrete) link in the README here, feel free to open a PR.

Thank you!

PR #347, opened 2018-10-18
Doc: add chinese readme link

chinanf-boy:
#322


sharkdp:
@chinanf-boy Thanks! What do you think about using the Chinese Flag emoji?

Issue #312, opened 2018-07-23
Optionally allow pattern elision


studoot:
This is related to #235 - maybe it's just me, but I search for files by just their extension often enough that when I see the warning below, I wish yet again that `fd` could fill in the match-all pattern for me...

```
Error: The search pattern '/mnt/e/code/vcpkg/' contains a path-separation character ('/') and will not lead to any search results.

If you want to search for all files inside the '/mnt/e/code/vcpkg/' directory, use a match-all pattern:

  fd . '/mnt/e/code/vcpkg/'

Instead, if you want to search for the pattern in the full path, use:

  fd --full-path '/mnt/e/code/vcpkg/'
```

Having an option to enable this feels beneficial (if only to me, I guess!). Maybe only in the case that a filtering option (such as `-e` or `-t`) has also been used?

sharkdp:
Interesting idea. I have to admit that I also see this error from time to time :smile:.

What I definitely want to avoid is any sort of ambiguity (or confusion on the user side). I think it would be quite unfortunate that the following two commands would do completely different things:

```
fd -e txt docs/  # Search for "*.txt" in the "docs" folder
fd -e txt docs   # Search current directory for entries matching "docs" and "*.txt"
```

sharkdp:
This is related to both #315 and #277.

sharkdp:
The `--path-before-pattern` doesn't feel very natural for me. I'm going to close this as "won't fix" for now. If someone comes with a better idea, I'm happy to reconsider this.

cup:
https://github.com/sharkdp/fd/issues/235
https://github.com/sharkdp/fd/issues/277
https://github.com/sharkdp/fd/issues/312
https://github.com/sharkdp/fd/issues/385
https://github.com/sharkdp/fd/pull/259
https://github.com/sharkdp/fd/pull/349

dezza:
I think the error is really badly worded, overly long and confusing. It makes no sense to me that you should have the first arg for this functionality (_go back and put current dir (.) first, what?_)

```
[fd error]: The search pattern '/home/dza/src/' contains a path-separation character ('/') and will not lead to any search results.

If you want to search for all files inside the '/home/dza/src/' directory, use a match-all pattern:

  fd . '/home/dza/src/'

Instead, if you want your pattern to match the full file path, use:

  fd --full-path '/home/dza/src/'
```

This error shouldn't exist.

It should be

```
fd -type d /home/dza/src # find all dirs
```

Like find.

[...]

PR #349, opened 2018-10-20
Add `search-path` option

### LINKED ISSUE #277: Option to explicitly set path instead of relying on argument order is missing
I believe `fd` needs an option to specify the path to be searched instead of relying on argument order in order to act as a proper `find` replacement. A lot of scripts (e.g. the popular `fzf`) can take a pre-formed `find` command as a "template" to which the search pattern is tacked on, relying on the fact that pattern comes last.

For example, `fzf` defines a variable `$dir` that can be dropped in to the search command template before the pattern is tacked on. With fd, this is unfortunately not workable since the path(s) must be the last parameter(s).

I know I can use a glob combined with `--full-path`  but then a separate search expression can't be supplied by fzf to be plugged in to the command line.

Ameobea:
Fixes #277

 * Multiple `--search-path` arguments can be supplied which replace the positional `path` argument for providing directories which will be searched.
 * When a `--search-path` argument is provided, positional `path` arguments will not be allowed

I didn't set this argument to hidden as was suggested in the issue because I feel that users wouldn't be able to know about this functionality otherwise.

Also, I didn't change the official `usage` because I feel that it would add unnecessary confusion for users that aren't looking for this specialized functionality.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -183,6 +183,14 @@ pub fn build_app() -> App<'static, 'static> {
         )
         .arg(arg("pattern"))
         .arg(arg("path").multiple(true))
+        .arg(
+            arg("search-path")
+                .long("search-path")
+                .takes_value(true)
+                .conflicts_with("path")
+                .multiple(true)
+                .number_of_values(1),
+        )
 }
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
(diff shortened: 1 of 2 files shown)


cup:
https://github.com/sharkdp/fd/pull/349
https://github.com/sharkdp/fd/issues/277
https://github.com/sharkdp/fd/issues/385

cup:
https://github.com/sharkdp/fd/issues/235
https://github.com/sharkdp/fd/issues/277
https://github.com/sharkdp/fd/issues/312
https://github.com/sharkdp/fd/issues/385
https://github.com/sharkdp/fd/pull/259
https://github.com/sharkdp/fd/pull/349

PR #350, opened 2018-10-22
Print errors when --show-errors is set

majecty:
I'm not sure `--verbose` name is good.
Maybe `--show-error` is better?

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -181,6 +181,7 @@ pub fn build_app() -> App<'static, 'static> {
                 .takes_value(true)
                 .number_of_values(1),
         )
+        .arg(arg("show-errors").long("show-errors"))
         .arg(arg("pattern"))
         .arg(arg("path").multiple(true))
         .arg(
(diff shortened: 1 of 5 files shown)


src/app.rs
```diff
@@ -181,6 +181,9 @@ pub fn build_app() -> App<'static, 'static> {
                 .takes_value(true)
                 .number_of_values(1),
         )
+        .arg(
+            arg("verbose")
```
sharkdp: I would prefer `--show-errors` (plural)

src/app.rs
```diff
@@ -314,5 +317,8 @@ fn usage() -> HashMap<&'static str, Help> {
         , "Limit results based on modification time within the duration provided:\n    \
              using a duration: <NUM>d <NUM>h <NUM>m <NUM>s (e.g. 10h, 1d, 35min...)\n    \
              or a date and time: YYYY-MM-DD HH:MM:SS");
+    doc!(h, "verbose"
+        , "Print errors to stdout"
```
sharkdp: (this should be `stderr` instead of `stdout`) I would suggest the following alternative help texts:

short:
- "Enable display of filesystem errors"

long:
- "Enable the display of filesystem errors for situations such as insufficient permissions or dead symlinks."

src/internal/opts.rs
```diff
@@ -72,4 +72,7 @@ pub struct FdOptions {
 
     /// Constraints on last modification time of files
     pub time_constraints: Vec<TimeFilter>,
+
+    /// Whether to print errors on stdout
```
sharkdp: This does not disable all errors. So the docstring should be something like: "Whether or not to display filesystem errors"

src/internal/opts.rs
```diff
@@ -72,4 +72,7 @@ pub struct FdOptions {
...
+    pub verbose: bool,
```
sharkdp: I suggest that we rename all variables from `verbose` to `show_filesystem_errors`.

majecty:
@sharkdp 
Thank you for your comments! I fixed them.

sharkdp:
Thank you very much!

Issue #274, opened 2018-03-30
Multiple instances of `subl` get launched


azdavis:
## Environment info

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.3
BuildVersion:   17D102
$ fd --version
fd 7.0.0
$ subl --version
Sublime Text Build 3143
```

## Steps to reproduce

1. Make sure Sublime Text is not running.
2. Do the following in a terminal:

```
$ mkdir test
$ cd test
$ touch {1..10}
$ fd -x subl
```

## Expected

Sublime Text gets launched, and opens the 10 files.

## Actual

Sublime Text gets launched, and opens the 10 files… four times.

The four windows you see are each from a different running instance of Sublime Text:
<img width="1280" alt="windows" src="https://user-images.githubusercontent.com/6496454/38150588-bf169828-342d-11e8-99e5-1b9933cf74e7.png">

Here's the Force Quit window as proof:

<img width="338" alt="app" src="https://user-images.githubusercontent.com/6496454/38150585-bba4fc70-342d-11e8-9ad6-88268fece891.png">

Sometimes different numbers of Sublimes get launched.

I have not tested this with other GUI apps with command-line launchers, but this issue may exist there too. Not sure if this is a fd issue, or a macOS issue, or a Sublime Text issue (I lean towards not a Sublime Text issue because I'm unable to reproduce when Sublime is already running).

sharkdp:
Thank you for the detailed report!

I can *not* reproduce this on Linux (same Sublime Text Build 3143). However, I have an idea what may be the cause of this.

`fd -x <cmd>` uses a thread pool with *N* workers to launch instances of `<cmd>` in a parallel way (one for each search result). The number of workers is determined by the number of available cores unless explicitly specified via `--threads`/`-j`.

My guess is the following: You have four cores(?) which causes `fd` to launch four instances of `subl` in parallel. For some reason, this causes Sublime Text to open four windows instead of just a single one(?) - even though it would just open a single window had there been a short delay between the different `subl` calls.

Things you could try:
- Use `fd`'s `--threads`/`-j` option to confirm if this is the root cause:
  ```
  fd -j1 -x subl
  ```

- Play with Sublime Text's `--background`/`-b` or `--wait`/`-w` options(?).
- Just for debugging purposes, use a short delay before launching `subl`:
  ``` bash
  echo 'sleep $1 && subl $1' > /tmp/launch-subl.sh
[...]

PR #360, opened 2018-11-11
Add support for batch execution of command

kimsnj:
This aims at implementing #274 

Note: for this first version, arg_max is not taken into account.

### DIFF
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -44,3 +44,16 @@ pub fn job(
         cmd.generate_and_execute(&value, Arc::clone(&out_perm));
     }
 }
+
+pub fn batch(rx: Receiver<WorkerResult>, cmd: &CommandTemplate, show_filesystem_errors: bool) {
+    let paths = rx.iter().filter_map(|value| match value {
+        WorkerResult::Entry(val) => Some(val),
+        WorkerResult::Error(err) => {
+            if show_filesystem_errors {
+                print_error!("{}", err);
+            }
+            None
+        }
+    });
+    cmd.generate_and_execute_batch(paths);
+}
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -13,7 +13,7 @@ mod job;
 mod token;
 
 use std::borrow::Cow;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::sync::{Arc, Mutex};
 
@@ -21,8 +21,17 @@ use regex::Regex;
 
 use self::command::execute_command;
 use self::input::{basename, dirname, remove_extension};
-pub use self::job::job;
+pub use self::job::{batch, job};
 use self::token::Token;
+
+/// Execution mode of the command
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ExecutionMode {
+    /// Command is executed for each path found
+    OneByOne,
+    /// Command is run for a batch of results at once
+    Batch,
+}
 
 /// Represents a template that is utilized to generate command strings.
 ///
@@ -31,10 +40,31 @@ use self::token::Token;
 #[derive(Debug, Clone, PartialEq)]
 pub struct CommandTemplate {
     args: Vec<ArgumentTemplate>,
+    mode: ExecutionMode,
 }
 
 impl CommandTemplate {
     pub fn new<I, S>(input: I) -> CommandTemplate
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        Self::build(input, ExecutionMode::OneByOne)
+    }
+
+    pub fn new_batch<I, S>(input: I) -> Result<CommandTemplate, &'static str>
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        let cmd = Self::build(input, ExecutionMode::Batch);
+        if cmd.tokens_number() > 1 {
+            return Err("Only one placeholder allowed for batch commands");
+        }
+        Ok(cmd)
+    }
+
+    fn build<I, S>(input: I, mode: ExecutionMode) -> CommandTemplate
     where
         I: IntoIterator<Item = S>,
         S: AsRef<str>,
@@ -91,7 +121,19 @@ impl CommandTemplate {
             args.push(ArgumentTemplate::Tokens(vec![Token::Placeholder]));
         }
 
-        CommandTemplate { args }
+        CommandTemplate { args, mode }
+    }
+
+    fn tokens_number(&self) -> usize {
+        self.args.iter().filter(|arg| arg.has_tokens()).count()
+    }
+
+    fn prepare_path(input: &Path) -> String {
+        input
+            .strip_prefix(".")
+            .unwrap_or(input)
+            .to_string_lossy()
+            .into_owned()
     }
 
     /// Generates and executes a command.
@@ -99,18 +141,44 @@ impl CommandTemplate {
     /// Using the internal `args` field, and a supplied `input` variable, a `Command` will be
     /// build. Once all arguments have been processed, the command is executed.
     pub fn generate_and_execute(&self, input: &Path, out_perm: Arc<Mutex<()>>) {
-        let input = input
-            .strip_prefix(".")
-            .unwrap_or(input)
-            .to_string_lossy()
-            .into_owned();
+        let input = Self::prepare_path(input);
 
         let mut cmd = Command::new(self.args[0].generate(&input).as_ref());
         for arg in &self.args[1..] {
             cmd.arg(arg.generate(&input).as_ref());
         }
 
-        execute_command(cmd, out_perm)
+        execute_command(cmd, &out_perm)
+    }
+
+    pub fn is_batch(&self) -> bool {
+        self.mode == ExecutionMode::Batch
+    }
+
+    pub fn generate_and_execute_batch<I>(&self, paths: I)
+    where
+        I: Iterator<Item = PathBuf>,
+    {
+        let mut cmd = Command::new(self.args[0].generate("").as_ref());
+        let mut paths = paths.map(|p| Self::prepare_path(&p));
+        let mut has_path = false;
+
+        for arg in &self.args[1..] {
+            if arg.has_tokens() {
+                // A single `Tokens` is expected
+                // So we can directy consume the iterator once and for all
+                for path in &mut paths {
+                    cmd.arg(arg.generate(&path).as_ref());
+                    has_path = true;
+                }
+            } else {
+                cmd.arg(arg.generate("").as_ref());
+            }
+        }
+
+        if has_path {
+            execute_command(cmd, &Mutex::new(()));
+        }
     }
 }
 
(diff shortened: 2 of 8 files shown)


src/exec/mod.rs
```diff
@@ -91,26 +121,64 @@ impl CommandTemplate {
...
+                // So we can directy consume the iterator once and for all
+                for path in &mut paths {
+                    cmd.arg(arg.generate(&path).as_ref());
+                    has_path = true;
+                }
+            } else {
+                cmd.arg(arg.generate("").as_ref());
+            }
+        }
+
```
sharkdp: If we add the following lines here:
```rust
cmd.stdin(Stdio::inherit());
cmd.stdout(Stdio::inherit());
cmd.stderr(Stdio::inherit());
```

we can use `--exec-batch` with interactive terminal commands such as `vim`.

src/exec/job.rs
```diff
@@ -44,3 +44,16 @@ pub fn job(
         cmd.generate_and_execute(&value, Arc::clone(&out_perm));
     }
 }
+
+pub fn batch(rx: Receiver<WorkerResult>, cmd: &CommandTemplate, show_filesystem_errors: bool) {
+    let paths = rx.iter().filter_map(|value| match value {
```
sharkdp: Nice!

src/exec/mod.rs
```diff
@@ -13,28 +13,58 @@ mod job;
...
 use self::command::execute_command;
 use self::input::{basename, dirname, remove_extension};
-pub use self::job::job;
+pub use self::job::{batch, job};
 use self::token::Token;
 
+/// Execution mode of the command
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ExecutionMode {
+    /// Command is executed for each path found
```
sharkdp: `path found` => `search results`?

src/exec/mod.rs
```diff
@@ -91,26 +121,64 @@ impl CommandTemplate {
             args.push(ArgumentTemplate::Tokens(vec![Token::Placeholder]));
         }
 
-        CommandTemplate { args }
+        CommandTemplate { args, mode }
+    }
+
+    fn tokens_number(&self) -> usize {
```
sharkdp: Can we rename this to `number_of_tokens` or something similar? It sounds like it returns the number of a specific token.

src/exec/mod.rs
```diff
@@ -91,26 +121,64 @@ impl CommandTemplate {
...
         let mut cmd = Command::new(self.args[0].generate(&input).as_ref());
         for arg in &self.args[1..] {
             cmd.arg(arg.generate(&input).as_ref());
         }
 
-        execute_command(cmd, out_perm)
+        execute_command(cmd, &out_perm)
+    }
+
+    pub fn is_batch(&self) -> bool {
```
sharkdp: Maybe rename this to `batch_mode` or `in_batch_mode`?

src/exec/mod.rs
```diff
@@ -91,26 +121,64 @@ impl CommandTemplate {
...
+        self.mode == ExecutionMode::Batch
+    }
+
+    pub fn generate_and_execute_batch<I>(&self, paths: I)
+    where
+        I: Iterator<Item = PathBuf>,
+    {
+        let mut cmd = Command::new(self.args[0].generate("").as_ref());
```
sharkdp: That `generate("")` call looks a bit weird. Do we expect the first argument to contain any tokens?

src/exec/mod.rs
```diff
@@ -125,6 +193,14 @@ enum ArgumentTemplate {
 }
 
 impl ArgumentTemplate {
+    pub fn has_tokens(&self) -> bool {
+        if let ArgumentTemplate::Tokens(_) = self {
```
sharkdp: if we are not capturing anything, maybe use `match` instead? Probably just a matter of taste..

tests/testenv/mod.rs
```diff
@@ -32,6 +32,9 @@ pub struct TestEnv {
 
     /// Path to the *fd* executable.
     fd_exe: PathBuf,
+
+    /// Normalize each line by splitting and sorting by whitespace as well
```
sharkdp: what does "sorting by whitespace" mean?

tests/testenv/mod.rs
```diff
@@ -121,19 +124,24 @@ fn format_output_error(args: &[&str], expected: &str, actual: &str) -> String {
...
     // Split into lines and normalize separators.
     let mut lines = s
         .replace('\0', "NULL\n")
         .lines()
         .map(|line| {
             let line = if trim_left { line.trim_left() } else { line };
-            line.replace('/', &std::path::MAIN_SEPARATOR.to_string())
+            let line = line.replace('/', &std::path::MAIN_SEPARATOR.to_string());
+            if normalize_line {
+                let mut worlds: Vec<_> = line.split_whitespace().collect();
```
sharkdp: `worlds` => `words`?

tests/testenv/mod.rs
```diff
@@ -121,19 +124,24 @@ fn format_output_error(args: &[&str], expected: &str, actual: &str) -> String {
...
+                worlds.sort();
+                return worlds.join(" ");
+            }
+            line
         })
         .collect::<Vec<_>>();
 
-    lines.sort_by_key(|s| s.clone());
-
+    lines.sort();
```
sharkdp: Haha. No idea why this was using `sort_by_key(|s| s.clone())` before...

tests/testenv/mod.rs
```diff
@@ -186,12 +203,47 @@ impl TestEnv {
...
+        cmd.current_dir(self.temp_dir.path().join(path));
+        cmd.args(args);
+
+        // Run *fd*.
+        let output = cmd.output().expect("fd output");
+
+        // Check for exit status.
+        if output.status.success() {
+            panic!(
+                "fd exited successfully. Expected error {} did not occur.",
```
sharkdp: can we use `error '{}' did not occur.` here?

tests/testenv/mod.rs
```diff
@@ -186,12 +203,47 @@ impl TestEnv {
...
     }
+
+    /// Assert that calling *fd* with the specified arguments produces the expected error.
+    pub fn assert_error(&self, args: &[&str], expected: &str) {
+        self.assert_error_subdirectory(".", args, expected)
+    }
+
+    /// Assert that calling *fd* in the specified path under the root working directory,
+    /// and with the specified arguments produces an error with the expected message.
+    fn assert_error_subdirectory<P: AsRef<Path>>(&self, path: P, args: &[&str], expected: &str) {
```
sharkdp: This is great!

tests/testenv/mod.rs
```diff
@@ -186,12 +203,47 @@ impl TestEnv {
...
+        if output.status.success() {
+            panic!(
+                "fd exited successfully. Expected error {} did not occur.",
+                expected
+            );
+        }
+
+        // Compare actual output to expected output.
+        let actual = String::from_utf8_lossy(&output.stderr);
+        if expected.len() <= actual.len() && expected != &actual[..expected.len()] {
```
sharkdp: Why not just `expected != actual`? Is this stripping the `[fd error]` prefix?

tests/tests.rs
```diff
@@ -987,6 +987,57 @@ fn assert_exec_output(exec_style: &str) {
     }
 }
 
+/// Shell script execution using -exec
+#[test]
+fn test_exec_batch() {
+    assert_exec_batch_output("--exec-batch");
+}
+
+// Shell script execution using -x
```
sharkdp: I think this is redundant. If you want to keep it, please modify the docstring to use `-X` instead of `-x`.

tests/tests.rs
```diff
@@ -987,6 +987,57 @@ fn assert_exec_output(exec_style: &str) {
...
+fn test_exec_batch_short_arg() {
+    assert_exec_batch_output("-X");
+}
+
+#[cfg(test)]
+fn assert_exec_batch_output(exec_style: &str) {
+    let (te, abs_path) = get_test_env_with_abs_path(DEFAULT_DIRS, DEFAULT_FILES);
+    let te = te.normalize_line(true);
+
+    // TODO Windows tests: D:file.txt \file.txt \\server\share\file.txt ...
```
sharkdp: I'm okay with leaving the Windows tests as To Do, but what does this comment describe?

src/exec/mod.rs
kimsnj: Maybe this is what causes the panic you mention for `fd -X "echo {}". Will have a look.

tests/testenv/mod.rs
kimsnj: *splitting* by whitespace :smile:

tests/testenv/mod.rs
kimsnj: ooops

tests/testenv/mod.rs
kimsnj: The errors regarding the conflict between `exec` and `exec-batch` is generated by `clap`… With the usage message.
This allows to have just the error message in the test.

tests/tests.rs
kimsnj: I must admit to a copy/paste (incl. this comment).
The `cfg!(windows)` clause is already present in the `exec` test, and being on linux right now, I couldn't try it out.
I guess the comment is mainly linked to the first test (absolute paths), which may have several forms on windows depending of file location (drive, network,…). But this is just a theory.

tests/testenv/mod.rs
kimsnj: That would actually be `start_with` (as usage is after the message). And it's indeed cleaner!

src/exec/mod.rs
kimsnj: It's indeed much better with these lines added.

kimsnj:
Thanks for the review!

Looking at:
```
fd -X "echo {}"
```
I only read from the `rx` iterator chain for arguments other than first one (supposed to be the executable).
So in this case, the receiver is not read at all, and thus the thread ends fast, drops `rx`… And the transmitters panics when trying to send!

I see a few options to handle this case:
- Just apply the same generation on the first argument. This would launch `echo a echo b echo c…`
- In case of a single argument in input, split by white-space before parsing the command.
- Reject entries that do not have a ArgumentTemplate::Text as first arg.

Any opinions?

I would be tempted by option 2, but not sure if there are some weird cases to consider.

sharkdp:
> Any opinions?
> 
> I would be tempted by option 2, but not sure if there are some weird cases to consider.

I would just go with option three. I think it doesn't make sense to allow placeholders in the first argument for `--exec-once` (for `--exec`, you could use something like `fd --type executable --exec "{}"` to run all executables in a directory).

src/exec/mod.rs
kimsnj: With the latest update, the first argument is not expected to have a token (checked at `new_batch`), so `input` to generate is known not to be used.

sharkdp:
Thank you very much for the updates!

kimsnj:
You're welcome!
Glad I can contribute to such a neat CLI ! :)

PR #373, opened 2018-12-23
Update info about debian & ubuntu

sylvestre:


sharkdp:
Awesome news, thank you very much!

Sorry to bring this up again, but do you know if Ubuntu has different policies than Debian with respect to the names of executables? Could it be possible to name the executable `fd` in Ubuntu?

sylvestre:
For most of the packages, Ubuntu isn't making any changes from Debian.
The Ubuntu policy is very similar to Debian's
https://people.canonical.com/~cjwatson/ubuntu-policy/policy.html/ch-files.html#s10.1

> Two different packages must not install programs with different functionality but with the same filenames. [...]. If this case happens, one of the programs must be renamed. 

Sorry

sharkdp:
Ok, Thank you!

PR #377, opened 2018-12-29
Made README text for -H and -I clearer.

### LINKED ISSUE #376: fd fails to find files named .DS_Store
Running brew-installed `fd` (version `7.2.0` on OS X `10.13.6`) I can't figure out how to convince it to find files named `.DS_Store`. `find` finds them no problem:

```sh
$ find . -name .DS_Store
./.DS_Store
./data/.DS_Store
./data/trees/.DS_Store
./data/alignments/.DS_Store
```

I've tried many combinations of args to `fd` (including `-H`, `-I`, `-F`, `-L`) and have tried giving it partial file name patterns (e.g., `tore`),  but it never prints any file names.  It's as though it doesn't even see the `.DS_Store` files in its traversal so can never print their names, regardless of arguments.

BTW, the `-H` option could be a bit better documented, as "hidden files" is quite vague. Does this mean files/dirs whose names start with `.`?

Thanks for `fd`!

terrycojones:
And, oops, I see that the `-HI` combination is already mentioned further down in `README.md`. Anyway, here are some suggestions.

Fixes #376.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,9 +257,16 @@ USAGE:
     fd [FLAGS/OPTIONS] [<pattern>] [<path>...]
 
 FLAGS:
-    -H, --hidden            Search hidden files and directories
+    -H, --hidden            Search hidden files and directories (those whose names
+                            start with a `.`). Note that if a hidden file or directory
+                            name is also mentioned in a .(git|fd)ignore file (including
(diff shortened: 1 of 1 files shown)


terrycojones:
OK, sounds good! Thanks @sharkdp.

PR #383, opened 2019-01-05
Edition upgrade

alexmaco:
This is just a minimal set of changes for the 2018 edition (since recently 1.31 is the minimum supported version).

This does not also change the `mod.rs` files into `$module_name.rs` files. Please let me know if you think this should also be done in the same PR.


sharkdp:
Thank you very much!

> This does not also change the `mod.rs` files into `$module_name.rs` files. Please let me know if you think this should also be done in the same PR.

That would be great, but we can do this in another PR.

PR #387, opened 2019-01-09
Add delete command to remove matching files and directories

gzsombor:
It's really trivial patch, but I think, this is a common task to delete the matching files.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -136,6 +136,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .conflicts_with("exec"),
         )
         .arg(
+            arg("delete")
+                .long("delete")
+                .short("D")
+                .conflicts_with_all(&["exec", "exec-batch"]),
+        )
+        .arg(
             arg("exclude")
                 .long("exclude")
                 .short("E")
@@ -304,6 +310,9 @@ fn usage() -> HashMap<&'static str, Help> {
              '{//}': parent directory\n  \
              '{.}':  path without file extension\n  \
              '{/.}': basename without file extension");
+    doc!(h, "delete"
+        , "Delete the matched files"
+        , "Delete the matched files.");
     doc!(h, "exclude"
         , "Exclude entries that match the given glob pattern"
         , "Exclude files/directories that match the given glob pattern. This overrides any \
--- a/src/main.rs
+++ b/src/main.rs
@@ -130,6 +130,10 @@ fn main() {
                 CommandTemplate::new_batch(m).unwrap_or_else(|e| {
                     print_error_and_exit!("{}", e);
                 })
+            }).or_else(|| {
+                matches.values_of("delete").map(|_m| {
+                    CommandTemplate::new(&["rm", "-r", "{}"])
+                })
             })
         });
 


gzsombor:
I thought, you need to write currently "fd -x rm \{\} ;" - based on the documentation, which I thought a bit inconvenient.

tavianator:
`fd -x rm {} \;` won't handle directories.  `fd -x rm -r {} \;` could have some quite odd effects, since the `rm` will recursively traverse a directory at the same time `fd` does.

sharkdp:
@tavianator Good point. But using `-X`/`--exec-batch` instead of `-x`/`--exec` would work. This option is new (only available on `master`) and is basically equivalent to `find`s `-exec cmd +` (although we can not yet handle the case of too many shell arguments). This would call `rm -r` just once with all the search results as arguments.

Granted, we would have to tell users of `fd` to use `fd ... -X rm -r` in cases where both directories and files are part of the search results.

I definitely now see why `find` has a `-delete` option.

Just FYI: `\;` can be omitted in `fd` and `{}` can also be omitted if the path should just be attached at the end.

tavianator:
```
$ mkdir -p foo/bar/baz
$ ./target/debug/fd . foo -X rm -r
rm: cannot remove 'foo/bar/baz': No such file or directory
```

Harmless I guess, since `foo/bar/baz` did get removed, but still.

`find -delete` turns on the `-depth` flag which traverses the directory tree in post-order (e.g. `a/b/c`, `a/b`, `a`).  Something like `fd . --post-order -X rm -d` would work reliably.  That's probably not trivial to implement though.

Maybe deleting directories and their contents doesn't fall in the 80% of use cases `fd` should support.  In which case it's maybe worth a reminder somewhere to filter out directories if you're going to use `-X rm`?  Speaking of which, is there an easy way to filter out directories?  `--type=f` is not quite the same thing.

sharkdp:
> Harmless I guess, since `foo/bar/baz` did get removed, but still.

Ahrg.. :smile: This is much more complicated than I naively thought, you are right.

> `find -delete` turns on the `-depth` flag which traverses the directory tree in post-order (e.g. `a/b/c`, `a/b`, `a`).

That is pretty cool!

(By the way, I am completely amazed by your diligence in re-implementing all of these things in [`bfs`](https://github.com/tavianator/bfs))

> Something like `fd . --post-order -X rm -d` would work reliably. That's probably not trivial to implement though.

No, definitely not :-/. The parallel traversal of the directory tree (and the parallel execution of the jobs) could potentially bring even more surprises, even if the traversal would be in post-order.

> Maybe deleting directories and their contents doesn't fall in the 80% of use cases `fd` should support.

Yes, that is often my way out :smile:. But in all seriousness, I have used `fd -x rm` to remove files quite often. I'm not sure if I have used it to remove directories, but I have certainly not experience any problems so far. I would look at this very differently if there was some way where this could actually go wrong and delete things that the user didn't intend to.

> In which case it's maybe worth a reminder somewhere to filter out directories if you're going to use `-X rm`?

Yes, I should definitely write a short chapter about deleting files with `fd`.

> Speaking of which, is there an easy way to filter out directories? `--type=f` is not quite the same thing.

There is not, unfortunately. We could consider adding support for something like `fd --type=^d`. That reminds me that I was once thinking about a very generic `--filter <cmd>` option that would have a syntax similar to `--exec` and run the given command for every file and filter results based on the return type of that command. That would allow for things like `fd --filter test ! -d`.

sharkdp:
I am going to close this PR (Sorry!).

I have added a section about deleting files in the README: https://github.com/sharkdp/fd#deleting-files

Issue #382, opened 2019-01-04
Design and refactoring discussion


alexmaco:
This is intended as a tracking issue for general code quality and design discussions. @sharkdp feel free to close this when you think quality has reached an overall satisfactory level.

To start, I'll quote a recent comment by @sharkdp from a discussion on a [PR thread](https://github.com/sharkdp/fd/pull/309#issuecomment-450738639):

> There are many things I have in mind that would hopefully help to improve the code quality:
>
> - Group related functions into `impl` blocks of appropriate `structs` such that we don't have to pass around so many parameters.
> - Think about a unified interface for filters (file type, file size, modification data, file extension, ..). 
Maybe a `trait`-based interface where the filters would be stored in a `Vec<Box<dyn Filter>>`.
> - Extract some of the functionality like the `OsStr`-based path handling into separate libraries / crates.

alexmaco:
First off, I think I can poke at the `dyn FIlter` idea. This will probably require 2 `traits`, a `Filter` and a `MetadataFilter`, the idea being that the `dyn Filter` ones will be called first, and then, if necessary, the metadata obtained and the `dyn MetadataFilter` ones called until one fails.

What I'm unsure about, is the performance aspect. A quick look around makes me suspect that, although virtual call penalty is negligible on modern/desktop CPUs with indirect branch predictors, smaller CPUs, including Atom, don't have one. So until it can be measured (hopefully on several CPUs, i only have access to some i7s), I think a defensive choice is to go with regular jumps inside a function, perhaps partitioning out the different sections for readability.

sharkdp:
Thank you for creating this ticket!

> First off, I think I can poke at the `dyn Filter` idea. This will probably require 2 `traits`, a `Filter` and a `MetadataFilter`, the idea being that the `dyn Filter` ones will be called first, and then, if necessary, the metadata obtained and the `dyn MetadataFilter` ones called until one fails.

Exactly what I was thinking!

> What I'm unsure about, is the performance aspect. A quick look around makes me suspect that, although virtual call penalty is negligible on modern/desktop CPUs with indirect branch predictors, smaller CPUs, including Atom, don't have one.

I thought about it as well, but I don't think it will be a problem. Yes, these calls will be in the "hot" part of the code, but my guess would be that we are IO-limited anyway and the virtual function call overload will hopefully be negligible. But we should definitely measure it, yes :+1:.

[...]

PR #398, opened 2019-01-26
Some cleanup and refactor

alexmaco:
Not a complete form, but this beings some cleanup and function splitting mentioned in #382 

This is not a very focused set of changes, just picking some low hanging refactoring fruit.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -84,20 +84,19 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) {
 
     for ignore_file in &config.ignore_files {
         let result = walker.add_ignore(ignore_file);
-        if let Some(err) = result {
-            match err {
-                ignore::Error::Partial(_) => (),
-                _ => {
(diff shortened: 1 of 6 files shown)


src/walk.rs
```diff
@@ -349,15 +374,4 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) {
             ignore::WalkState::Continue
         })
     });
-
-    // Drop the initial sender. If we don't do this, the receiver will block even
-    // if all threads have finished, since there is still one sender around.
-    drop(tx);
```
sharkdp: If I understand this correctly, calling `drop` explicitly is not needed anymore because `tx` is moved into the `spawn_senders` function and dropped at the end of the function scope?

sharkdp:
Awesome work, thank you very much!

src/walk.rs
alexmaco: That is correct. It's explained in the commit message of a05312e

PR #401, opened 2019-01-29
avoid cloning command, in the wake of 9d26b74

alexmaco:
What it says on the tin. Feedback welcome if changes are required.


sharkdp:
Very nice, thank you so much!

PR #402, opened 2019-01-29
Lazy metadata

alexmaco:
Functionality-wise, I think this is complete, but design-wise it might be considered experimental.

There are 2 issues i attempted to address with this:
* in scenarios with multiple filters, metadata was potentially fetched repeatedly (via `lstat`)
* some filters only apply on files, and for them the `is_file()` the check was performed on the `Path` object, which invokes `stat` (which can be even slower than `lstat`)

The proposed design only fetches the metadata once (on first use) from the entry provided by ignore. `lazy.rs` contains my rationale for slightly reinventing the wheel and not (yet) using a crate, but if anyone has better ideas suggestions are welcome.

DISCLAIMER: I did not test the performance before/after the change, nor did I count the actual syscalls performed. I am aware of `hyperfine` (awesome tool btw), but i'm looking to see if this design is acceptable first.

Let me know what you think.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -281,7 +281,9 @@ fn spawn_senders(
             // Filter out unwanted file types.
 
             if let Some(ref file_types) = config.file_types {
-                if let Some(ref entry_type) = entry.file_type() {
+                if let Ok(metadata) = entry_metadata.get() {
+                    let entry_type = metadata.file_type();
+
                     if (!file_types.files && entry_type.is_file())
                         || (!file_types.directories && entry_type.is_dir())
                         || (!file_types.symlinks && entry_type.is_symlink())
@@ -321,8 +323,8 @@ fn spawn_senders(
 
             // Filter out unwanted sizes if it is a file and we have been given size constraints.
             if !config.size_constraints.is_empty() {
-                if entry_path.is_file() {
-                    if let Ok(metadata) = entry_metadata.get() {
+                if let Ok(metadata) = entry_metadata.get() {
+                    if metadata.is_file() {
                         let file_size = metadata.len();
                         if config
                             .size_constraints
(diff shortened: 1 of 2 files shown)


sharkdp:
Very cool idea!

I didn't properly review this or spend time thinking about your questions, but out of curiosity I played around with your `Lazy` implementation and I think I found a slightly simpler variant which avoids the `Optional` (not that it would be a problem). I had to change from `FnOnce` to `FnMut`, though (which allows calling with just a reference to self). Not really sure if this would cause any limitations.

``` rust
pub struct Lazy<T, F>
where
    F: FnMut() -> T,
{
    inner: LazyImpl<T, F>,
}

impl<T, F> Lazy<T, F>
where
    F: FnMut() -> T,
{
    pub fn new(f: F) -> Self {
        Self {
            inner: LazyImpl::Pending(f),
        }
    }

    pub fn get(&mut self) -> &T {
        if let LazyImpl::Pending(ref mut obtain) = self.inner {
            self.inner = LazyImpl::Data(obtain());
        };

        match self.inner {
            LazyImpl::Data(ref data) => data,
            _ => unreachable!(),
        }
    }
}

enum LazyImpl<T, F>
where
    F: FnMut() -> T,
{
    Pending(F),
    Data(T),
}
```

sharkdp:
> I did consider using `FnMut` instead. The only technical upside of `FnOnce` is that it is allowed to move from its "stack" frame, while `FnMut` is just allowed to mutate its frame.

Ah, I see. That might be useful, yes.

> `FnOnce` also I think better expresses the idea, that a `Lazy` should only ever call the closure at most once.

Absolutely.

> I think that `Option` might also be removable by sticking a `*Cell` of some sort somewhere. It could only ever be observed to be `None` if the closure would panic, the panic would be caught, and the same `Lazy` used again, which is quite unlikely.

I was actually just curious why the `Option` was there in the first place. I don't think it necessarily has to be removed.

I did also think about `Cell` shortly because it might be worth thinking about whether or not the `Lazy::get` method should require `&mut self`. From a correctness point of view: yes. From an API point of view: maybe :smiley:.

sharkdp:
Sorry for leaving this hanging for so long.

There were some recent changes by @tavianator in #434 which also addressed metadata calls. However, this PR is probably still relevant. If we can get some benchmarks on this and resolve the merge conflicts  (I'm happy to help on both accounts), I'd be very interested in getting this merged.

PR #412, opened 2019-02-27
Update FreeBSD install instructions

t6:
Installing fd does not require using portmaster (it will also build
fd from source).  IMHO this is pretty misleading.  Let's update the
FreeBSD install instructions to recommend using the binary package
in the official repository like on other systems.


sharkdp:
Thank you!

Issue #416, opened 2019-03-07
Ubuntu .deb package (at least) leads to package being described as "proprietary" in package installer


cathalgarvey:
It's a small issue, but it's there anyway. :)

Probably there's some small metadata item that needs to be set to indicate the appropriate licenses.

sharkdp:
Thank you for the feedback. If someone knows how to fix this, please let us know.

PR #419, opened 2019-03-09
Add copyright file for deb package

spk:
Hi this should fix https://github.com/sharkdp/fd/issues/416; cheers

### DIFF
--- a/ci/before_deploy.bash
+++ b/ci/before_deploy.bash
@@ -103,7 +103,36 @@ Source: $homepage
 
 Files: *
 Copyright: $maintainer
+License: Apache-2.0 or MIT
+
 License: Apache-2.0
+ On Debian systems, the complete text of the Apache-2.0 can be found in the
+ file /usr/share/common-licenses/Apache-2.0.
+
+License: MIT
+ Permission is hereby granted, free of charge, to any
+ person obtaining a copy of this software and associated
+ documentation files (the "Software"), to deal in the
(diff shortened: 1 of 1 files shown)


ci/before_deploy.bash
```diff
@@ -91,6 +96,15 @@ make_deb() {
...
     install -Dm644 LICENSE-MIT "$tempdir/usr/share/doc/$PROJECT_NAME/LICENSE-MIT"
     install -Dm644 LICENSE-APACHE "$tempdir/usr/share/doc/$PROJECT_NAME/LICENSE-APACHE"
+    cat > "$tempdir/usr/share/doc/$PROJECT_NAME/copyright" <<EOF
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: $PROJECT_NAME
+Source: $homepage
+
+Files: *
+Copyright: $maintainer
+License: Apache-2.0
```
sharkdp: Is it possible to add the dual license here? `fd` is licensed under either of Apache or MIT.

sharkdp:
Thank you very much!

ci/before_deploy.bash
spk: Ah ok, yes it is found how in https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ applied in d5daae3948ada562e1ee8f7127b53a5de04f3715

Issue #428, opened 2019-04-10
Support ripgrep's --path-separator option


mookid:
On windows (both cygwin and cmd.exe shells) the output is with backslashes as path separator. However, one sometimes would want to interop with windows port of unix tools and get slash-separated paths.

When using ripgrep, one can override the path separator with the `--path-separator` option. I think it would be great to support it in fd as well.

This should also allow working around issues like #153.

reference: the ripgrep discussion https://github.com/BurntSushi/ripgrep/issues/275

vvuk:
What happened to this in #429?  I see that PR got closed without comment.

mookid:
I did not see a great interest for it. I would be happy to see this merged, but I don't want to keep pinging.

When I need the feature, I use my fork.

sharkdp:
I think there is some interest for it (see also the discussion in #153).

If you closed the PR because there was no response from my side, let me assure you that it was only a lack of time, not lack of interest 😄

sharkdp:
Fixed in #429 by @mookid

rajeshduggal:
Great! When can we expect to see a windows release which includes this? https://github.com/sharkdp/fd/releases

sharkdp:
Released in [v7.4.0](https://github.com/sharkdp/fd/releases/tag/v7.4.0).

PR #429, opened 2019-04-10
Add path-separator option.

mookid:
Example usage: `fd.exe --path-separator /` on windows.

- [x] add tests
- [x] measure performance impact without path-separator option
- [x] measure performance impact with path-separator option

## benchmarks

Average over 15 runs (with warm cache), using the command `fd . %userprofile%` [(full results):](https://docs.google.com/spreadsheets/d/1IApnoBDVvL873xqoI8CkaXjE4Y0n0stEZyDgKP9Lfjc/edit?usp=sharing)

  | feature, sep=default | feature, sep='=' | master
-- | -- | -- | --
avg | 5942.6667 | 6363.6000 | 5885.4000
std | 105.8730 | 109.2185 | 115.7299
std / avg | 0.0178 | 0.0172 | 0.0197
  |   |   |  
compared to master | 1.01 | 1.08 | 1.00

without changing the separator, the change does not impact performance; when changing it, it seems to be ~8% slower. 

see #428

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -210,8 +210,9 @@ pub fn build_app() -> App<'static, 'static> {
         .arg(
             arg("path-separator")
                 .takes_value(true)
+                .value_name("separator")
                 .long("path-separator")
-                .number_of_values(1),
+                .hidden_short_help(true),
         )
         .arg(arg("path").multiple(true))
         .arg(
--- a/src/app.rs
+++ b/src/app.rs
@@ -207,6 +207,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .hidden_short_help(true),
         )
         .arg(arg("pattern"))
+        .arg(
+            arg("path-separator")
+                .takes_value(true)
+                .long("path-separator")
+                .number_of_values(1),
+        )
         .arg(arg("path").multiple(true))
         .arg(
             arg("search-path")
@@ -249,6 +255,10 @@ fn usage() -> HashMap<&'static str, Help> {
     doc!(h, "absolute-path"
         , "Show absolute instead of relative paths"
         , "Shows the full path starting from the root as opposed to relative paths.");
+    doc!(h, "path-separator"
+        , "Set the path separator to use when printing file paths."
+        , "Set the path separator to use when printing file paths. \
+           A path separator is limited to a single byte.");
     doc!(h, "follow"
         , "Follow symbolic links"
         , "By default, fd does not descend into symlinked directories. Using this flag, symbolic \
--- a/src/app.rs
+++ b/src/app.rs
@@ -258,7 +258,8 @@ fn usage() -> HashMap<&'static str, Help> {
         , "Shows the full path starting from the root as opposed to relative paths.");
     doc!(h, "path-separator"
         , "Set the path separator to use when printing file paths."
-        , "Set the path separator to use when printing file paths.");
+        , "Set the path separator to use when printing file paths. The default is the OS-specific \
+           separator ('/' on Unix, '\\' on Windows).");
     doc!(h, "follow"
         , "Follow symbolic links"
         , "By default, fd does not descend into symlinked directories. Using this flag, symbolic \
--- a/src/app.rs
+++ b/src/app.rs
@@ -258,8 +258,7 @@ fn usage() -> HashMap<&'static str, Help> {
         , "Shows the full path starting from the root as opposed to relative paths.");
     doc!(h, "path-separator"
         , "Set the path separator to use when printing file paths."
-        , "Set the path separator to use when printing file paths. \
-           A path separator is limited to a single byte.");
+        , "Set the path separator to use when printing file paths.");
     doc!(h, "follow"
         , "Follow symbolic links"
         , "By default, fd does not descend into symlinked directories. Using this flag, symbolic \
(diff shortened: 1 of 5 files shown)


src/app.rs
```diff
@@ -207,6 +207,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .hidden_short_help(true),
         )
         .arg(arg("pattern"))
+        .arg(
+            arg("path-separator")
+                .takes_value(true)
+                .long("path-separator")
+                .number_of_values(1),
```
sharkdp: is this needed?

src/app.rs
```diff
@@ -207,6 +207,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .hidden_short_help(true),
         )
         .arg(arg("pattern"))
+        .arg(
+            arg("path-separator")
```
sharkdp: we should probably also add `.hidden_short_help(true)`. I don't think it is important enough to be shown in the concise overview (`fd -h`).

src/app.rs
```diff
@@ -207,6 +207,12 @@ pub fn build_app() -> App<'static, 'static> {
...
+                .takes_value(true)
```
sharkdp: can we also add `.value_name("separator")`?

src/app.rs
```diff
@@ -249,6 +255,10 @@ fn usage() -> HashMap<&'static str, Help> {
     doc!(h, "absolute-path"
         , "Show absolute instead of relative paths"
         , "Shows the full path starting from the root as opposed to relative paths.");
+    doc!(h, "path-separator"
+        , "Set the path separator to use when printing file paths."
+        , "Set the path separator to use when printing file paths. \
+           A path separator is limited to a single byte.");
```
sharkdp: Maybe "limited to a single character" would be better for users that don't immediately associated 1 byte with 1 character. I know that it's probably not quite correct technically (multi-byte characters).

That makes me wonder: why not allow for general Unicode characters? I would change `u8` to `char` and change the help text to:

```
Set the path separator to use when printing file paths.
The separator must be a single character like '/' or '\'.
```

src/internal/opts.rs
```diff
@@ -75,4 +75,7 @@ pub struct FdOptions {
 
     /// Whether or not to display filesystem errors
     pub show_filesystem_errors: bool,
+
+    /// The separator used to print file paths.
+    pub path_separator: Option<u8>,
```
sharkdp: why not `char` (see above)

src/main.rs
```diff
@@ -113,6 +113,20 @@ fn main() {
         _ => atty::is(Stream::Stdout),
     };
 
+    let path_separator: Option<u8> = match matches.value_of("path-separator") {
```
sharkdp: Is the type annotation needed here?

src/main.rs
```diff
@@ -113,6 +113,20 @@ fn main() {
...
+        Some(sep_str) => {
+            let sep = sep_str.as_bytes();
+            if sep.len() != 1 {
+                print_error_and_exit!(
+                    "'{}' is not a valid path separator. See 'fd --help'.",
```
sharkdp: Instead of referring to the help text, maybe just say explicitly what the problem is: *"The argument to --path-separator must be a single character"*.

src/main.rs
```diff
@@ -113,6 +113,20 @@ fn main() {
...
+                    sep_str
+                );
+            }
+            Some(sep[0])
+        }
+        None => None,
```
sharkdp: Actually, I think that case can not happen (since you require `--path-separator` to have a value). If this is the case, I think you can simply unwrap the Option:
```
matches.value_of("path-separator").except("path-separator")
```

src/output.rs
```diff
@@ -67,7 +67,21 @@ fn print_entry_colorized(
             .map(Style::to_ansi_term_style)
             .unwrap_or(default_style);
 
-        write!(stdout, "{}", style.paint(component.to_string_lossy()))?;
+        let path_string = component.to_string_lossy();
+
+        match config.path_separator {
+            None => write!(stdout, "{}", style.paint(path_string))?,
+            Some(sep) => {
```
sharkdp: A couple of things:
* This should probably be in a separate function
* I don't think this is the best way to solve this (convert to byte vector, loop over vector, convert to string). Why not use a string-replacement function?
* Instead of checking for `cfg!(windows)`, this code should replace `std::path::MAIN_SEPARATOR` with `config.path_separator`. Otherwise, on Windows, you would replace both `/` and `\` with the new separator.

src/main.rs
mookid: I confirm that not passing the argument triggers the None case.

src/app.rs
mookid: Honestly, I don't really know what I am doing here; I am inexperienced with clap and tried to find something that works ok.

src/output.rs
mookid: If we are to use a string replacement, we can also handle any string replacement; it would make the code simpler. Do you have anything against it?

mookid:
> Printing paths is not the only place where we deal with paths. We should search the whole code base for std::path::MAIN_SEPARATOR and see if something needs to be done.

I am a bit confused: this PR deals only with the presentation logic, that is why the changes are bound to output.rs + command line arguments.

As far as I can tell the rest of the logic has no issue on windows.

mookid:
@sharkdp I think I addressed most points. Tell me if you think anything should be added.

mookid:
re @sharkdp 

do you think that anything prevents merging?

mookid:
ping!

Issue #432, opened 2019-04-22
fd performs worse than find on NFS


sbocq:
I have a script that scrubs broken transfer files on a live FTP stored on a NFS share using `find` like this:
```
find /mnt/ftp/home/redacted ! -newermt "1 day ago" -type f -name '.in.*' -or -name '*.part'
```
As this can take quite some time I decided to put `fd` to the test. Unfortunately, its performance is worse on my use case (I hope I didn't make any mistake in the translation) and I think it might be amplified by NFS, which makes some system calls more expensive than others.

Here is a first comparison on `CentOS 7` using `fd-v7.3.0-x86_64-unknown-linux-musl.tar.gz`:

```
$ time find /mnt/ftp/home/redacted ! -newermt "1 day ago" -type f -name '.in.*' -or -name '*.part'

real	0m0.259s
user	0m0.202s
sys	0m0.032s

$ time ./fd -H '^\.in|\.part$' -s --type f --changed-before 1day /mnt/ftp/home/redacted

real	0m9.970s
user	0m3.297s
sys	0m9.209s
```

I tried with the options `-j 1` a while later to be more equivalent to `find`, which is single threaded, but it didn't help:
```
$ time find /mnt/ftp/home/redacted ! -newermt "1 day ago" -type f -name '.in.*' -or -name '*.part'

real	2m59.994s
user	0m1.806s
sys	0m11.001s

$ time ./fd -H '^\.in|\.part$' -j 1 -s --changed-before 1day /mnt/ftp/home/redacted

real	6m16.614s
user	0m2.533s
sys	0m26.904s
```

As you can see, benchmarking on NFS is a bit wild depending on if metadata is cached or not on the client or server side. But I went on collecting more evidence using `strace` with and without tracing the calls to `futex`:

```
$ strace -c -f ./fd -H '^\.in|\.part$' -j 1 -s --changed-before 1day /mnt/ftp/home/redacted
strace: Process 9451 attached
strace: Process 9452 attached
strace: Process 9454 attached
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 72.60   80.094044    20023511         4           futex
 17.18   18.957540          81    234691        55 stat
  7.56    8.343416        2526      3303           getdents64
  2.52    2.783272         113     24661           lstat
  0.11    0.116978          29      4072           brk
  0.02    0.026199         105       250       202 open
  0.00    0.002726         101        27           madvise
[...]

PR #434, opened 2019-04-26
Check the pattern before anything else, since it doesn't require metadata

tavianator:
This should partially address #432 by decreasing the number of stat() calls:

    $ strace -c -f ./fd-before '.h$' -j1 /usr -S +1k >/dev/null
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
    100.00    3.700169           3    983938     38022 stat
    $ strace -c -f ./fd-after '.h$' -j1 /usr -S +1k >/dev/null
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
    100.00    0.671723           4    162052     38021 stat

Though it's not as good as possible:

    $ strace -c -f find /usr -name '*.h' -size +1k >/dev/null
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
    18.75    0.449866           3    136199           newfstatat
    $ strace -c -f bfs /usr -name '*.h' -size +1k >/dev/null
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
    15.01    0.216024           3     60211         1 statx

Performance is much better when metadata is required:

    $ hyperfine ./fd-{before,after}" '.h$' -j1 /usr -S +1k"
    Benchmark #1: ./fd-before '.h$' -j1 /usr -S +1k
      Time (mean ± σ):      2.707 s ±  0.042 s    [User: 890.8 ms, System: 1939.7 ms]
      Range (min … max):    2.659 s …  2.786 s    10 runs

    Benchmark #2: ./fd-after '.h$' -j1 /usr -S +1k
      Time (mean ± σ):      1.562 s ±  0.034 s    [User: 726.2 ms, System: 957.9 ms]
      Range (min … max):    1.536 s …  1.648 s    10 runs

    Summary
      './fd-after '.h$' -j1 /usr -S +1k' ran
        1.73 ± 0.05 times faster than './fd-before '.h$' -j1 /usr -S +1k'

While remaining the same when it's not:

    tavianator@graviton $ hyperfine ./fd-{before,after}" '.h$' -j1 /usr"
    Benchmark #1: ./fd-before '.h$' -j1 /usr
      Time (mean ± σ):      1.341 s ±  0.016 s    [User: 664.3 ms, System: 761.2 ms]
      Range (min … max):    1.309 s …  1.361 s    10 runs

    Benchmark #2: ./fd-after '.h$' -j1 /usr
      Time (mean ± σ):      1.338 s ±  0.012 s    [User: 684.1 ms, System: 741.1 ms]
      Range (min … max):    1.310 s …  1.350 s    10 runs

    Summary
      './fd-after '.h$' -j1 /usr' ran
        1.00 ± 0.02 times faster than './fd-before '.h$' -j1 /usr'


sharkdp:
Thank you very much for your contribution and for the detailed benchmarks. This looks great.

I want to take a more detailed look before I merge this, but this will probably take me a few days.

A small thing I noticed: You are comparing `find|bfs /usr -name '*.h' -size +1k` with the following `fd` call:
``` bash
fd '.h$' -j1 /usr -S +1k
```
It will probably not make a huge difference, but the `.` will actually match *any* character instead of a literal `.` character. So this will also find `.sh` files, for example. In addition, `fd` uses smart-case by default, so it will also match against a capital `H` as last character. To make the comparison fair, I think you'd have to use the `--case-sensitive`/`-s` option and escape the `.`:
``` bash
fd --case-sensitive '\.h$' -j1 /usr -S +1k
```
Even easier, if your are just searching for a file extension, we can use the `--extension`/`-e` option:
``` bash
fd . -e h -j1 /usr -S +1k
```

tavianator:
All good points.  I updated the commit message with some improved benchmarks (on a different computer), and also moved the extension check up with the pattern check.

Issue #450, opened 2019-06-09
How to adjust the job pool size for parallel execution is unclear


detly:
I would like to increase the job pool for parallel execution of my `-x` command beyond the number of cores to take advantage of my job being IO bound.

The [documentation](https://github.com/sharkdp/fd/blob/master/README.md) says:

> If the `-x/--exec` option is specified alongside a command template, a job pool will be created for executing commands in parallel for each discovered path as the input.

The man page says:

```text
       -j, --threads num
              Number of threads to use for searching (default: number of available CPUs).
```

It's not clear whether the two are related.

sharkdp:
Thank you for reporting this. The documentation should be improved.

Currently, `-j, --threads num` sets both: the number of threads used for searching AND the number of threads used for execution. So you can definitely use it to increase the job pool size.

sharkdp:
This was added in #454 by @detly.

PR #454, opened 2019-06-11
Added detail on parallel execution job pool to man page and README.

detly:
Proposed for #450 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ fd -e flac -x ffmpeg -i {} -c:a libopus {.}.opus
 fd -x wc -l \; -e rs
 ```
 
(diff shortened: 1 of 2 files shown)


sharkdp:

The man page is usually kept in sync with the `--help` text of `fd`. The `--help` text acually already mentions it:
```
    -j, --threads <num>
            Set number of threads to use for searching & executing
            (default: number of available CPU cores)
```

Also, I think we should probably not use the word "job pool". I'm not sure if all users will be familiar with the term.

detly:
Yep, that works too. I totally missed it.

detly:
Redid it so the wording is consistent between the help text and man page.

sharkdp:
Thank you very much!

PR #455, opened 2019-06-14
Add instructions for habitat

jasonwbarnett:
Habitat is a way to... Codify how the application is built, how it runs, and all of its dependencies to free the app from underlying infrastructure and make updates easy.

It's a way to install packages completely decoupled from the underlying OS and its package manager (if it even has one).


sharkdp:

Could you please add a short description? I'm not familiar with habitat. What is it? Who use likely to use it instead of another packaging solution?

PR #466, opened 2019-08-12
Add a line about completions

sylvestre:


sharkdp:

I'd rather not add this line in this list. I don't really think that this is a distinguishing feature of `fd`.

sylvestre:
No worries :)

Issue #92, opened 2017-10-10
Options -u (and -uu) as in ripgrep


unhammer:
As mentioned in https://github.com/sharkdp/fd/issues/33 , `-u` as an alias for `--no-ignore` and `-uu` as an alias for `--no-ignore --hidden` would make it easier/less surprising to use this alongside ripgrep.

sharkdp:
I'm going to copy over what I said in #33:

I would like to keep the command-line interface as clean and simple as possible (i.e., I don't want the `-h` text to be several pages long), but adding `-u` (and `-u -u`) seems like a good idea. We could possible hide it from the standard help text and add it as a hidden option that would only be mentioned in a man page(?)

unsignedint:
I would like to take this on, if I may :)

sharkdp:
Sure!

n8henrie:
For some reason, I can never remember the `fd` flags to "search everything" when I'm not finding something I *know* should be there, and today for some reason (to my delight) I tried the `ripgrep` flag `-uuu`, and it worked!

However, not finding any documentation about it in the either the `--help` or manpage was a little disconcerting, which led me to the relevant PR and this discussion.

> We could possible hide it from the standard help text and add it as a hidden option that would only be mentioned in a man page(?)

Should something about this be added to the manpage as suggested above?

sharkdp:
Sounds like a good idea, yes.

n8henrie:
Is https://github.com/sharkdp/fd/blob/master/doc/fd.1 manually written? https://github.com/sharkdp/fd/commit/ddfa0873fb42029922a31565fe142fd645970963

I see `-uuu` [appropriately documented in the source code](https://github.com/sharkdp/fd/blob/630749f706957635e982267ed9ef3bb2fc9e5e59/src/app.rs#L53), it's just not in the man page, and even building from source it's not showing up with `--help`.

I think I can figure out the syntax if this file is just built by hand, but I'd be a little surprised if that was the case.

sharkdp:
> Is https://github.com/sharkdp/fd/blob/master/doc/fd.1 manually written? [ddfa087](https://github.com/sharkdp/fd/commit/ddfa0873fb42029922a31565fe142fd645970963)

No. It's mostly auto-generated.

[...]

PR #483, opened 2019-09-16
Add `-u` to long `--help`

n8henrie:
Fixes sharkdp/fd#92

Does not yet add to the manpage though.


sharkdp:
Thank you very much. Maybe we should also work on the `--help` text for `-u`? 

Something like
```
A single occurrence of '-u' is an alias for '--no-ignore'. Two occurrences '-uu' serve as an alias for '--no-ignore' and '--hidden'.
```

PR #490, opened 2019-10-01
Update man page

### LINKED ISSUE #408: Man page doesn't explain search patterns
The man page doesn't say what form search patterns should take. The README here on GitHub says they are regular expressions, but not what kind of REs (I guess they are not PCREs, because there's an issue for adding that).

Suggestion: document in man page that the pattern argument is a regular expression, and say what kind of regular expression ops are supported.

rootbid:
man page updated for explaining search patterns.
- [x] Add syntax patterns for RE.
- [x] Add examples with RE syntax.

This commit resolves #408

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -234,9 +234,15 @@ path without file extension
 basename without file extension
 .RE
 .SH PATTERN SYNTAX
-The
-.RI [ pattern ]
-syntax supported in the tool is documented at https://docs.rs/regex/1.0.0/regex/#syntax
+The regular expression syntax used by fd is documented here:
+
+.UR https://docs.rs/regex/1.0.0/regex/#syntax
+.UE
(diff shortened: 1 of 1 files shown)


sharkdp:

It will probably take a few days until I find the time to review it.

sharkdp:
To be honest, I'm not sure we should "clutter" the man page with a full description of regular expression syntax. I'd rather just refer to https://docs.rs/regex/#syntax which includes an up-to-date description of the supported syntax.

rootbid:
I do agree as this would mean updating the syntax on the man page as per original revisions.
So, do you think a link directing to the supported syntax on the man page would suffice?

rootbid:
I hope this does it.

sharkdp:
Thank you very much

Issue #488, opened 2019-09-26
--hidden flag is not idempotent


laur89:
Creating an alias for fd providing the `--hidden` flag has been recommended here for numerous times.

I did that, but now am seeing problems when another shell abstraction (eg fzf's `FZF_DEFAULT_COMMAND` env var) is providing `--hidden` as well, causing fd command to end up with multiple `--hidden` flags, resulting in

    error: The argument '--hidden' was provided more than once, but cannot be used multiple times

error. Now, said command _could_ bypass alias by prefixing fd with `command` (at least in bash), but generally such flags are allowed to be provided many times.

Is there a reason for current behavior?

Running ver `7.0.0`

sharkdp:
Thank you for the feedback!

> Is there a reason for current behavior?

No. We should change it. This should be easy to implement. We need to use `clap`s [`overrides_with`](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.overrides_with) option:
``` rust
arg("hidden").long("hidden").short("H").overrides_with("hidden")
```

This should probably also be done for some other similar arguments.

rootbid:
I am hoping this resolves the mentioned issue.
@sharkdp, you also mentioned overriding some other similar arguments. Can you name those arguments for me to implement?

sharkdp:
@rootbid Thank you for looking into this. Similar arguments would essentially be all flags (the ones which do not take values):
* `--no-ignore`
* `--no-ignore-vcs`
* `--case-sensitive`
* `--ignore-case`
* `--glob`
* `--regex`
* `--fixed-strings`
* `--absolute-path`
* `--follow`
* `--full-path`
* `--print0`
* `--show-errors`

Care must be taken that `--ignore-case` still overwrites with `--case-sensitive` and similar restrictions.

sharkdp:
Released in [fd 7.5.0](https://github.com/sharkdp/fd/releases/tag/v7.5.0)

PR #496, opened 2019-10-08
Fix idempotent flags

rootbid:
Add 'overrides_with' to format flags.
Resolves sharkdp/fd#488


sharkdp:
Thank you very much. Could you please run `cargo fmt` to fix the CI error.

Issue #507, opened 2019-11-13
No option to limit search to one file system


FallenWarrior2k:
Many tools surrounding file system traversal include an option to disable crossing file system boundaries. find includes the `-xdev` option, which accomplishes this. fd, however, seems to lack such a feature at the present time. I've looked through the man page and the issue list and was unable to find anything relating to this.

Looking at the docs of the ignore crate, implementation should be quite simple as well: it only requires adding a call to `.same_file_system(config.same_file_system)` to the [`WalkBuilder`](https://github.com/sharkdp/fd/blob/master/src/walk.rs#L72), assuming a corresponding option `same_file_system: bool` is added to `FdOptions`, with corresponding changes in `build_app()` and `main()`.

sharkdp:
see also: #227 #74 #89 

Didn't know about the `.same_file_system` option, though.

FallenWarrior2k:
Oh, sorry. Seems my search terms weren't precise enough.

However, since this uses an existing feature of the ignore crate, there _shouldn't_ be any performance degradation if the option is not used, since the ignore crate defaults it to false anyway. Since it is basically just adding a public API for a feature that already existed, albeit unused, I wonder if you'd consider adding it.

If not, feel free to close this and the corresponding PR.

sharkdp:
> However, since this uses an existing feature of the ignore crate, there _shouldn't_ be any performance degradation if the option is not used, since the ignore crate defaults it to false anyway. Since it is basically just adding a public API for a feature that already existed, albeit unused, I wonder if you'd consider adding it.

Ok, that sounds reasonable. I'll take a look at the PR soon, thank you.

sharkdp:
Released in [fd 7.5.0](https://github.com/sharkdp/fd/releases/tag/v7.5.0)

PR #508, opened 2019-11-15
Add option to limit walk to one file system

FallenWarrior2k:
As I described in #507, fd currently lacks a way to restrict a search to a single file system. Since, in my opinion, it's a very useful feature to have and the required code changes were fairly minimal, I decided to prepare this as a kind of "working draft" in case the feature is considered for addition.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -286,9 +286,14 @@ pub fn build_app() -> App<'static, 'static> {
 
     // Make `--one-file-system` available only on Unix and Windows platforms, as per the
     // restrictions on the corresponding option in the `ignore` crate.
+    // Provide aliases `mount` and `xdev` for people coming from `find`.
     // It's not pretty, but I'm unaware of a way to make just part of a builder chain conditional
     if cfg!(unix) || cfg!(windows) {
-        app.arg(arg("one-file-system").long("one-file-system"))
+        app.arg(
+            arg("one-file-system")
+                .long("one-file-system")
+                .aliases(&["mount", "xdev"]),
+        )
     } else {
         app
     }
--- a/src/app.rs
+++ b/src/app.rs
@@ -477,7 +477,7 @@ fn usage() -> HashMap<&'static str, Help> {
             , "Don't cross file system boundaries"
             , "By default, fd will traverse the file system tree as far as other options dictate. \
                With this flag, fd ensures that it does not descend into a different file system \
-               than the one it started in.");
+               than the one it started in. Comparable to the -mount or -xdev filters of find(1).");
     }
 
     h
--- a/src/app.rs
+++ b/src/app.rs
@@ -37,7 +37,7 @@ pub fn build_app() -> App<'static, 'static> {
             .long_help(helps[name].long)
     };
 
-    App::new("fd")
+    let app = App::new("fd")
         .version(crate_version!())
         .usage("fd [FLAGS/OPTIONS] [<pattern>] [<path>...]")
         .setting(AppSettings::ColoredHelp)
@@ -115,7 +115,6 @@ pub fn build_app() -> App<'static, 'static> {
                 .alias("dereference")
                 .overrides_with("follow"),
         )
-        .arg(arg("one-file-system").long("one-file-system"))
         .arg(
             arg("full-path")
                 .long("full-path")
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -115,6 +115,7 @@ pub fn build_app() -> App<'static, 'static> {
                 .alias("dereference")
                 .overrides_with("follow"),
         )
+        .arg(arg("same-file-system").long("same-file-system"))
```
sharkdp: ripgrep calls this option `--one-file-system`. I think we should use the same name.

src/app.rs
```diff
@@ -321,6 +322,11 @@ fn usage() -> HashMap<&'static str, Help> {
         , "Follow symbolic links"
         , "By default, fd does not descend into symlinked directories. Using this flag, symbolic \
            links are also traversed.");
+    doc!(h, "same-file-system"
+        , "Don't cross file system boundaries (only Unix/Windows)"
+        , "By default, fd will traverse the file system tree as far as other options dictate. \
+           With this flag, fd ensures that it does not descend into a different file system than \
+           the one it started in. Does nothing if not on Unix or Windows.");
```
sharkdp: Maybe we could also mention that this is similar to `find`s `-mount` option. That would be useful for users that look for `--mount`. We could even provide `--mount` as a hidden alias.

src/walk.rs
```diff
@@ -78,6 +78,8 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<FdOptions>) -
         .git_exclude(config.read_vcsignore)
         .overrides(overrides)
         .follow_links(config.follow_links)
+        // Same file system is only supported on Unix and Windows platforms
+        .same_file_system(config.same_file_system && (cfg!(unix) || cfg!(windows)))
```
sharkdp: I don't think that this is the best place to add the `cfg!(unix) || cfg!(windows)` part.

Ideally, we would disable the availability of that option completely for unsupported platforms.

If that doesn't work, I'd rather add this part to `main.cpp`.

tests/tests.rs
```diff
@@ -555,6 +555,34 @@ fn test_follow() {
...
+// Limited to Unix because, to the best of my knowledge, there is no easy way to test a use case
+// file systems mounted into the tree on Windows.
+// Not limiting depth causes massive delay under Darwin, see BurntSushi/ripgrep#1429
+#[test]
+#[cfg(unix)]
+fn test_file_system_boundaries() {
+    // Can't simulate file system boundaries
+    let te = TestEnv::new(&[], &[]);
+
+    // /dev/null should exist in all sane Unixes
```
sharkdp: I like the way the corresponding test in the `ignore` crate is written:
https://github.com/BurntSushi/ripgrep/blob/8892bf648cfec111e6e7ddd9f30e932b0371db68/ignore/src/walk.rs#L2041-L2073

We should simply skip this test if `/dev/null` doesn't exist OR if `/dev/null` happens to be on the same file system.

sharkdp:

src/walk.rs
FallenWarrior2k: It's not pretty, but I've adjusted the code so that the option only exists on Unix and Windows platforms.

src/app.rs
FallenWarrior2k: I've decided to provide both `--mount` and `--xdev` as aliases, since `--xdev` is the preferred name, at least according to the man page of GNU find(1).

src/app.rs
FallenWarrior2k: Changed for both CLI and internal options.

tests/tests.rs
FallenWarrior2k: I borrowed the Unix `device_num()` function from the linked test, the test now `return`s if one of those conditions is met.

FallenWarrior2k: On second thought, I should probably credit them in a comment.

FallenWarrior2k:
If you want me to squash some of those commits together (e.g. the ones that are just renames), just tell me and I'll take care of it.

src/app.rs
```diff
@@ -282,7 +282,21 @@ pub fn build_app() -> App<'static, 'static> {
                 .multiple(true)
                 .hidden_short_help(true)
                 .number_of_values(1),
+        );
+
+    // Make `--one-file-system` available only on Unix and Windows platforms, as per the
+    // restrictions on the corresponding option in the `ignore` crate.
+    // Provide aliases `mount` and `xdev` for people coming from `find`.
+    // It's not pretty, but I'm unaware of a way to make just part of a builder chain conditional
```
sharkdp: I think a cleaner way to do this would be to declare `app` as `mut` and then just
```
if cfg!(any(unix, windows)) {
  app = app.arg(…);
}

app
```

sharkdp:
Thank you very much for the updates!

PR #512, opened 2019-11-25
Remove OpenBSD install instructions

evitalis:
EPBSD was taken down awhile back due to inactivity. Remove the instructions for its use from the readme.


sharkdp:
Thank you

Issue #441, opened 2019-05-13
Should -X behavior match xargs?


kbd:
As discussed in [this `bat` issue](https://github.com/sharkdp/bat/issues/563), the behavior of `fd`'s `-X` flag is different from the behavior of `fd | xargs`. My initial impression upon reading the docs was that `-X`'s behavior would be equivalent to piping to `xargs`, and I was surprised when that wasn't the case.

Shell session for reference:

```shell
$ for i in a b c d; do echo $i > $i.txt; done
$ fd
a.txt
b.txt
c.txt
d.txt
$ fd -X bat -p
c
b
a
d
$ fd | xargs bat -p
a
b
c
d
$ fd --version
fd 7.3.0
```

In short, I would expect `-X`'s order to be the order returned by a normal call to `fd` (while I would expect `-x`'s order to be random because the docs make clear that operation is parallelized).

sharkdp:
Thank you for reporting this.

I'm going to summarize from what I said in the linked ticket:

* `fd`'s output order is *not* deterministic (even if it often looks like it actually is, especially in small examples). This means that the execution order can not be "predicted" by first running `fd …` and then running `fd … | xargs …`.
* The same is true for `-x`/`--exec` and `-X`/`--exec-batch`. `fd` doesn't make any promises on the order of the search results.

That being said, I don't really know why the order of `fd -X echo` seems to be reversed from what `fd` outputs. If someone want's to look into this, :+1:.

tavianator:
I believe this can be explained by the buffering logic, specifically https://github.com/sharkdp/fd/blob/5cbd8405ec14e4fc1c2e90836bf398908e5b8145/src/walk.rs#L231

sharkdp:
Oh, of course. I completely forgot about this.

If the search finishes within the buffering time (100 ms, by default), we actually sort the search results before printing them to the console.

For `fd -X`, we currently don't do this. I think it might be reasonable to do something similar because we are collecting all results anyway. Given that we are currently limited by the amount of search results anyway (for `-X`), the time to sort all results is probably negligible. If we would not be restricted, it might make sense to only sort the results if there are less than some fixed number.

sharkdp:
Closed via #524 by @MarcoIeni and @crash-g.

sharkdp:
Released in [fd 7.5.0](https://github.com/sharkdp/fd/releases/tag/v7.5.0)

PR #513, opened 2019-11-30
Sort search results when using -X option if results are not many. Solve Issue #441

MarcoIeni:
[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

This PR solves #441.

As suggested by @sharkdp in [this](https://github.com/sharkdp/fd/issues/441#issuecomment-499675495) comment we created a constant called SORT_THRESHOLD. If the found paths are < SORT_THRESHOLD, then the paths are sorted.

Feel free to propose changes if you have ideas to clean the code or to set a proper value for the SORT_THRESHOLD constant.

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -182,13 +182,13 @@ impl CommandTemplate {
                 // A single `Tokens` is expected
                 // So we can directly consume the iterator once and for all
 
-                // It would be too expansive to collect all the paths at once, so we take
+                // It would be too expensive to collect all the paths at once, so we take
                 // only the first ones.
                 let mut first_paths = take(&mut paths, SORT_THRESHOLD);
                 has_path = !first_paths.is_empty();
 
                 if first_paths.len() < SORT_THRESHOLD {
-                    // sort the paths because they are not too much
+                    // sort the paths because they are not too many
                     first_paths.sort();
                     CommandTemplate::add_to_cmd(&mut cmd, arg, &mut first_paths.into_iter());
                 } else {
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -182,13 +182,13 @@ impl CommandTemplate {
                 // A single `Tokens` is expected
                 // So we can directly consume the iterator once and for all
 
-                // It would be too expansive to collect all the paths at once, so I take
+                // It would be too expansive to collect all the paths at once, so we take
                 // only the first ones.
                 let mut first_paths = take(&mut paths, SORT_THRESHOLD);
                 has_path = !first_paths.is_empty();
 
                 if first_paths.len() < SORT_THRESHOLD {
-                    // I sort the paths because they are not too much
+                    // sort the paths because they are not too much
                     first_paths.sort();
                     CommandTemplate::add_to_cmd(&mut cmd, arg, &mut first_paths.into_iter());
                 } else {
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -27,6 +27,10 @@ use self::input::{basename, dirname, remove_extension};
 pub use self::job::{batch, job};
 use self::token::Token;
 
+/// When `ExecutionMode` is `Batch` and number of results is less
+/// than this value, results will be sorted
+const SORT_THRESHOLD: usize = 100;
+
 /// Execution mode of the command
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum ExecutionMode {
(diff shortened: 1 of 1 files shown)


src/exec/mod.rs
```diff
@@ -27,6 +27,10 @@ use self::input::{basename, dirname, remove_extension};
 pub use self::job::{batch, job};
 use self::token::Token;
 
+/// When `ExecutionMode` is `Batch` and number of results is less
+/// than this value, results will be sorted
+const SORT_THRESHOLD: usize = 100;
```
sharkdp: This would be similar to `MAX_BUFFER_LENGTH` from `src/internal.mod`. Could we reuse that constant?

src/exec/mod.rs
```diff
@@ -176,10 +180,22 @@ impl CommandTemplate {
         for arg in &self.args[1..] {
             if arg.has_tokens() {
                 // A single `Tokens` is expected
-                // So we can directy consume the iterator once and for all
-                for path in &mut paths {
-                    cmd.arg(arg.generate(&path).as_ref());
-                    has_path = true;
+                // So we can directly consume the iterator once and for all
+
+                // It would be too expansive to collect all the paths at once, so we take
```
sharkdp: typo: expensive

src/exec/mod.rs
```diff
@@ -176,10 +180,22 @@ impl CommandTemplate {
...
+                // only the first ones.
+                let mut first_paths = take(&mut paths, SORT_THRESHOLD);
+                has_path = !first_paths.is_empty();
+
+                if first_paths.len() < SORT_THRESHOLD {
+                    // sort the paths because they are not too much
```
sharkdp: not too many

src/exec/mod.rs
```diff
@@ -176,10 +180,22 @@ impl CommandTemplate {
...
-                    cmd.arg(arg.generate(&path).as_ref());
-                    has_path = true;
+                // So we can directly consume the iterator once and for all
+
+                // It would be too expansive to collect all the paths at once, so we take
+                // only the first ones.
+                let mut first_paths = take(&mut paths, SORT_THRESHOLD);
+                has_path = !first_paths.is_empty();
+
+                if first_paths.len() < SORT_THRESHOLD {
```
sharkdp: shouldn't this be `<= SORT_THRESHOLD`?

src/exec/mod.rs
```diff
@@ -192,6 +208,33 @@ impl CommandTemplate {
...
+            I: Iterator<Item = String>,
+    {
+        for path in paths {
+            cmd.arg(arg.generate(&path).as_ref());
+        }
+    }
+}
+
+/// This function behaves exactly like the standard library `take` for iterators,
+/// except that the original iterator is not consumed in the process.
```
sharkdp: Isn't that what [`Iterator::peekable()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.peekable) is for?

src/exec/mod.rs
crash-g: In principle I'm slightly in favour of a specific constant, possibly with the same value. Anyway, you have a better picture of the whole code, so if you think that the semantic is similar enough then I'm OK with reusing `MAX_BUFFER_LENGTH`.

src/exec/mod.rs
crash-g: Actually, as it is now `first_paths.len() <= SORT_THRESHOLD` is always true. Maybe it's clearer if we refactor as
```rust
let mut first_paths = take(&mut paths, SORT_THRESHOLD + 1);

if first_paths.len() <= SORT_THRESHOLD {
```

src/exec/mod.rs
crash-g: As far as I know there is no equivalent of this function in the standard library: could you elaborate on how you wanted to use `peek`?

crash-g:
> Thank you very much for your contribution!

Thanks to you for the careful review @sharkdp

src/exec/mod.rs
sharkdp: I'm not sure I understand what the intention is?

sharkdp: I appreciate that you were concerned about performance and introduced the `SORT_THRESHOLD`, but I am not sure if this is really required. I would expect the sort time to be much smaller than the actual filesystem-search. I would therefore remove the limit and simplify the logic here. There is a limit to the number of paths for `-X` anyway (see #410), so I don't think this will cause any problems.

MarcoIeni: Our motivating reason was [this](https://github.com/sharkdp/fd/issues/441#issuecomment-499675495). If sorting time is not an issue, please close this pull request and we'll submit another with less noise.

sharkdp: Ah, sorry. I forgot about my *"If we would not be restricted, it might make sense to only sort the results if there are less than some fixed number."* comment.

I would like to start with an easy solution where we always sort the paths. If this turns out to be a problem (performance wise), we can go back to an approach like this.

PR #517, opened 2019-12-07
fd-485: Clarified usage of ignore-files

### LINKED ISSUE #485: Is a ~/.ignore (or ~/.fdignore) file always looked at?
The README shows the following example:
```
> cat ~/.fdignore
/mnt/external-drive
*.bak
Note: fd also supports .ignore files that are used by other programs such as rg or ag.
```
On macOS I tried creating ~/.ignore and also a ~/.fdignore symlinked to ~/.ignore with the following in it as I'm trying to avoid mounted/network file systems when using `fd` from `fzf`:
```
.git
/Volumes
/net
/keybase
```
The expectation based on the README example if I just `cd / && fd -t d keybase` is to not return `/keybase` or `/Volumes/Keybase` but they are returned which makes me think that ~/.ignore or ~/.fdignore are not looked at.
If I'm explicit about using the ignore file `fd --ignore-file ~/.ignore -t d keybase` then it seems to work.

Not sure if this is a bug or expected behavior. It is very hard to tell as there's no documentation that lists the expected behavior, it's all very confusing.

I understand there have been a lot of discussions around ignore, it may be worth having a table of *expected* behavior then people can work from there when reporting issues.

kontrafiktion:
Attempts to fix #485

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,10 @@ this behavior, we can use the `-I` (or `--no-ignore`) option:
 target/debug/deps/libnum_cpus-f5ce7ef99006aa05.rlib
 ```
 
+*fd* respects the global git-ignore file as well. But as said above: Only when the directory belongs to a git repository.
+
+*fd* reads and uses ignore-files (`.ignore` and `.fdignore`), for the directory containing the ignore-file and its sub-directories. But a global `~/.ignore-file` or `~/.fdignore` are not used.
+
 To really search *all* files and directories, simply combine the hidden and ignore features to show
 everything (`-HI`).
 


README.md
```diff
@@ -417,6 +417,10 @@ this behavior, we can use the `-I` (or `--no-ignore`) option:
 target/debug/deps/libnum_cpus-f5ce7ef99006aa05.rlib
 ```
 
+*fd* respects the global git-ignore file as well. But as said above: Only when the directory belongs to a git repository.
+
+*fd* reads and uses ignore-files (`.ignore` and `.fdignore`), for the directory containing the ignore-file and its sub-directories. But a global `~/.ignore-file` or `~/.fdignore` are not used.
+
```
bits01:
```suggestion
But global `~/.ignore` or `~/.fdignore` are not used outside of `~`.
```

kontrafiktion: Really? I try to avoid 'negations' like that. Calling them "global" and then saying they are only used in "~" would confuse me. 

Would it be better to say:
```
.ignore and .fdignore files are only considered in the directory they are in (and the subdirectories). That means: .ignore and .fdignore in the HOME directory are not considered globally, but only for the HOME directory and its subdirectories
```

README.md
```diff
@@ -417,6 +417,10 @@ this behavior, we can use the `-I` (or `--no-ignore`) option:
 target/debug/deps/libnum_cpus-f5ce7ef99006aa05.rlib
 ```
 
+*fd* respects the global git-ignore file as well. But as said above: Only when the directory belongs to a git repository.
```
sharkdp: maybe add a link to the Git documentation ("global git-ignore file").

Wording: "*fd* also respects global git-ignore files when searching inside a Git repository."

README.md
```diff
@@ -417,6 +417,10 @@ this behavior, we can use the `-I` (or `--no-ignore`) option:
...
+
+*fd* reads and uses ignore-files (`.ignore` and `.fdignore`), for the directory containing the ignore-file and its sub-directories. But a global `~/.ignore-file` or `~/.fdignore` are not used.
```
sharkdp: If we really need this, I guess it should be moved to the "Excluding specific files or directories" section?

sharkdp: I would remove the "But a global `~/.ignore-file` or `~/.fdignore` are not used." part.

sharkdp:
Going to close this for now.

PR #524, opened 2020-01-02
Sort search results when using -X option

MarcoIeni:
This PR solves #441.

After reading [this](https://github.com/sharkdp/fd/issues/441#issuecomment-499675495) comment, me and [crash-g](https://github.com/crash-g) submitted pull request #513.
However, the solution was a little bit difficult to read, so @sharkdp said that he _"would like to start with an easy solution where we always sort the paths"_ because _"There is a limit to the number of paths for -X anyway (see #410)"_.

Therefore, since sorting time is not an issue, we preferred to create another pull request without all the noise of the old commits, where the paths are always sorted.

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -170,13 +170,15 @@ impl CommandTemplate {
         cmd.stdout(Stdio::inherit());
         cmd.stderr(Stdio::inherit());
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much for your contribution and the detailed documentation!

PR #554, opened 2020-04-01
Add link for Korean

spearkkk:
Translate into Korean

Hello, this is spearkkk from South korea.

I saw your nice command line util and I read docs to use `fd`.
While in reading, I think that this nice tool should be read in Korean too.
So, I just translate some part of README.md

I want to refer this translated docs into your repo.
Could you check this PR?

Let me know if you have any concern about this.
I will check again.


sharkdp:
Hi!

Thank you so much for your contribution!

PR #556, opened 2020-04-02
Add `-l`/`--list-details` option

### LINKED ISSUE #159: deterministic order - by default or by option flag
I'm pretty sure I get a different ordering of results on occasion when I use the app in the regular manner versus when I pipe it to some other app. (I had a concrete example of this but lost it and now can't find.) This would be the case perhaps if you collected results in parallel but did not force a deterministic output except when gathering it to deliver to a pipe (or vice versa). 

If this is the case, it's bad for me since sometimes I pipe the results to other commands, for example, `| line 2` to extract the second result, but now the desired item is in a different position.

Please add an option to lock it to deterministic output or enable determinism by default.

If I'm mistaken, please close.

### LINKED ISSUE #196: Recommended way to sort results
Before with `find` i could use the `printf` option https://unix.stackexchange.com/a/29901/196971

what the best way to solve this with fd ?

### LINKED ISSUE #324: Add option for natural ordering (a la `-v` in `ls`)
lt would be really, really handy for scripting to have a flag to arrange output using natural ordering.

Given the following `tree`:

```.
├── file-1
├── file-11
├── file-12
├── file-2
├── file-22
└── file-3

0 directories, 6 files
```

The output of `ls -1`/`fd` is currently:

```
file-1
file-11
file-12
file-2
file-22
file-3
```

The proposed output for natural order (`ls -1v`, proposed to be something like `fd -v`) would be:

```
file-1
file-2
file-3
file-11
file-12
file-22
```

If you need a dependency for this, [`rust-natord`](https://github.com/lifthrasiir/rust-natord) is small and seems like it could fit the bill.

### LINKED ISSUE #405: How to make the output of `fd` print only once for each file it finded?
E.g. for files that respect [pattern] in current directory, `$ fd [pattern] . . .` will print each of them 3 times. How to make it print only once?

### LINKED ISSUE #482: Show symlink target?
Thanks for making `fd`. I love it and I've been using it heavily for about a year.

I was wondering if it was possible to display the targets of symlinks for commands like:

```
fd --type l
```

I couldn't find anything in the documentation about this, but it would be a nice addition.

### LINKED ISSUE #491: Add a -l flag for listing more details about the files.
fd is great for listing the files that match some pattern, but if I want to know more details about those files (eg permissions, size, owner, modification date) I need to run them through `ls` or `exa`. `fd pattern -x ls -l` isn't such a great experience though: it isn't colored, the order is random, and the columns don't align if the file sizes aren't similar. Given that `-l` is the most common flag to `ls`, why not also support it in `fd`?

sharkdp:
Add a new `-l`/`--list-details` option to show more details about the search results. This is basically
an alias for `--exec-batch ls -l` with some additional `ls` options.
This can be used in order to:
* see metadata like permissions, owner, file size, modification times (#491)
* see symlink targets (#482)
* achieve a deterministic output order (#324, #196, #159)
* avoid duplicate search results when multiple search paths are given (#405)

image

It calls `ls` in this way:

``` rust
let color = matches.value_of("color").unwrap_or("auto");
let color_arg = ["--color=", color].concat();

Some(
    CommandTemplate::new_batch(&[
        "ls",
        "-l",               // long listing format
        "--human-readable", // human readable file sizes
        "--directory",      // list directories themselves, not their contents
        &color_arg,         // enable colorized output, if enabled
    ])
    .unwrap(),
)
```

closes #491 
closes #482
closes #324
closes #196
closes #159
closes #405

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 ## Features
 
-- Added `--max-results=<count>` option to limit the number of search results, see #472 and #476
+- Add a new `-l`/`--list` option to show more details about the search results. This is basically
+  an alias for `--exec-batch ls -l` with some additional `ls` options.
+  This can be used in order to:
+    * see metadata like permissions, owner, file size, modification times (#491)
+    * see symlink targets (#482)
+    * achieve a deterministic output order (#324, #196, #159)
+    * avoid duplicate search results when multiple search paths are given (#405)
+- Add a new `--max-results=<count>` option to limit the number of search results, see #472 and #476
   This can be useful to speed up searches in cases where you know that there are only N results.
   Using this option is also (slightly) faster than piping to `head -n <count>` where `fd` can only
   exit when it finds the search results `<count> + 1`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Features
 
-- Add a new `-l`/`--list` option to show more details about the search results. This is basically
-  an alias for `--exec-batch ls -l` with some additional `ls` options.
+- Add a new `-l`/`--list-details` option to show more details about the search results. This is
+  basically an alias for `--exec-batch ls -l` with some additional `ls` options.
   This can be used in order to:
     * see metadata like permissions, owner, file size, modification times (#491)
     * see symlink targets (#482)
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -66,6 +66,12 @@ Treat the pattern as a literal string instead of a regular expression.
 .B \-a, \-\-absolute\-path
 Shows the full path starting from the root as opposed to relative paths.
 .TP
+.B \-l, \-\-list
+Use a detailed listing format like 'ls -l'. This is basically an alias
+for '--exec-batch ls -l' with some additional 'ls' options. This can be used
+to see more metadata, to show symlink targets, to achieve a deterministic
+sort order and to avoid duplicate search results when using multiple search paths.
+.TP
 .B \-L, \-\-follow
 By default, fd does not descend into symlinked directories. Using this flag, symbolic links are
 also traversed.
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -66,7 +66,7 @@ Treat the pattern as a literal string instead of a regular expression.
 .B \-a, \-\-absolute\-path
 Shows the full path starting from the root as opposed to relative paths.
 .TP
-.B \-l, \-\-list
+.B \-l, \-\-list\-details
 Use a detailed listing format like 'ls -l'. This is basically an alias
 for '--exec-batch ls -l' with some additional 'ls' options. This can be used
 to see more metadata, to show symlink targets, to achieve a deterministic
--- a/src/app.rs
+++ b/src/app.rs
@@ -109,6 +109,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .overrides_with("absolute-path"),
         )
         .arg(
+            arg("list")
+                .long("list")
+                .short("l")
+                .conflicts_with("absolute-path"),
+        )
+        .arg(
             arg("follow")
                 .long("follow")
                 .short("L")
@@ -125,7 +131,8 @@ pub fn build_app() -> App<'static, 'static> {
             arg("null_separator")
                 .long("print0")
                 .short("0")
-                .overrides_with("print0"),
+                .overrides_with("print0")
+                .conflicts_with("list"),
         )
         .arg(arg("depth").long("max-depth").short("d").takes_value(true))
         // support --maxdepth as well, for compatibility with rg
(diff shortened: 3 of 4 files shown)


sharkdp:
Thinking about it, the `ls` option `-l` is probably an abbreviation for "long", not "list". Any suggestions on how to name the `--long` option? `--list`, `--long`, `--list-details`?

ErichDonGubler:
Wait...what about us poor Windows users with no `ls` command?

sharkdp:
> Wait...what about us poor Windows users with no `ls` command?

I honestly completely forgot about you, sorry :sweat_smile:.

Any suggestions? (I guess everything would work fine in a `Cygwin`-style environment where some GNU utils are available?)

itchyny:
Note that `--human-readable` and `--directory` options are not available on macOS (`-h` is ok).

sharkdp:
> Note that `--human-readable` and `--directory` options are not available on macOS (`-h` is ok).

Thank you for the feedback. That should be fixed before this feature is released.

Hrxn:
> > Wait...what about us poor Windows users with no `ls` command?
> 
> I honestly completely forgot about you, sorry 😅.
> 
> Any suggestions? (I guess everything would work fine in a `Cygwin`-style environment where some GNU utils are available?)

Good question. However, I think this should work without using any crutches like Cygwin.
GNU `ls` has been ported to Windows before, after all. The issue is closed, and this has been addressed by simply trying to run a `ls.exe` on Windows, it seems.

Just FYI, I'll leave this here just in case someone else stumbles upon this:

Because the traditional ports of basic GNU utils on Windows (http://gnuwin32.sourceforge.net/, http://unxutils.sourceforge.net/) are horribly outdated, I would recommend trying alternatives such as

- msls: Adaption for Windows, using WinAPI. This has some benefits for Windows user, and it should still be compatible, although I haven't tested this for myself yet. 
https://github.com/u-tools/msls

- uutils/coreutils: A rewrite of the GNU coreutils in Rust to enable proper cross-platform support.
https://github.com/uutils/coreutils

sharkdp:
> Good question. However, I think this should work without using any crutches like Cygwin.
> GNU `ls` has been ported to Windows before, after all. The issue is closed, and this has been addressed by simply trying to run a `ls.exe` on Windows, it seems.

I tried this and it worked in both cygwin as well as PowerShell/cmd on Windows 10. I'm not a Windows user, so I'm not sure if a GNU version of `ls` is really standard (probably not). I might have installed a version with cygwin or some other tool.

In the release notes (https://github.com/sharkdp/fd/releases/tag/v8.0.0), I suggested that package maintainers maybe add an optional dependency to a GNU-compatible `ls`.

That being said, I'm open to allowing other alternatives as well. For macOS, we try both `gls` and the macOS-native `ls`.

Hrxn:
> I tried this and it worked in both cygwin as well as PowerShell/cmd on Windows 10. I'm not a Windows user, so I'm not sure if a GNU version of `ls` is really standard (probably not). I might have installed a version with cygwin or some other tool.

Yeah, must be something like this, because out of the box there is definitely no `ls` on Windows. (Well, it works with WSL (Windows Subsystem for Linux), but one has to install that manually first)

 
> In the release notes (https://github.com/sharkdp/fd/releases/tag/v8.0.0), I suggested that package maintainers maybe add an optional dependency to a GNU-compatible `ls`.
>

Yes, absolutely right.
 
> I also tried to implement a fallback to Windows-native `dir`, but it is simply too limited. I could not pass multiple files, and there is no equivalent of `--directory` (show the directory entry itself, _not_ the contents) which is really essential for this new `--list-details` option.
> 

Oh, please don't bother wasting time with the old `dir`, definitely not worth it 😄

ErichDonGubler:
Sorry, I've been busy job hunting and haven't felt like I could drop in and give my two cents until now. I'll do so now. :)

So, I'm interested: there doesn't seem to me to be a huge tail of complexity with displaying file information, though I'm sure I'm missing something. So what AM I missing? What were the tradeoffs considered an alternative like rolling `ls`-alike input into `fd`?

tmccombs:
I think there actually is quite a bit of complexity in implementing it across multiple operating systems.  In particular, permissions are completely different on windows than they are in UNIX-like oses. symlinks work differently, etc.

sharkdp:
> I think there actually is quite a bit of complexity in implementing it

I also think so. I'm sure there are a lot of edge cases which would need to be handled correctly. And would have potential for a lot of future bug reports. The text formatting is non trivial, there is date handling, locale support, etc.

That being said, calling external programs is also something that comes with some problems. Especially when we have to support all kinds of different operating systems. We already had our first bug report (and fix: #573). Also, we have to wait for all search results before we can call `ls`. Implementing it within `fd` would allow streaming of results. So I'm not (anymore) completely opposed to implementing the required functionality within `fd`.

In any case, I am really happy about that new feature. I was originally not convinced by the idea but now that we have implemented it, I absolutely love it. Thanks to @tewalds for the great idea :+1:

tewalds:
Thanks for implementing it! 

I'm guessing using [exa](https://github.com/ogham/exa) as a library might make this easier to implement.

Issue #561, opened 2020-04-08
Add -1 option as an alias for --max-results=1


sharkdp:
see #476

tpiekarski:
Did anyone started already? Would like to offer a PR and implement that tomorrow.

SimplyDanny:
I did. See #563.

tpiekarski:
Thanks, so guess I move along and search for other good first issues to get into Rust. :disappointed:

sharkdp:
closed in #563 by @SimplyDanny.

@tpiekarski There are a few more "good first issues" in `fd`s issue tracker. Some of these tickets are [staged for the upcoming v8.0 release](https://github.com/sharkdp/fd/issues?q=is%3Aopen+is%3Aissue+milestone%3Av8.0+label%3A%22good+first+issue%22).

Ordered by increasing complexity:
* #469 (in v8.0 milestone)
* #562  (in v8.0 milestone)
* #485 
* #511 (in v8.0 milestone)

tpiekarski:
@sharkdp Thanks, I'll take a closer look at those issues 😊

sharkdp:
This has now been released in [fd v8.0](https://github.com/sharkdp/fd/releases/tag/v8.0.0).

PR #563, opened 2020-04-09
Add short form -1 for --max-results=1

SimplyDanny:
This is an implementation for #561. Please have in mind that these are some of my first steps with Rust. :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -416,6 +416,15 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help("Limit the number of search results to 'count' and quit immediately."),
         )
         .arg(
+            Arg::with_name("max-one-result")
+                .short("1")
+                .hidden_short_help(true)
+                .overrides_with("max-results")
+                .conflicts_with_all(&["exec", "exec-batch"])
+                .long_help("Limit the search to a single result and quit immediately. \
+                                This is an alias for '--max-results=1'.")
+        )
+        .arg(
             Arg::with_name("show-errors")
                 .long("show-errors")
                 .hidden_short_help(true)
--- a/src/main.rs
+++ b/src/main.rs
@@ -295,7 +295,8 @@ fn run() -> Result<ExitCode> {
         max_results: matches
             .value_of("max-results")
             .and_then(|n| usize::from_str_radix(n, 10).ok())
-            .filter(|&n| n != 0),
+            .filter(|&n| n != 0)
+            .or_else(|| if matches.is_present("max-one-result") { Some(1) } else { None }),
     };
 
     let re = RegexBuilder::new(&pattern_regex)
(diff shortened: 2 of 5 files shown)


tests/tests.rs
```diff
@@ -1458,7 +1458,7 @@ fn test_max_results() {
     );
 
     // Limited to one result. We could find either C.Foo2 or c.foo
-    let output = te.assert_success_and_get_output(".", &["--max-results=1", "c.foo"]);
```
sharkdp, line 1461: I would like to keep this test written with `--max-results=1`. If we want to add a test for the new option (which definitely doesn't hurt), I suggest we move these few lines into a closure or new function and call it twice, with `--max-results=1` and `-1`.

src/main.rs
```diff
@@ -295,7 +295,8 @@ fn run() -> Result<ExitCode> {
         max_results: matches
             .value_of("max-results")
             .and_then(|n| usize::from_str_radix(n, 10).ok())
-            .filter(|&n| n != 0),
+            .filter(|&n| n != 0)
+            .or_else(|| if matches.is_present("max-one-result") { Some(1) } else { None }),
```
sharkdp, line 299: :+1: 

Just a side note: this type of pattern
``` rust
if boolean_expression { Some(value) } else { None }
```
can possibly be written as `boolean_expression.then_some(value)` in the future: https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

This line would then simplify to:
``` rust
.or_else(|| if matches.is_present("max-one-result").then_some(1)), 
```

src/app.rs
```diff
@@ -415,6 +415,14 @@ pub fn build_app() -> App<'static, 'static> {
                 .hidden_short_help(true)
                 .long_help("Limit the number of search results to 'count' and quit immediately."),
         )
+        .arg(
+            Arg::with_name("max-one-result")
+                .short("1")
+                .hidden_short_help(true)
+                .overrides_with("max-results")
+                .conflicts_with_all(&["exec", "exec-batch"])
+                .long_help("Short form of '--max-results=1'.")
```
sharkdp, line 424: Even if it is just the short form for `--max-results=1`, I think it would be good to repeat what it does. Also, we use the word "alias" throughout the `--help` text. So maybe:
```rust
.long_help("Limit the search to a single result. This is an alias for '--max-results=1'.")
```
What do you think?

(same for the man page entry)

SimplyDanny: I agree. Good point.

src/main.rs
SimplyDanny: Yes, I was looking for something like this and found the method you mentioned, just to realize that it is currently experimental. Too bad.

tests/tests.rs
SimplyDanny: I have done so with a closure.

SimplyDanny:
> This looks good. I have added a few minor comments.

 Thank you very much for your thorough review!

> Could you please also update the `CHANGELOG.md` and add an appropriate entry below the line introducing the `--max-results` feature?

Done.

Issue #523, opened 2019-12-31
RFE: Add --perm flag


kclevenger:
In conjunction with #328 (specifically the --owner option), please consider adding a --perm flag accepting the standard four digit permission format as a filter

sharkdp:
Thank you for your feedback.

I always a little hesitant to add new options to `fd`. This one in particular, because it would be OS specific.

> accepting the standard four digit permission format

What do you mean by "standard for digit … format"? I am familar with the user/group/other three digit format (e.g. 644), but what would the fourth digit be?

Apart from that, I have several questions:

- Most importantly: What are actual real world use cases where this option would be useful? If `--perm 644` simply searches for files with permissions set to `644` *exactly*, I don't really see much use for this. I'd rather expect queries like *"files/directories which are writable/readable by me"* or *"files/directories which are NOT writable/readable by me"* to be useful.
- `find` seems to have invented a whole new syntax for permission filters (`-perm mode`, `-perm -mode`, `-perm /mode`), which goes a long with the `find`-specific logic operations for negation, conjunction, disjunction, etc. I am not planning to implement such a complex logic system within `fd`. Is this option really going to be useful without all of this?
- How are we going to deal with other operating systems? I'd really like to avoid having a completely UNIX-centric command line option like `--perm UGO`. `find` has several more options dealing with permissions (`-readable`, `-writable`), which sound a bit more appealing to me. Those could probably also be implemented in a OS-independent way(?).

tmccombs:
> but what would the foruth digit be?

The fourth digit is actually the first one, which is usually omitted or 0.  It is used for the setuid, setgid, and restricted-deletion/sticky flags.  The chmod man page has more information on them.

FranklinYu:
Might be related to #400.

sharkdp:
None of the questions above have been answered after two years. I'd like to close this for now.

colindean:
I just encountered a need for this. I needed to find all subdirectories in a tree that have 777 permissions.

```
fd \
  --full-path /the/root/dir \
  --max-depth 1 \
  --changed-after 2023-02-19 \
  --exec find {}/ -type d -perm 0777
```

I was surprised to discover that `fd` didn't have this baked in already and that I needed to effectively filter through ye olde `find` to accomplish it. Something in `fd` would scratch the itch, but an arbitrary filtration command provided by #400 would suffice, too.

Kernald:
[...]

PR #565, opened 2020-04-11
Adding --perm option to filter by Unix 3 digits numeric permission notation

eduriol:
- Added --perm option for Unix systems accepting a 3 digit number to filter by Unix permission system. Example: 'fd --perm 755' will yield only results with 755 permission level enabled.
- I took mainly the implementation of the --size option as inspiration for the implementation of this one.
- Added unitary and integration tests to cover this new option.


sharkdp:

Maybe we should first discuss this in a more general sense in the corresponding feature request ticket: #523. I will comment there.

eduriol:
I'm trying to figure out how to overcome the problem of this solution not compiling in Windows platforms.

eduriol:
Ok, I'm having a hard time making it compiling in Windows platform to pass the CI tests, but you are right, let's discuss it at high level first.

sharkdp:
closing for now (cleaning up list of open PRs).

PR #571, opened 2020-04-15
MacOS and Windows support for --list-details/-l

### LINKED ISSUE #562: Make sure --list-details works on Windows and MacOS
- [x] MacOS (use `gls`, if available. otherwise, use `ls`)
- [x] Windows

see comments in #556

sharkdp:
closes #562

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -176,10 +176,15 @@ fn run() -> Result<ExitCode> {
                 ]
             } else {
                 // MacOS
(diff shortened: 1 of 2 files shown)


sharkdp:
@itchyny This should fix `-l`/`--list-details` for MacOS users.

@ErichDonGubler I have also implemented a check for `ls` on Windows. If it exists, everything should work fine. If it does not, we print a corresponding error.

PR #573, opened 2020-04-16
Fix --list-details on FreeBSD

t6:


sharkdp:
Thank you very much for fixing this!

Ah, I was afraid that it might not work on BSD-variants, but I'm really not familar with those (and have never had the time to set up a virtual machine).

I'm not sure how similar FreeBSD is to OpenBSD and NetBSD... would it make sense to add those as well?

src/main.rs
```diff
@@ -194,9 +194,9 @@ fn run() -> Result<ExitCode> {
                     gnu_ls("gls")
                 } else {
                     let mut cmd = vec![
-                        "ls", // MacOS version of ls
+                        "ls", // BSD version of ls
```
sharkdp, line 197: :+1:

t6:
Not that familiar with OpenBSD and NetBSD and I cannot test on either of them, but judging by the man pages both do not have `-G`. I suppose we can exclude that argument there.

https://netbsd.gw.com/cgi-bin/man-cgi?ls++NetBSD-current
https://man.openbsd.org/ls

t6:
Added NetBSD and OpenBSD support too

sharkdp:
Perfect - thank you very much!

sharkdp:
I see that you maintain the FreeBSD package for `fd`? (Thank you!)

Let me know if it would help you if I release a new version including this bugfix.

t6:
Thanks. I've already updated the package with the two commits applied. It's not really needed from that perspective and can wait until you'd put out a new release naturally anyway. OTOH a new release might help people who don't use the package and just do `cargo install fd-find` on FreeBSD.

sharkdp:
Ok, thank you.

In that case, I'll probably wait for a few more days and see if other things need fixing as well.

sharkdp:
Released in [fd 8.1.0](https://github.com/sharkdp/fd/releases/tag/v8.1.0)

PR #575, opened 2020-04-22
Add global fdignore support

soedirgo:
Rust newbie here :-)

This implements global fdignore support in `~/.config/fd/ignore` as mentioned [here](https://github.com/sharkdp/fd/issues/485#issuecomment-532869261).

I have some questions:
- I'm using `std::env::home_dir()` (which is deprecated) instead of using the `dirs` crate, because this is what the `ignore` crate does. Which would you prefer?
- I'm not too sure how to handle the tests. If I have, say, `*` in my `~/.config/fd/ignore`, the tests will break. The simplest way I can think of to deal with this is to add a `!*` using `--ignore-file`.
- Are there parts of the doc I should change?

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::env;
 use std::ffi::OsStr;
 use std::fs::{FileType, Metadata};
 use std::io;
@@ -80,6 +81,34 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Options>) -> 
 
     if config.read_fdignore {
         walker.add_custom_ignore_filename(".fdignore");
+    }
+
+    if config.read_global_ignore {
+        #[cfg(target_os = "macos")]
+        let config_dir_op = env::var_os("XDG_CONFIG_HOME")
+            .map(PathBuf::from)
+            .filter(|p| p.is_absolute())
+            .or_else(|| dirs::home_dir().map(|d| d.join(".config")));
+
+        #[cfg(not(target_os = "macos"))]
+        let config_dir_op = dirs::config_dir();
+
+        if let Some(global_ignore_file) = config_dir_op
+            .map(|p| p.join("fd").join("ignore"))
+            .filter(|p| p.is_file())
+        {
+            let result = walker.add_ignore(global_ignore_file);
+            match result {
+                Some(ignore::Error::Partial(_)) => (),
+                Some(err) => {
+                    print_error(format!(
+                        "Malformed pattern in global ignore file. {}.",
+                        err.to_string()
+                    ));
+                }
+                None => (),
+            }
+        }
     }
 
     for ignore_file in &config.ignore_files {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ humantime = "2.0"
 lscolors = "0.7"
 globset = "0.4"
 anyhow = "1.0"
+dirs = "2.0"
 
 [dependencies.clap]
 version = "2.31.2"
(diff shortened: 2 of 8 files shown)


sharkdp:
> Rust newbie here :-)

Welcome!

> This implements global fdignore support in `~/.config/fd/ignore` as mentioned [here](https://github.com/sharkdp/fd/issues/485#issuecomment-532869261).

Great!



> I'm using `std::env::home_dir()` (which is deprecated) instead of using the `dirs` crate, because this is what the `ignore` crate does. Which would you prefer?

I prefer using the same solution [that we have for `bat`](https://github.com/sharkdp/bat/blob/abeac8b12e9e4c0bc6d9c03ce0d12751a81e9757/src/bin/bat/directories.rs#L6-L8): follow the [XDG standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), i.e. use `$XDG_CONFIG_HOME`, if available. Otherwise, fall back to OS-specific standard directories. On macOS, use `~/.config/fd` as well (not what `dirs` suggests). We could copy some code from `bat`. We don't need the `get_cache_dir` opart.

src/walk.rs
```diff
@@ -82,6 +83,27 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Options>) ->
         walker.add_custom_ignore_filename(".fdignore");
     }
 
+    // Global fdignore, has lowest precedence.
+    if let Some(global_ignore_file) = env::var_os("XDG_CONFIG_HOME")
+        .and_then(|x| if x.is_empty() { None } else { Some(PathBuf::from(x)) })
+        // std::env::home_dir() is deprecated. We should switch to the `dirs` crate in the future.
+        .or_else(|| env::home_dir().map(|p| p.join(".config")))
+        .map(|x| x.join("fd/ignore"))
```
sharkdp, line 91: I guess it works on Windows as well, but maybe just use `.join("fd").join("ignore")`.

soedirgo: Got it 👍.

soedirgo:
Thanks for the review!

I changed the global ignore file's behavior so that it gets ignored with `-I`. I think it should behave like `--ignore-file`s, but `-I` should ignore anything that isn't explicit (i.e. `--ignore-file`s).

sharkdp:
Thank you for the updates.

Two things:

First, could we please make the documentation less Unix-specific? (i.e. paths like `~/.config/fd/ignore`)

Second, I know this will be hard to add to integration tests. But could you please show a list of tests that you have performed locally to make sure that this works as intended? In particular, I'm thinking about
* Main functionality. Does it work with globs like `*.txt` and absolute paths like `/sys`?
* Prioritization of different ignore files
* What happens if invalid patterns are in the global ignore file?
* Interplay with `--no-ignore` and `--unrestricted`
* Interplay with `--no-ignore-vcs` (files from global ignore file should still be ignored)

soedirgo:
> First, could we please make the documentation less Unix-specific? (i.e. paths like ~/.config/fd/ignore)

Updated.

> Second, I know this will be hard to add to integration tests. But could you please show a list of tests that you have performed locally to make sure that this works as intended?

- Functionality is like `--ignore-file`s, i.e. it works with globs, but absolute paths are matched such that the root is the working directory.
- From what I can see, priority is `fd/ignore` < `--ignore-file`s < `.gitignore` < `.ignore` < `.fdignore`.
- Invalid patterns will print an error and ignore the pattern, but other patterns will still be matched.
- `--no-ignore` and `--unrestricted` ignores the global ignore file.
- `--no-ignore-vsc` has no effect.

sharkdp:
Released in [fd 8.1.0](https://github.com/sharkdp/fd/releases/tag/v8.1.0)

PR #579, opened 2020-04-26
Fix --glob behavior with empty pattern

SeamusConnor:
With an empty pattern, --glob results in no matches. This does not
align with the behavior of --fixed-strings or --regex. Update the
initialization code to handle globs with empty patterns.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -122,7 +122,7 @@ fn run() -> Result<ExitCode> {
         ));
     }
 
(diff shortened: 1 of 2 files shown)


sharkdp:

SeamusConnor:
> Thank you very much for your contribution!

Thank you very much for this nice tool :)

sharkdp:
Fixed in [fd 8.1.0](https://github.com/sharkdp/fd/releases/tag/v8.1.0)

PR #123, opened 2017-10-15
Fix path check (fixes #113)

iology:

PR #581, opened 2020-05-01
Add -o/--owner to filter files by owner and/or group

### LINKED ISSUE #307: Finding by owner username/group/uid/gid

alexmaco:
Closes #307

This implements a single constraint.
Implementation is unix-specific for now. ~~It can easily be extended to cover redox, but I think it needs some testing first: the `users` crate makes no mention of redox support, and there is also [redox_users](https://crates.io/crates/redox_users)~~

Possible questions:
- should this PR also add support for multiple `--owner` arguments (like with time_constraints)
- help text refinements

Edit: it seems redox uses the unix `target_family` as of https://github.com/rust-lang/rust/issues/60139, so no special handling is needed.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -529,7 +529,18 @@ pub fn build_app() -> App<'static, 'static> {
             Arg::with_name("owner")
                 .long("owner")
                 .short("o")
-                .takes_value(true),
+                .takes_value(true)
+                .value_name("user:group")
+                .help("Filter by owning user and/or group")
+                .long_help(
+                    "Filter files by their user and/or group. \
+                     Format: [(user|uid)][:(group|gid)]. Either side is optional. \
+                     Precede either side with a '!' to exclude files instead.\n\
+                     Examples:\n    \
+                         --owner john\n    \
+                         --owner :students\n    \
+                         --owner '!john:students'",
+                ),
         );
     }
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -524,6 +524,15 @@ pub fn build_app() -> App<'static, 'static> {
                 ),
         );
 
+    if cfg!(unix) {
+        app = app.arg(
+            Arg::with_name("owner")
+                .long("owner")
+                .short("o")
+                .takes_value(true),
+        );
+    }
+
     // Make `--one-file-system` available only on Unix and Windows platforms, as per the
     // restrictions on the corresponding option in the `ignore` crate.
     // Provide aliases `mount` and `xdev` for people coming from `find`.
--- a/src/filter/owner.rs
+++ b/src/filter/owner.rs
@@ -1,0 +1,61 @@
+use anyhow::{anyhow, Result};
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct OwnerFilter {
+    uid: Option<u32>,
+    gid: Option<u32>,
+}
+
+impl OwnerFilter {
+    pub fn from_string(input: &str) -> Result<Self> {
+        let mut it = input.split(':');
+        let (fst, snd) = (it.next(), it.next());
+
+        let uid = match fst {
+            Some("") | None => None,
+            Some(s) => Some(s.parse()?),
+        };
+        let gid = match snd {
+            Some("") | None => None,
+            Some(s) => Some(s.parse()?),
+        };
+
+        if uid.is_none() && gid.is_none() {
+            Err(anyhow!(
+                "'{}' is not a valid user/group specifier. See 'fd --help'.",
+                input
+            ))
+        } else {
+            Ok(OwnerFilter { uid, gid })
+        }
+    }
+}
+
+#[cfg(test)]
+mod owner_parsing {
+    use super::OwnerFilter;
+
+    macro_rules! owner_tests {
+        ($($name:ident: $value:expr => $result:pat,)*) => {
+            $(
+                #[test]
+                fn $name() {
+                    let o = OwnerFilter::from_string($value);
+                    match o {
+                        $result => {},
+                        _ => panic!("{:?} does not match {}", o, stringify!($result)),
+                    }
+                }
+            )*
+        };
+    }
+
+    owner_tests! {
+        empty:      ""      => Err(_),
+        uid_only:   "5"     => Ok(OwnerFilter { uid: Some(5), gid: None   }),
+        uid_gid:    "9:3"   => Ok(OwnerFilter { uid: Some(9), gid: Some(3)}),
+        gid_only:   ":8"    => Ok(OwnerFilter { uid: None,    gid: Some(8)}),
+        colon_only: ":"     => Err(_),
+        trailing:   "5:"    => Ok(OwnerFilter { uid: Some(5), gid: None   }),
+    }
+}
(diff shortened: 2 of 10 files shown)


src/filter/owner.rs
```diff
@@ -0,0 +1,113 @@
...
+enum Check<T> {
+    Equal(T),
+    NotEq(T),
+    Ignore,
+}
+
+impl OwnerFilter {
+    pub fn from_string(input: &str) -> Result<Self> {
+        let mut it = input.split(':');
+        let (fst, snd) = (it.next(), it.next());
```
sharkdp, line 20: Could we add a check here to make sure that `it.next()` is `None` after calling it twice? To disallow usage of two or more colons like `--owner shark:lxd:whatever` which would likely be a mistake.

src/filter/owner.rs
```diff
@@ -0,0 +1,113 @@
...
+                .ok_or_else(|| anyhow!("'{}' is not a recognized user name", s))
+        })?;
+        let gid = Check::parse(snd, |s| {
+            s.parse()
+                .ok()
+                .or_else(|| users::get_group_by_name(s).map(|group| group.gid()))
+                .ok_or_else(|| anyhow!("'{}' is not a recognized group name", s))
+        })?;
+
+        if let (Check::Ignore, Check::Ignore) = (uid, gid) {
```
sharkdp, line 35: This would only trigger if a user called `--owner :`, right? I'm just a bit irritated by the error message "is not a valid user/group specifier".

src/app.rs
```diff
@@ -524,6 +524,22 @@ pub fn build_app() -> App<'static, 'static> {
...
+            Arg::with_name("owner")
+                .long("owner")
+                .short("o")
+                .takes_value(true)
+                .value_name("user:group")
+                .help("Filter by owning user and/or group")
+                .long_help(
+                    "Filter files by their user and/or group. \
+                     Format: [(user|uid)][:(group|gid)]. Either side is optional. \
+                     Precede either side with a ! to exclude files instead.",
```
sharkdp, line 538: Minor: `…either side with a '!' to exclude…`

Specifying the full format is great, but a few examples would make things even easier to understand, I think. Maybe:

```
Examples:
  --owner john
  --owner :students
  --owner '!john:students'
```
```

sharkdp:
Thank you very much for this contribution! The code looks great.

> should this PR also add support for multiple `--owner` arguments (like with time_constraints)

It would be great, but we can also add this at a later stage. Multiple `--owner` arguments would act in an *OR*-like fashion, right? Similar to `--extension`. Maybe we should think about this in a bit more detail. Combining this with negation could lead to surprising results(?).

A few questions / comments:
- Did you do some research into other crates that would work in an OS-independent way? I am not familiar with how file ownership works on Windows, but they certainly have the concept of owners (https://docs.microsoft.com/en-us/windows/win32/secauthz/finding-the-owner-of-a-file-object-in-c--)
- I guess it would be hard to write actual integration tests for this?
- For some reason, the AppVeyor tests did not run for this PR and I can not manually trigger a build. Travis CI tests also didn't run for the last commit. Could you maybe rebase this PR to see if that triggers a build?
- Could you please add an entry to the "unreleased" section in the `CHANGELOG.md` under "Features"? The format for entries is:
  ```
  - Description what has been changed, see #123 (@user)
  ```
  where `#123` links to the bug ticket and/or this PR and `user` is your username.
- It would be great if you could also update the man page. Otherwise, I can also sync it with the `--help` text before the next release.

src/filter/owner.rs
alexmaco: Yes, seems like a good thing to check. I'll add it.

src/filter/owner.rs
alexmaco: Yes, that's the only case for which it should trigger.
Do you have any proposals for the error message ?
Maybe something like "At least least a user or group must be specified as owner" ?

sharkdp: Maybe we should simply allow it? That would remove some error handling code and I don't see this being used by accident. It might even be useful when someone calls `fd --owner …` from a script.

Hm.. I guess it would also trigger for `--owner ''`? Do we want to allow this as well?

alexmaco:
> Thank you very much for this contribution! The code looks great.
You're welcome!
> > should this PR also add support for multiple `--owner` arguments (like with time_constraints)
> 
> It would be great, but we can also add this at a later stage. Multiple `--owner` arguments would act in an _OR_-like fashion, right? Similar to `--extension`. Maybe we should think about this in a bit more detail. Combining this with negation could lead to surprising results(?).

Yes, it should work by OR-ing, and negations should work by AND-ing, which i think would make sense when seen in examples, but may not be intuitive to everyone.
I'm waiting for any clear ideas on this topic.

>     * Did you do some research into other crates that would work in an OS-independent way? I am not familiar with how file ownership works on Windows, but they certainly have the concept of owners (https://docs.microsoft.com/en-us/windows/win32/secauthz/finding-the-owner-of-a-file-object-in-c--)

Unfortunately no, I didn't dig very far, since I can only properly try it out on linux systems. Finding a cross-platform way to do this would be interesting however. I imagine that it could also account for more interesting setups, such as LDAP/ActiveDirectory/whatever-the-equivalent-on-OSX is. But I'm not familiar enough with windows to propose solutions that look idiomatic on the platform.

>     * I guess it would be hard to write actual integration tests for this?

Integration tests would certainly be interesting. I guess setting up a docker image with known users/groups and IDs would be sufficient. It looks like travis can speak docker.

>     * For some reason, the AppVeyor tests did not run for this PR and I can not manually trigger a build. Travis CI tests also didn't run for the last commit. Could you maybe rebase this PR to see if that triggers a build?

Sure

>     * Could you please add an entry to the "unreleased" section in the `CHANGELOG.md` under "Features"? 
>     ...
>     * It would be great if you could also update the man page. 

Ok, I did not realize more docs existed :smile:. I will update them.

src/filter/owner.rs
alexmaco: FOllowing the script usage scenario, it seems reasonable that when an empty string is passed, `--owner ''` should be a noop as well.

alexmaco: I ended up leaving this last condition in, but this time  to make `--owner ''` truly equivalent to not passing `--owner` at all

sharkdp:
> Integration tests would certainly be interesting. I guess setting up a docker image with known users/groups and IDs would be sufficient. It looks like travis can speak docker.

That sounds rather involved. Let's try without an integration test for now.

Issue #406, opened 2019-02-09
Support multiple --exec <cmd> instances


terlar:
Sometimes I run a command in several different matching directories with --exec, but I would like to print each directory for each command that is executed without having to add an intermediate shell just to print the path.

Perhaps adding an option that would print the matched file-name as well as executing the command in `--exec`. I tried with multiple `--exec` but it is not allowed, the only workaround I found was to have an intermediate shell.

Here is an example (print the git-status in all matching git directories:
```
fd -HFtd .git -x git -C '{//}' status -s
```

I can imagine that there are other use cases where you would want to print the matched file you execute the `-x` command for.

sharkdp:
Thank you very much for the feedback.

I definitely see how that could be useful, yes. Maybe supporting multiple `--exec`s would be a good idea in general. To support this use case here, users could simply use `-x echo -x whatever`.

asyncsrc:
I actually ran into this while testing `fd` today.  I love the utility btw!

As a workaround for right now, I noticed you can have `-x` call the bash interpreter where then you can use multiple `{}` placeholders like:

``` bash
$> fd -e rs -x bash -c "grep main {} && echo {}"

fn main() {
main.rs
```

tavianator:
@asyncsrc Note that that will break if you have filenames with shell punctuation like apostrophes.  This is safe though:
```
fd -e rs -x bash -c 'grep main "$1" && echo "$1"' bash {}
```

cup:
https://github.com/sharkdp/fd/issues/514

sasacocic:
Any progress on this? This would be really useful I think.

sharkdp:
I tried to look into this today, but I think we currently run into a limitation of `clap` (https://github.com/clap-rs/clap/issues/1026) which prevents us from using two instances of `--exec`/`-x`.

SeamusConnor:
I looked into clap with an eye on adding this feature, and was surprised by the overall size of the project. While I agree that it would be better if clap just did what you needed, there is enough info in the MatchedArg to rebuild multiple commands:

```
let exec_args = matches.args.get("exec").unwrap();
println!("{:?}", exec_args);

let mut last = exec_args.indices[0] - 1;
for it in exec_args.vals.iter().zip(exec_args.indices.iter()) {
    let (arg, idx) = it;
    if *idx == last + 1 {
        print!("{} ", arg.to_string_lossy());
    } else {
        print!(", {} ", arg.to_string_lossy());
    }
[...]

PR #582, opened 2020-05-03
Multi exec parsing prototype

SeamusConnor:
This is a prototype for handling command line parameters for multiple --exec options (#406). The actual exec'ing is not implemented. All of the code is implemented in main, but the iterator would be moved elsewhere as part of exiting prototype.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,6 +11,7 @@ mod regex_helper;
 mod walk;
 
 use std::env;
+use std::iter::Peekable;
 use std::path::{Path, PathBuf};
 use std::process;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -154,10 +202,20 @@ fn run() -> Result<ExitCode> {
...
-    } else if let Some(args) = matches.values_of("exec-batch") {
-        Some(CommandTemplate::new_batch(args)?)
+    let mut commands: Vec<CommandTemplate> = vec![];
+
+    if matches.is_present("exec") {
+        let ranges = ContinuousRangeIterator::new(matches.indices_of("exec").unwrap().peekable());
+        let args: Vec<&str> = matches.values_of("exec").unwrap().collect();
+        for (start, end) in ranges {
+            commands.push(CommandTemplate::new(&args[start..end]));
+        }
```
SeamusConnor, line 212: I wanted to call `.take(n)` in a loop on the value iterator, but calling take moves the iterator. Since that didn't work, I tried having the iterator return slices of some internal value array, which also made the borrow checker sad. I gave up and just  built a vec out of the values. If there is some way to have the Iterator own the values, and return references to those values, I would like to know it.

src/main.rs
```diff
@@ -154,10 +202,20 @@ fn run() -> Result<ExitCode> {
...
+    } else if matches.is_present("exec-batch") {
+        let ranges = ContinuousRangeIterator::new(matches.indices_of("exec-batch").unwrap().peekable());
+        let args: Vec<&str> = matches.values_of("exec-batch").unwrap().collect();
+        for (start, end) in ranges {
+            commands.push(CommandTemplate::new_batch(&args[start..end])?);
```
SeamusConnor, line 217: Its a shame to duplicate this code for both exec and exec batch. This is a result of my borrow checker woes. Any ideas?

src/main.rs
sharkdp: I think you need to store the values somehow as long as you have to pass something like `&args[start..end]` to `CommandTemplate::new` if the only thing you have is the `Iterator` over single values. An alternative might be to rewrite `CommandTemplate::new` to take an iterator. This way, you could maybe pass `X.take(n)` to `CommandTemplate::new(…)`.

(Maybe it would also help to provide both indices and the values to `ContinuousRangeIterator`?)

src/main.rs
sharkdp: I don't see any problem in moving this code to a separate function or lambda except for the `::new(…)` and `::new_batch` calls. One way to solve it would be to merge these two functions into a single one (`::new`) that takes a `ExecutionMode` argument.

sharkdp:
Thank you for your contribution!

sharkdp:
Not sure what to do with this in the current state, so I'm closing this for now to clean up my long list of open PRs.

Issue #522, opened 2019-12-31
Don't buffer command output when running on one thread?


guihkx:
Hey there.

I'm not sure if this is possible with `fd` already (sorry if it is), but wouldn't it make sense to *not* buffer command output, by default, when using `-j 1 / --threads 1`? For example:

```bash
$ fd --type directory --max-depth 1 --threads 1 --exec-batch bash -c "date && sleep 3"
```

Currently, `fd` will buffer the output of `bash` and only display it after the process exits. But in this particular case, since `bash` will always run consecutively and not in parallel, I think it would make sense to not buffer them.

What do you think? 
Thanks.

sharkdp:
Thank you very much for your feedback.

That sounds like a reasonable request to me. Not sure how difficult this would be to implement.

maxtriano:
Is this issue still open? If so I could take a stab at it

guihkx:
I guess so :)

maxtriano:
Wonderful, could you be a bit more specific on the behavior you're suggesting so I can get a better handle on where to start?

guihkx:
Okay, I think I can give you an example.

Imagine you have a folder with a bunch of audio files (mp3, aac, opus, flac). Now let's say you want to list only the `mp3` files, _but_ you want to listen to these audios as well, to make sure you're not deleting an important mp3 file, or whatever.

You'd do something like this:

```
$ fd -t f -e mp3 -j 1 -x mpv --no-audio-display {}
```

The `-j 1` flag is important there to ensure `mpv` won't get spawned multiple times so you can listen to the audios one by one.

But here's the problem: Since `fd` is buffering the output of `mpv`, you can't send keystrokes to `mpv` using the terminal.

Using `find`, however, this is possible:

```
$ find . -type f -name '*.mp3' -exec mpv --no-audio-display {} \;
```

Since the output isn't going to be buffered in this case, you could, for example, press `p` to pause the current audio, or `q` to quit `mpv` and move to the next audio, and so on.

sharkdp:
Also, any long-running command that shows intermediate output that you would like to see. For example: search for some files and copy all of them somewhere with `rsync` with its `--progress` option.

guihkx:
Closing, as this has been implemented by #620 

Thanks a lot for everyone's work! :)

PR #620, opened 2020-06-23
No buffering of command output when running on a single thread

maxtriano:
This is an implementation for #522

### DIFF
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -7,18 +7,20 @@ use crate::error::print_error;
 use crate::exit_codes::ExitCode;
 
 /// Executes a command.
-pub fn execute_command(mut cmd: Command, out_perm: &Mutex<()>, is_multithread: bool) -> ExitCode {
-    let output;
+pub fn execute_command(
+    mut cmd: Command,
+    out_perm: &Mutex<()>,
+    enable_output_buffering: bool,
+) -> ExitCode {
     // Spawn the supplied command.
-    if is_multithread {
-        output = cmd.output();
+    let output = if enable_output_buffering {
+        cmd.output()
     } else {
         // If running on only one thread, don't buffer output
         // Allows for viewing and interacting with intermediate command output
-        let child = cmd.spawn().expect("Failed to execute command");
-        output = child.wait_with_output();
-    }
-    
+        cmd.spawn().and_then(|c| c.wait_with_output())
+    };
+
     // Then wait for the command to exit, if it was spawned.
     match output {
         Ok(output) => {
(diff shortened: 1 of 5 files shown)


src/walk.rs
```diff
@@ -170,12 +170,17 @@ fn spawn_receiver(
 
     let show_filesystem_errors = config.show_filesystem_errors;
     let threads = config.threads;
-
+    // This will be used to check if output should be buffered when only running a single thread
+    let is_multithread: bool = if threads > 1 {
```
maxtriano, line 174: Added an 'is_multithread' bool that will be passed down to execute_command

sharkdp:
``` rs
if threads > 1 {
        true
    } else {
        false
    }
```
could be just `threads > 1`.

Also, maybe we should rename the variable and the parameter to `enable_output_buffering`? On the callee side, I think this would be much clearer. And here, on the caller side, we could add a comment like:
``` rust
// We only buffer the output if we are using multiple threads. Otherwise, parallel execution
// can lead to interleaved output:
let enable_output_buffering = threads > 1;
```

src/exec/command.rs
```diff
@@ -7,10 +7,18 @@ use crate::error::print_error;
...
-    let output = cmd.output();
-
+    if is_multithread {
+        output = cmd.output();
+    } else {
+        // If running on only one thread, don't buffer output
+        // Allows for viewing and interacting with intermediate command output
+        let child = cmd.spawn().expect("Failed to execute command");
+        output = child.wait_with_output();
+    }
```
sharkdp, line 20: You could rewrite this to:
``` rust
let output = if … {
  cmd.output()
} else {
  …
  child.wait_with_output()
};
```

src/exec/command.rs
```diff
@@ -7,10 +7,18 @@ use crate::error::print_error;
...
+    let output;
     // Spawn the supplied command.
-    let output = cmd.output();
-
+    if is_multithread {
+        output = cmd.output();
+    } else {
+        // If running on only one thread, don't buffer output
+        // Allows for viewing and interacting with intermediate command output
+        let child = cmd.spawn().expect("Failed to execute command");
```
sharkdp, line 18: This looks like it breaks our error handling. What happens for a non-existing command if you run it with `--threads=2` and `--threads=1`? The output should be the same.

sharkdp:

sharkdp:
@maxtriano Are you planning to continue working on this, or should somebody else take over?

guihkx:
@maxtriano Gentle ping.

I'm still very interested in this feature :)

sharkdp:
Looks like @maxtriano is not available anymore to work on this. Maybe someone else wants to finish this PR?

sharkdp (approved):
Thank you @tmccombs. This looks good to go, right?

tmccombs:
I think so.

PR #639, opened 2020-08-08
Fix a link to ubuntu official page

y-yagi:
Currently, https://packages.ubuntu.com/disco/fd-find shows an error.
This is because of the package page that unsupported version was removed.

So I removed the version from the link. That link shows a search result of `fd-find` and can find OSs that can install `fd-find`.


sharkdp:
Thank you!

Issue #624, opened 2020-07-06
Improve time option usage


jonforums:
For fd's [time options](https://github.com/sharkdp/fd/blob/master/src/app.rs#L411-L444) consider adding these two tweaks

-  `--newer` and `--older` [visible clap aliases](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.visible_aliases)
- default the time component to `00:00:00` when only a date is given

which would turn this verbose incantation

    $ fd -t f --change-newer-than '2020-06-29 00:00:00' abcF08m1002r00ec12843000000a5p /opt/datastore/data/prod/xml/schema/xyz/

into a slightly less verbose and more memorable

    $ fd -t f --newer 2020-06-29 abcF08m1002r00ec12843000000a5p /opt/datastore/data/prod/xml/schma/xyz/

Looks like you're using modified time metadata, so these shorter long names are potentially confusing. But I didn't like `--newermt` and `--oldermt`. Ah, the age old naming challenge.

sharkdp:
Thank you for your feedback.

Both ideas sound good to me!

FliegendeWurst:
This issue can be closed (PR #645 implemented both features).

sharkdp:
released in [v8.2.0](https://github.com/sharkdp/fd/releases/tag/v8.2.0).

jonforums:
The #645 changes to add `--newer` and `--older` aliases are not showing up in the short or long `--help` text, nor is there mention that the time component defaults to 00:00:00 when only a date is given.

~~~
$ fd --help
fd 8.2.1
...
        --changed-within <date|dur>
            Filter results based on the file modification time. The argument can be provided as a
            specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). '--change-
            newer-than' can be used as an alias.
            Examples:
                --changed-within 2weeks
                --change-newer-than '2018-10-27 10:00:00'
        --changed-before <date|dur>
            Filter results based on the file modification time. The argument can be provided as a
            specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). '--change-
            older-than' can be used as an alias.
            Examples:
                --changed-before '2018-10-27 10:00:00'
                --change-older-than 2weeks
~~~

sharkdp:
@jonforums Thanks. Let's fix that.

PR #645, opened 2020-09-03
improve time option usage

gorogoroumaru:
issue #624 

Currently, aliases for fd's time option are not so memorable and we have to specify full timestamp.

So, I changed the aliases for time option into more memorable ones and default the time component to 00:00:00 when only a date is given

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -411,7 +411,7 @@ pub fn build_app() -> App<'static, 'static> {
         .arg(
             Arg::with_name("changed-within")
                 .long("changed-within")
-                .alias("change-newer-than")
+                .alias("newer")
                 .takes_value(true)
                 .value_name("date|dur")
                 .number_of_values(1)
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -411,7 +411,7 @@ pub fn build_app() -> App<'static, 'static> {
         .arg(
             Arg::with_name("changed-within")
                 .long("changed-within")
-                .alias("change-newer-than")
```
tmccombs, line 414: I think removing this alias isn't backwards compatible. Why not keep both aliases?

gorogoroumaru:
Thank you @tmccombs!
I applied your suggested change.

tmccombs (approved):

sharkdp:
@tmccombs Thank you for reviewing and merging this! Just a minor thing: It would be great if we can ask contributors to adapt the `CHANGELOG.md` file accordingly. It's a new process that I have started that makes releases so much easier.

Something like:

---

Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
- Description what has been changed, see #123 (@user)
```
where `#123` links to the bug ticket and/or the PR and `user` is your username.

---

I guess we should add this to a PR template or some contributing guide.

tmccombs:
Ok, I'll do that next time.

PR #654, opened 2020-09-23
Add native Zsh completion

### LINKED ISSUE #189: Better integration with zsh
Currently autocomplete by `Tab` for `[<path>]` argument does not works.

smancill:
Replace the auto-generated Zsh completion with a full completion script. This script improves completion support for options and arguments, which is hard to obtain from an auto-generated completion. Fixes #189.

- Full completion support for external commands and options when using `--exec`/`--exec-batch`.
- The completion for `--type` and `--color` adds descriptions instead of  just values.
- Complete files when using `--ignore-file`.
- Stop completing files for the _pattern_ argument.
- Improve completion for _path_ arguments. Only directories are completed.
- Single-use options are not offered again.
- Updated exclusions lists for mutual-exclusive options.
- Add support for hidden options (as defined in `app.rs`). They are not offered for completion, but if present are considered (and their value may be completed).

The `before_deploy.bash` script has been updated accordingly, and `build.rs` doesn't generate the completion for Zsh anymore.

----

I've decided to rewrite my original script to closely follow [ripgrep's completion script](https://github.com/okdana/ripgrep/blob/master/complete/_rg), so I've kept the copyright notice. Most of the work was done by @okdana in https://github.com/BurntSushi/ripgrep/pull/970. I figured it would be good to use the same style. As commented in the script, full support for exclusion lists depends of options groups, which was added in Zsh 5.4. Older Zsh versions will just offer most options. I didn't copy the [Zsh completion reference](https://github.com/okdana/ripgrep/blob/b1e3de246c7cfd88f43a25cfdce19adf4d2c9283/complete/_rg#L441) from that script, though. I don't know if you want it.

Command completion for `--exec`/`--exec-batch` was taken from Zsh's `_find` completion. I guess `--owner` and `--size` could also be completed, but I am not sure how to do it.

The options are up to date with `master`, as defined in `app.rs`.

### Completion overview

Full set of completion options:

``` console
$ fd -<TAB>
>> completing option
--absolute-path   -a  -- show absolute paths instead of relative paths
--case-sensitive  -s  -- perform a case-sensitive search
--changed-before      -- limit search to files/directories modified before the given date/duration
--changed-within      -- limit search to files/directories modified within the given date/duration
--color           -c  -- declare when to colorize search results
--exact-depth         -- only search at the exact given directory depth
--exclude         -E  -- exclude files/directories that match the given glob pattern
--exec            -x  -- execute command for each search result
--exec-batch      -X  -- execute command for all search results at once
--extension       -e  -- filter search by file extension
--fixed-strings   -F  -- treat pattern as literal string instead of a regex
--follow          -L  -- follow symbolic links to directories
--full-path       -p  -- match the pattern against the full path instead of the basename
--glob            -g  -- perform a glob-based search
--help            -h  -- display help message
--hidden          -H  -- search hidden files/directories
--ignore-case     -i  -- perform a case-insensitive search
--ignore-file         -- add a custom, low-precedence ignore-file with .gitignore format
--list-details    -l  -- use a long listing format with file metadata
--max-depth       -d  -- set max directory depth to descend when searching
--max-results         -- limit number of search results to given count and quit
--min-depth           -- set directory depth to descend before start searching
--no-ignore       -I  -- don't respect .(git|fd)ignore and global ignore files
--no-ignore-vcs       -- don't respect .gitignore files
--owner           -o  -- filter by owning user and/or group
--print0          -0  -- separate search results by the null character
--regex               -- perform a regex-based search (default)
--size            -S  -- limit search by file size
--threads         -j  -- set the number of threads for searching and executing
--type            -t  -- filter search by type
--version         -v  -- display version information
-1                    -- limit to a single search result and quit
```

Options that cannot be repeated will not be offered again. In this example `--exact-depth` is not shown again, but `--extension` is:

``` console
$ fd --exact-depth=2 --extension=rs --ex<TAB>
>> completing option
--exclude     -- exclude files/directories that match the given glob pattern
--exec        -- execute command for each search result
--exec-batch  -- execute command for all search results at once
--extension   -- filter search by file extension
```

I've updated the set of excluded options, this is, options that should not be offered for completion when some given options are already present in the command line. For example, by using the following options:
``` completion
$ fd --no-ignore --case-sensitive --glob a* --list-details --max-depth 4 -<TAB>
```
then these options will not be offered: `-1 --absolute-path --exact-depth --exec --exec-batch --fixed-strings --ignore-case --max-results --no-global-ignore-file --no-ignore-vcs --print0`

Completing `-t`/`--type`:

``` console
$ fd --type=<TAB>
>> completing type
directory   d  -- directories
empty       e  -- empty files or directories
executable  x  -- executable (files)
file        f  -- regular files
pipe        p  -- named pipes (FIFOs)
socket      s  -- sockets
symlink     l  -- symbolic links
```

**Update.** Completing size limit prefix (`+` or `-`) and then the units after the number.

``` console
$ fd --size=<TAB>
>> completing size limit prefix (full format: <prefix><number><unit>)
+  -- file size must be greater or equal to
-  -- file size must be less than or equal to

$ fd --size=+10<TAB>
>> completing size limit units
B   -- bytes
K   -- kilobytes  (10^3  = 1000   bytes)
M   -- megabytes  (10^6  = 1000^2 bytes)
G   -- gigabytes  (10^9  = 1000^3 bytes)
T   -- terabytes  (10^12 = 1000^4 bytes)
Ki  -- kibibytes  ( 2^10 = 1024   bytes)
Mi  -- mebibytes  ( 2^20 = 1024^2 bytes)
Gi  -- gigibytes  ( 2^30 = 1024^3 bytes)
Ti  -- tebibytes  ( 2^40 = 1024^4 bytes)
```

Completing `-x`/`--exec` or `-X`/`--exec-batch` with external commands, subcommands and options until `;` is found:

``` console
$ fd -x gi<TAB>  # complete external command
>> completing external command
gid                         git-force-clone             git-release
ginstall                    git-fork                    git-rename-branch
git                         git-fresh-branch            git-rename-remote
...

$ fd -x git b<TAB>  # complete subcommand
>> completing main porcelain command
bisect -- find, by binary search, change that introduced a bug
branch -- list, create, or delete branches
bundle -- move objects and refs by archive
>> completing ancillary interrogator command
blame  -- show what revision and author last modified each line of a file

$ fd -x git blame --di<TAB>  # complete command options
>> completing option
--diff-algorithm   -- choose a diff algorithm
--diff-filter      -- select certain kinds of files for diff
--dirstat          -- generate dirstat by amount of changes
--dirstat-by-file  -- generate dirstat by number of files

$ fd -x git blame --diff-filter=A \; --t<TAB>  # back to fd completion again
>> completing option
--threads  -- set the number of threads for searching and executing
--type     -- filter search by type
```

**Update.** Completing owner and group (the colon after _user_ is automatically inserted, and removed when pressing space if no group completion is started):

``` console
$ fd --owner<TAB>
>> completing user
_applepay           _hidd              _sandbox                nixbld7
_appowner           _iconservices      _screensaver            nixbld8
_appserver          _installassistant  _scsd                   nixbld9
_appstore           _installer         _securityagent          nixbld10
...

$ fd --owner smancill:<TAB>
>> completing group
_appserveradm       _mbsetupuser            _xserverdocs
_appserverusr       _mcxalr                 accessibility
_appstore           _mdnsresponder          admin
_ard                _mobileasset            authedusers
....
```

Completing `--ignore-file`:

``` console
$ fd --ignore-file=<TAB>
>> completing file
CHANGELOG.md     Cargo.lock  LICENSE-APACHE  README.md     build.rs  doc/  tests/
CONTRIBUTING.md  Cargo.toml  LICENSE-MIT     appveyor.yml  ci/       src/
```

Completing `-c`/`--color`:

``` console
$ fd --color=<TAB>
>> completing when to colorize
always  -- always use colorized output
auto    -- show colors if the output goes to an interactive console
never   -- do not use colorized output
```

Completing _pattern_ doesn't actually offer completion, but shows a message to help remembering that the argument should be a pattern:

``` console
$ fd <TAB>
>> completing pattern
```

Completing _path_ arguments offers only directories (works with multiple paths):

``` console
$ fd '' <TAB>
>> completing directory
ci/     doc/    src/    tests/

$ fd '' doc src/<TAB>
>> completing directory
exec/    filter/
```

If `--search-path` is set, the _path_ arguments will not be completed:

``` console
$ fd --search-path=src -- '' <TAB>
>> no more arguments
```

Some options are not shown for completion, but they are supported if present (for example, when `-u` is present `-I` will not be offered). The options `--base-directory` and `--search-path` are between these, but they will completed if the user types `--ba<TAB>` or `--se<TAB>`.

### DIFF
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -1,0 +1,265 @@
+#compdef fd
+
+##
+# zsh completion function for fd
+#
+# Based on ripgrep completion function.
+# Originally based on code from the zsh-users project — see copyright notice
+# below.
+
+autoload -U is-at-least
+
+_fd() {
+  local curcontext="$curcontext" no='!' ret=1
+  local -a context line state state_descr _arguments_options fd_types fd_args
+  local -A opt_args
+
+  if is-at-least 5.2; then
+    _arguments_options=( -s -S )
+  else
+    _arguments_options=( -s )
+  fi
+
+  fd_types=(
+    {f,file}'\:"regular files"'
+    {d,directory}'\:"directories"'
+    {l,symlink}'\:"symbolic links"'
+    {e,empty}'\:"empty files or directories"'
+    {x,executable}'\:"executable (files)"'
+    {s,socket}'\:"sockets"'
+    {p,pipe}'\:"named pipes (FIFOs)"'
+  )
+
+  # Do not complete rare options unless either the current prefix
+  # matches one of those options or the user has the `complete-all`
+  # style set. Note that this prefix check has to be updated manually to account
+  # for all of the potential negation options listed below!
+  if
+    # (--[bpsu]* => match all options marked with '$no')
+    [[ $PREFIX$SUFFIX == --[bopsu]* ]] ||
+    zstyle -t ":complete:$curcontext:*" complete-all
+  then
+    no=
+  fi
+
+  # We make heavy use of argument groups here to prevent the option specs from
+  # growing unwieldy. These aren't supported in zsh <5.4, though, so we'll strip
+  # them out below if necessary. This makes the exclusions inaccurate on those
+  # older versions, but oh well — it's not that big a deal
+  fd_args=(
+    + '(hidden)' # hidden files
+    {-H,--hidden}'[search hidden files/directories]'
+
+    + '(no-ignore-full)' # all ignore files
+    '(no-ignore-partial)'{-I,--no-ignore}"[don't respect .(git|fd)ignore and global ignore files]"
+    $no'(no-ignore-partial)*'{-u,--unrestricted}'[alias for --no-ignore, when repeated also alias for --hidden]'
+
+    + no-ignore-partial # some ignore files
+    "(no-ignore-full --no-ignore-vcs)--no-ignore-vcs[don't respect .gitignore files]"
+    "!(no-ignore-full --no-global-ignore-file)--no-global-ignore-file[don't respect the global ignore file]"
+
+    + '(case)' # case-sensitivity
+    {-s,--case-sensitive}'[perform a case-sensitive search]'
+    {-i,--ignore-case}'[perform a case-insensitive search]'
+
+    + '(regex-pattern)' # regex-based search pattern
+    '(no-regex-pattern)--regex[perform a regex-based search (default)]'
+
+    + '(no-regex-pattern)' # non-regex-based search pattern
+    {-g,--glob}'[perform a glob-based search]'
+    {-F,--fixed-strings}'[treat pattern as literal string instead of a regex]'
+
+    + '(match-full)' # match against full path
+    {-p,--full-path}'[match the pattern against the full path instead of the basename]'
+
+    + '(follow)' # follow symlinks
+    {-L,--follow}'[follow symbolic links to directories]'
+
+    + '(abs-path)' # show absolute paths
+    '(long-listing)'{-a,--absolute-path}'[show absolute paths instead of relative paths]'
+
+    + '(null-sep)' # use null separator for output
+    '(long-listing)'{-0,--print0}'[separate search results by the null character]'
+
+    + '(long-listing)' # long-listing output
+    '(abs-path null-sep max-results exec-cmds)'{-l,--list-details}'[use a long listing format with file metadata]'
+
+    + '(max-results)' # max number of results
+    '(long-listing exec-cmds)--max-results=[limit number of search results to given count and quit]:count'
+    '(long-listing exec-cmds)-1[limit to a single search result and quit]'
+
+    + '(fs-errors)' # file-system errors
+    $no'--show-errors[enable the display of filesystem errors]'
+
+    + '(fs-traversal)' # file-system traversal
+    $no"--one-file-system[don't descend into directories on other file systems]"
+    '!--mount'
+    '!--xdev'
+
+    + dir-depth # directory depth
+    '(--exact-depth -d --max-depth)'{-d+,--max-depth=}'[set max directory depth to descend when searching]:depth'
+    '!(--exact-depth -d --max-depth)--maxdepth:depth'
+    '(--exact-depth --min-depth)--min-depth=[set directory depth to descend before start searching]:depth'
+    '(--exact-depth -d --max-depth --maxdepth --min-depth)--exact-depth=[only search at the exact given directory depth]:depth'
+
+    + filter-misc # filter search
+    '*'{-t+,--type=}"[filter search by type]:type:(($fd_types))"
+    '*'{-e+,--extension=}'[filter search by file extension]:extension'
+    '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:glob pattern'
+    '*'{-S+,--size=}'[limit search by file size]:size limit:->size'
+    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner and/or group:->owner'
+
+    + ignore-file # extra ignore files
+    '*--ignore-file=[add a custom, low-precedence ignore-file with .gitignore format]: :_files'
+
+    + '(filter-mtime-newer)' # filter by files modified after than
+    '--changed-within=[limit search to files/directories modified within the given date/duration]:date or duration'
+    '!--change-newer-than=:date/duration'
+    '!--newer=:date/duration'
+
+    + '(filter-mtime-older)' # filter by files modified before than
+    '--changed-before=[limit search to files/directories modified before the given date/duration]:date or duration'
+    '!--change-older-than=:date/duration'
+    '!--older=:date/duration'
+
+    + '(color)' # colorize output
+    {-c+,--color=}'[declare when to colorize search results]:when to colorize:((
+      auto\:"show colors if the output goes to an interactive console (default)"
+      never\:"do not use colorized output"
+      always\:"always use colorized output"
+    ))'
+
+    + '(threads)'
+    {-j+,--threads=}'[set the number of threads for searching and executing]:number of threads'
+
+    + '(exec-cmds)' # execute command
+    '(long-listing max-results)'{-x+,--exec=}'[execute command for each search result]:command: _command_names -e:*\;::program arguments: _normal'
+    '(long-listing max-results)'{-X+,--exec-batch=}'[execute command for all search results at once]:command: _command_names -e:*\;::program arguments: _normal'
+
+    + other
+    '!(--max-buffer-time)--max-buffer-time=[set amount of time to buffer before showing output]:time (ms)'
+
+    + '(about)' # about flags
+    '(: * -)'{-h,--help}'[display help message]'
+    '(: * -)'{-v,--version}'[display version information]'
+
+    + path-sep # set path separator for output
+    $no'(--path-separator)--path-separator=[set the path separator to use when printing file paths]:path separator'
+
+    + search-path
+    $no'(--base-directory)--base-directory=[change the current working directory to the given path]:directory:_files -/'
+    $no'(*)*--search-path=[set search path (instead of positional <path> arguments)]:directory:_files -/'
+
+    + args # positional arguments
+    '1: :_guard "^-*" pattern'
+    '(--search-path)*:directory:_files -/'
+  )
+
+  # Strip out argument groups where unsupported (see above)
+  is-at-least 5.4 ||
+  fd_args=( ${(@)args:#(#i)(+|[a-z0-9][a-z0-9_-]#|\([a-z0-9][a-z0-9_-]#\))} )
+
+  _arguments $_arguments_options : $fd_args && ret=0
+
+  case ${state} in
+    owner)
+      compset -P '(\\|)\!'
+      if compset -P '*:'; then
+        _groups && ret=0
+      else
+        if
+          compset -S ':*' ||
+          # Do not add the colon suffix when completing "!user<TAB>
+          # (with a starting double-quote) otherwise pressing tab again
+          # after the inserted colon "!user:<TAB> will complete history modifiers
+          [[ $IPREFIX == (\\|\!)*  && ($QIPREFIX == \"* && -z $QISUFFIX) ]]
+        then
+          _users && ret=0
+        else
+          local q
+          # Since quotes are needed when using the negation prefix !,
+          # automatically remove the colon suffix also when closing the quote
+          if [[ $QIPREFIX == [\'\"]* ]]; then
+            q=${QIPREFIX:0:1}
+          fi
+          _users -r ": \t\n\-$q" -S : && ret=0
+        fi
+      fi
+      ;;
+
+    size)
+      if compset -P '[-+][0-9]##'; then
+        local -a suff=(
+          'B:bytes'
+          'K:kilobytes  (10^3  = 1000   bytes)'
+          'M:megabytes  (10^6  = 1000^2 bytes)'
+          'G:gigabytes  (10^9  = 1000^3 bytes)'
+          'T:terabytes  (10^12 = 1000^4 bytes)'
+          'Ki:kibibytes  ( 2^10 = 1024   bytes)'
+          'Mi:mebibytes  ( 2^20 = 1024^2 bytes)'
+          'Gi:gigibytes  ( 2^30 = 1024^3 bytes)'
+          'Ti:tebibytes  ( 2^40 = 1024^4 bytes)'
+        )
+        _describe -t units 'size limit units' suff -V 'units'
+      elif compset -P '[-+]'; then
+        _message -e 'size limit number (full format: <+-><number><unit>)'
+      else
+        _values 'size limit prefix (full format: <prefix><number><unit>)' \
+          '\+[file size must be greater or equal to]'\
+          '-[file size must be less than or equal to]' && ret=0
+      fi
+      ;;
+  esac
+
+  return ret
+}
+
+_fd "$@"
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the zsh-users nor the
+#       names of its contributors may be used to endorse or promote products
+#       derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL ZSH-USERS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for fd
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * smancill (https://github.com/smancill)
+#
+# ------------------------------------------------------------------------------
+
+# Local Variables:
+# mode: shell-script
+# coding: utf-8-unix
+# indent-tabs-mode: nil
+# sh-indentation: 2
+# sh-basic-offset: 2
+# End:
+# vim: ft=zsh sw=2 ts=2 et
--- a/build.rs
+++ b/build.rs
@@ -26,6 +26,5 @@ fn main() {
     let mut app = build_app();
     app.gen_completions("fd", Shell::Bash, &outdir);
     app.gen_completions("fd", Shell::Fish, &outdir);
-    app.gen_completions("fd", Shell::Zsh, &outdir);
     app.gen_completions("fd", Shell::PowerShell, &outdir);
 }
--- a/ci/before_deploy.bash
+++ b/ci/before_deploy.bash
@@ -40,7 +40,7 @@ pack() {
     # various autocomplete
     cp target/"$TARGET"/release/build/"$PROJECT_NAME"-*/out/"$PROJECT_NAME".bash "$tempdir/$package_name/autocomplete/${PROJECT_NAME}.bash-completion"
     cp target/"$TARGET"/release/build/"$PROJECT_NAME"-*/out/"$PROJECT_NAME".fish "$tempdir/$package_name/autocomplete"
-    cp target/"$TARGET"/release/build/"$PROJECT_NAME"-*/out/_"$PROJECT_NAME" "$tempdir/$package_name/autocomplete"
+    cp contrib/completion/_"$PROJECT_NAME" "$tempdir/$package_name/autocomplete"
 
     # archiving
     pushd "$tempdir"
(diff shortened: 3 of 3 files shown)


smancill:
I am open to suggestions about the descriptions of the options, if something need to be changed.

smancill:
I was using "do not" and "don't" in the descriptions, now it is just "don't".

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+# below.
+
+autoload -U is-at-least
+
+_fd() {
+  local no='!' ret=1
+  local -a _arguments_options fd_file_types fd_args
+
+  if is-at-least 5.2; then
+      _arguments_options=( -s -S )
```
okdana, line 16: Indentation here and below is inconsistent with rest of file

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+    '!(--exact-depth -d --max-depth)--maxdepth:depth'
+    '(--exact-depth --min-depth)--min-depth=[set directory depth to descend before start searching]:depth'
+    '(--exact-depth -d --max-depth --maxdepth --min-depth)--exact-depth=[only search at the exact given directory depth]:depth'
+
+    + filter-misc # Filter search
+    '*'{-t+,--type=}"[filter search by type]:type:(($fd_file_types))"
+    '*'{-e+,--extension=}'[filter search by file extension]:extension'
+    '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:pattern'
+    '*'{-S+,--size=}'[limit search by file size]:size limit'
+    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner/group'
```
okdana, line 108: Not sure how this argument works, whether it's supposed to be `user:group` or what, but this could probably call `_users` if nothing else. (The `user:group` thing is so common that we should probably have a helper function for it, but we don't yet. It's not too hard to write out, see e.g. `_chown` or `_chroot`, but it's understandable if you want to avoid that)

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+    + '(no-ignore-full)' # all ignore files
+    '(no-ignore-partial)'{-I,--no-ignore}"[don't respect .(git|fd)ignore and global ignore files]"
+    $no'(no-ignore-partial)*'{-u,--unrestricted}'[alias for --no-ignore, when repeated also alias for --hidden]'
+
+    + no-ignore-partial # some ignore files
+    "(no-ignore-full --no-ignore-vcs)--no-ignore-vcs[don't respect .gitignore files]"
+    "(no-ignore-full --no-global-ignore-file)--no-global-ignore-file[don't respect the global ignore file]"
+
+    + '(case)' # case-sensitivity
+    {-s,--case-sensitive}'[perform a case-sensitve search]'
```
okdana, line 60: Typos here (`sensitve`)

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+    '(--exact-depth -d --max-depth)'{-d+,--max-depth=}'[set max directory depth to descend when searching]:depth'
+    '!(--exact-depth -d --max-depth)--maxdepth:depth'
+    '(--exact-depth --min-depth)--min-depth=[set directory depth to descend before start searching]:depth'
+    '(--exact-depth -d --max-depth --maxdepth --min-depth)--exact-depth=[only search at the exact given directory depth]:depth'
+
+    + filter-misc # Filter search
+    '*'{-t+,--type=}"[filter search by type]:type:(($fd_file_types))"
+    '*'{-e+,--extension=}'[filter search by file extension]:extension'
+    '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:pattern'
+    '*'{-S+,--size=}'[limit search by file size]:size limit'
```
okdana, line 107: For arguments like this it's common to describe the units/format in parentheses after the description. So this one might be like `size limit (bytes)`. If it accepts stuff like `100k` and `1G`, i'm not sure there's a common way of wording that, but maybe something like `size limits (bytes with optional suffix)`

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+
+    + '(threads)'
+    {-j+,--threads=}'[set the number of threads for searching and executing]:threads'
+
+    + '(exec-cmds)' # execute command
+    '(long-listing max-results)'{-x+,--exec=}'[execute command for each search result]:command: _command_names -e:*\;::program arguments: _normal'
+    '(long-listing max-results)'{-X+,--exec-batch=}'[execute command for all search results at once]:command: _command_names -e:*\;::program arguments: _normal'
+
+    + other
+    $no'(--max-buffer-time)--max-buffer-time=:milliseconds'
```
okdana, line 138: As with `--size`, you'd probably do like `buffer time (ms)`. If there is a pre-defined default, you could also put that in square brackets, as in `buffer time (ms) [1000]`

contrib/completion/_fd
```diff
@@ -0,0 +1,210 @@
...
+    + args # positional arguments
+    '1: :_guard "^-*" pattern'
+    '*:directory:_files -/'
+  )
+
+  # Strip out argument groups where unsupported (see above)
+  is-at-least 5.4 ||
+  fd_args=( ${(@)args:#(#i)(+|[a-z0-9][a-z0-9_-]#|\([a-z0-9][a-z0-9_-]#\))} )
+
+  _arguments $_arguments_options : $fd_args && ret=0
```
okdana, line 157: The `ret=` and `return` stuff here is redundant; since `_arguments` is the last command in the function its return value will just fall through. The reason many zsh completion functions do this `ret=` thing is that they need it to ensure that state-handling code passes back the correct status without immediately returning; you don't have any state handling here, so it's superfluous. Though if you expect to add that some day, maybe that's an argument for keeping it, i suppose

okdana:
Hey there. I don't use `fd` much, and i didn't test this or study the descriptions/organisation very much, but i did add some feedback if you're interested

contrib/completion/_fd
smancill: Fixed.

contrib/completion/_fd
smancill: Fixed.

contrib/completion/_fd
smancill: The arguments for `--owner` can be `user`, `user:group` or `:group`, and then any of those three can be prefixed with `!` to exclude matching files instead (maybe a `--not-owner` option would have been better). I don't understand from the manual if `user!:group` is valid (I don't think so).

I am still trying to figure out more advanced completion with Zsh. I think it can be done, but the optional `!` threw me off.

contrib/completion/_fd
smancill: I followed the man page and use now `(format: <+-><num><unit>)`.

contrib/completion/_fd
smancill: I set it to `set amount of time to buffer before showing output (ms)`

contrib/completion/_fd
smancill: I originally had it that way, without `ret`. But as you said, I though that in the future someone may add more logic to the function, like a `case $state in`, and may not know about the `ret` idiom, so I added it to make sure it is noticed.

smancill:
Thanks to @okdana  for reviewing, I have fixed the obvious issues and updated the descriptions. Since I force-pushed the commit, here is the full diff with the changes to the descriptions:

```diff
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -88,7 +88,7 @@ _fd() {
     '(long-listing exec-cmds)-1[limit to a single search result and quit]'
 
     + '(fs-errors)' # file-system errors
-    $no'--show-errors'
+    $no'--show-errors[enable the display of filesystem errors]'
 
     + '(fs-traversal)' # file-system traversal
     $no"--one-file-system[don't descend into directories on other file systems]"
@@ -104,39 +104,39 @@ _fd() {
     + filter-misc # filter search
     '*'{-t+,--type=}"[filter search by type]:type:(($fd_file_types))"
     '*'{-e+,--extension=}'[filter search by file extension]:extension'
-    '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:pattern'
-    '*'{-S+,--size=}'[limit search by file size]:size limit'
-    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner/group'
+    '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:glob pattern'
+    '*'{-S+,--size=}'[limit search by file size (format: <+-><num><unit>)]:size limit'
+    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner and/or group'
 
     + ignore-file # extra ignore files
     '*--ignore-file=[add a custom, low-precedence ignore-file with .gitignore format]: :_files'
 
     + '(filter-mtime-newer)' # filter by files modified after than
-    '--changed-within=[limit search to files/directories modified within the given date/duration]:date/duration'
+    '--changed-within=[limit search to files/directories modified within the given date/duration]:date or duration'
     '!--change-newer-than=:date/duration'
     '!--newer=:date/duration'
 
     + '(filter-mtime-older)' # filter by files modified before than
-    '--changed-before=[limit search to files/directories modified before the given date/duration]:date/duration'
+    '--changed-before=[limit search to files/directories modified before the given date/duration]:date or duration'
     '!--change-older-than=:date/duration'
     '!--older=:date/duration'
 
     + '(color)' # colorize output
     {-c+,--color=}'[declare when to colorize search results]:when to colorize:((
-      auto\:"show colors if the output goes to an interactive console"
+      auto\:"show colors if the output goes to an interactive console (default)"
       never\:"do not use colorized output"
       always\:"always use colorized output"
     ))'
 
     + '(threads)'
-    {-j+,--threads=}'[set the number of threads for searching and executing]:threads'
+    {-j+,--threads=}'[set the number of threads for searching and executing]:number of threads'
 
     + '(exec-cmds)' # execute command
     '(long-listing max-results)'{-x+,--exec=}'[execute command for each search result]:command: _command_names -e:*\;::program arguments: _normal'
     '(long-listing max-results)'{-X+,--exec-batch=}'[execute command for all search results at once]:command: _command_names -e:*\;::program arguments: _normal'
 
     + other
-    '!(--max-buffer-time)--max-buffer-time=:milliseconds'
+    '!(--max-buffer-time)--max-buffer-time=[set amount of time to buffer before showing output (ms)]:milliseconds'
 
     + '(about)' # about flags
     '(: * -)'{-h,--help}'[display help message]'
@@ -147,7 +147,7 @@ _fd() {
 
     + search-path
     $no'(--base-directory)--base-directory=[change the current working directory to the given path]:directory:_files -/'
-    $no'(*)*--search-path=:directory:_files -/'
+    $no'(*)*--search-path=[set search path (instead of positional <path> arguments)]:directory:_files -/'
 
     + args # positional arguments
     '1: :_guard "^-*" pattern'
```

I've also added the missing `--path-separator` option, sorry I missed it:

``` diff
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -142,6 +142,9 @@ _fd() {
     '(: * -)'{-h,--help}'[display help message]'
     '(: * -)'{-v,--version}'[display version information]'
 
+    + path-sep # set path separator for output
+    $no'(--path-separator)--path-separator=[set the path separator to use when printing file paths]:path separator'
+
     + search-path
     $no'(--base-directory)--base-directory=[change the current working directory to the given path]:directory:_files -/'
     $no'(*)*--search-path=:directory:_files -/'

```

And I made sure that `--search-path` is not offered if a _path_  positional argument is set:

```diff
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -151,7 +151,7 @@ _fd() {
 
     + args # positional arguments
     '1: :_guard "^-*" pattern'
-    '*:directory:_files -/'
+    '(--search-path)*:directory:_files -/'
   )
 
   # Strip out argument groups where unsupported (see above)

```

I've also fixed `--no-global-ignore-file` and `--max-buffer-time` to ensure they are never shown, since they are marked as hidden in `app.rs` and not shown in the help nor the man page (they need to be part of the arguments so the completion system can recognize them).

Since in the PR description I wasn't more clear, the options `--base-directory`, `--path-separator` and `--search-path` are not listed in the man page, so I have marked them with the `$no` prefix. The default value for `$no` is `!`, this is, the options will not be offered for completion. But following ripgrep's completion, if the user start typing any of those options, for example `--s<TAB>` or `--p<TAB>`, then they will be completed.

I've also included `--show-errors` and `--one-file-system` in this group of options that will not be shown by default. Not sure if it is OK, but I think they are rare options so they would just make the default completion listing too long. And I did the same for `--unrestricted`, I don't think it should be listed by default, most people will just use it like `-u` or `-uu` and there is no point showing it in the default completion list.

Let me know if you do want these options listed in the default completion list (i.e. `fd -<TAB>`).

contrib/completion/_fd
okdana: By convention, the units/default go in the description for the argument itself, not the option. See here for examples:

https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide#L66

That way, when the user does like `--max-buffer-time <TAB>`, they can still get a complete explanation of what the expected value is (even if zsh doesn't actually have any possibilities to offer).

contrib/completion/_fd
okdana: Without having tested it much (some escaping issues with `!`, but this is roughly how other functions do it), maybe something like this:

```
local ret=1
local -a context line state state_descr

# ...
'*'{-o+,--owner=}'[filter by owning user and/or group]: :->owner'
# ...

_arguments ... && ret=0

if [[ $state == owner ]]; then
  if compset -P '*:'; then
    compset -P '(\\|)!'
    _groups && ret=0
  else
    compset -P '(\\|)!'
    _users -qS : && ret=0
  fi
fi

return ret
```

Note the addition of the extra local variables at the top; you need those whenever you have state handling.

Also, i looked at the manual and saw that `--owner` can be supplied more than once, so i added `*` to the arg spec.

Hope that makes sense

smancill: I wasn't really planning to see how to complete `--owner` and `--size` yet, but you just kinda passed me the ball to do it now 😂 Just joking, your help is much appreciated. I have been trying to learn as much of possible of Zsh completion, I have been writing a few other completion scripts and that's why I ended up doing this one.

I have the `--owner` completion working with this, which is taking the same structure as `_chown` as you mentioned, but I tried to add some fixes due to the optional `!` prefix:

``` zsh
compset -p '(\\|)\!'
if compset -p '*:'; then
  _groups && ret=0
else
  if
    compset -s ':*' ||
    # do not add the colon suffix when completing "!user<tab>
    # (with a starting double-quote) otherwise pressing tab again
    # after the inserted colon "!user:<tab> will complete history modifiers
    [[ $iprefix == (\\|\!)*  && ($qiprefix == \"* && -z $qisuffix) ]]
  then
    _users && ret=0
  else
    local q
    # since quotes are needed when using the negation prefix !,
    # automatically remove the colon suffix also when closing the quote
    if [[ $qiprefix == [\'\"]* ]]; then
      q=${qiprefix:0:1}
    fi
    _users -r ": \t\n\-$q" -s : && ret=0
  fi
fi
```

First, without quotes it works just like `_chown`. But since quotes are needed in order to use the `!` prefix (I tried escaping the `!` but when the completion inserted a full value it would remove the backslash), I tried to fix a couple of annoyances:

- When using double-quotes, the completion will not insert the `:` suffix after the user. This is to fix the next <kbd>Tab</kbd> that is supposed to complete groups but it will show history modifiers instead if no closing double-quote is present. With single-quotes the suffix _is_ inserted. A <kbd>Tab<kbd> after that will complete the group just fine.

- When using quotes and the colon suffix is inserted, I use `-r` instead of `-q` to also enable removing the suffix when typing the closing quote (it just saves the <kbd>Backspace<kbd> to remove the colon and they typing the quote, but I think it is nice to have it)

As a side note, @sharkdp I think it would be better if in the manpage, the `--owner` part shows the exclusion example with single-quotes: `--owner '!john:students'`. The output of `--help` does it.

contrib/completion/_fd
smancill: I added completion for `--size` now, so I remove the `(format ...)` from the option description and added it to the completion message.

``` zsh
if compset -P '[-+][0-9]##'; then
  local -a suff=(
    'B:bytes'
    'K:kilobytes  (10^3  = 1000   bytes)'
    'M:megabytes  (10^6  = 1000^2 bytes)'
    'G:gigabytes  (10^9  = 1000^3 bytes)'
    'T:terabytes  (10^12 = 1000^4 bytes)'
    'Ki:kibibytes  ( 2^10 = 1024   bytes)'
    'Mi:mebibytes  ( 2^20 = 1024^2 bytes)'
    'Gi:gigibytes  ( 2^30 = 1024^3 bytes)'
    'Ti:tebibytes  ( 2^40 = 1024^4 bytes)'
  )
  _describe -t units 'size limit units' suff -V 'units'
elif compset -P '[-+]'; then
  _message -e 'size limit number (full format: <+-><number><unit>)'
else
  _values 'size limit prefix (full format: <prefix><number><unit>)' \
    '\+[file size must be greater or equal to]'\
    '-[file size must be less than or equal to]' && ret=0
fi
```

It first completes the prefix (`+` or `-`), and then after a number is inserted it completes the units.
If the prefix was typed manually and then <kbd>Tab</kbd> is pressed, it will show in the completion message that a number is needed.

@sharkdp the convention for GNU utils (find, du, ls, etc) is to use K,M,G,T... for powers of 1024, and KB,MB,GB,TB... for powers of 1000. Binary prefixes are also supported, with KiB=K, MiB=M, etc.

But fd is using K,M,G,T for powers of 1000. I think it may be good to follow convention.

contrib/completion/_fd
smancill: I've updated to `[set amount of time to buffer before showing output]:time (ms)`. Now the unit is in the completion message for the argument.

smancill:
As commented in the review discussion, I've added the completion for `--size` and `--owner`. I've also updated the PR description with the examples for them.

Here is the diff comparte to the previous force-push:

``` diff
diff --git a/contrib/completion/_fd b/contrib/completion/_fd
index 7c19463..20f8144 100644
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -10,8 +10,9 @@
 autoload -U is-at-least
 
 _fd() {
-  local no='!' ret=1
-  local -a _arguments_options fd_types fd_args
+  local curcontext="$curcontext" no='!' ret=1
+  local -a context line state state_descr _arguments_options fd_types fd_args
+  local -A opt_args
 
   if is-at-least 5.2; then
     _arguments_options=( -s -S )
@@ -105,8 +106,8 @@ _fd() {
     '*'{-t+,--type=}"[filter search by type]:type:(($fd_types))"
     '*'{-e+,--extension=}'[filter search by file extension]:extension'
     '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:glob pattern'
-    '*'{-S+,--size=}'[limit search by file size (format: <+-><num><unit>)]:size limit'
-    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner and/or group'
+    '*'{-S+,--size=}'[limit search by file size]:size limit:->size'
+    '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner and/or group:->owner'
 
     + ignore-file # extra ignore files
     '*--ignore-file=[add a custom, low-precedence ignore-file with .gitignore format]: :_files'
@@ -160,6 +161,56 @@ _fd() {
 
   _arguments $_arguments_options : $fd_args && ret=0
 
+  case ${state} in
+    owner)
+      compset -P '(\\|)\!'
+      if compset -P '*:'; then
+        _groups && ret=0
+      else
+        if
+          compset -S ':*' ||
+          # Do not add the colon suffix when completing "!user<TAB>
+          # (with a starting double-quote) otherwise pressing tab again
+          # after the inserted colon "!user:<TAB> will complete history modifiers
+          [[ $IPREFIX == (\\|\!)*  && ($QIPREFIX == \"* && -z $QISUFFIX) ]]
+        then
+          _users && ret=0
+        else
+          local q
+          # Since quotes are needed when using the negation prefix !,
+          # automatically remove the colon suffix also when closing the quote
+          if [[ $QIPREFIX == [\'\"]* ]]; then
+            q=${QIPREFIX:0:1}
+          fi
+          _users -r ": \t\n\-$q" -S : && ret=0
+        fi
+      fi
+      ;;
+
+    size)
+      if compset -P '[-+][0-9]##'; then
+        local -a suff=(
+          'B:bytes'
+          'K:kilobytes  (10^3  = 1000   bytes)'
+          'M:megabytes  (10^6  = 1000^2 bytes)'
+          'G:gigabytes  (10^9  = 1000^3 bytes)'
+          'T:terabytes  (10^12 = 1000^4 bytes)'
+          'Ki:kibibytes  ( 2^10 = 1024   bytes)'
+          'Mi:mebibytes  ( 2^20 = 1024^2 bytes)'
+          'Gi:gigibytes  ( 2^30 = 1024^3 bytes)'
+          'Ti:tebibytes  ( 2^40 = 1024^4 bytes)'
+        )
+        _describe -t units 'size limit units' suff -V 'units'
+      elif compset -P '[-+]'; then
+        _message -e 'size limit number (full format: <+-><number><unit>)'
+      else
+        _values 'size limit prefix (full format: <prefix><number><unit>)' \
+          '\+[file size must be greater or equal to]'\
+          '-[file size must be less than or equal to]' && ret=0
+      fi
+      ;;
+  esac
+
   return ret
 }
 
```

smancill:
I messed up squashing the new changes to a single commit. Now I just force-pushed a single commit again.

I think the completion script is almost ready to be merged, unless there is something that you think needs to be changed.

sharkdp:
@smancill Thank you so much for this fantastic contribution! Also, thank you very much for the detailed explanations, documentation and screenshots.

I have just tested this (Zsh user myself) and it's just great :smile:.

@okdana Thank you for the review.

> @sharkdp the convention for GNU utils (find, du, ls, etc) is to use K,M,G,T... for powers of 1024, and KB,MB,GB,TB... for powers of 1000. Binary prefixes are also supported, with KiB=K, MiB=M, etc.
> 
> But fd is using K,M,G,T for powers of 1000. I think it may be good to follow convention.

I know it's unfortunate, but I'd rather follow the official SI / EIC standards (see https://github.com/sharkdp/fd/pull/286#discussion_r183286795).

> As a side note, @sharkdp I think it would be better if in the manpage, the `--owner` part shows the exclusion example with single-quotes: `--owner '!john:students'`. The output of `--help` does it.

> As a side note, @sharkdp I think it would be better if in the manpage, the `--owner` part shows the exclusion example with single-quotes: `--owner '!john:students'`. The output of `--help` does it.

:+1:

sharkdp:
released in [v8.2.0](https://github.com/sharkdp/fd/releases/tag/v8.2.0).

PR #659, opened 2020-10-03
Travis config cleanup

reima:
This PR fixes two warnings and two infos in the Travis build config. They were reported by the Travis build config validation. See https://travis-ci.org/github/sharkdp/fd/builds/731765627/config for an example.

This is how I resolved the issues:

Validation message | Resolution
-|-
`root`: deprecated key `sudo` (The key \`sudo\` has no effect anymore.) | Removed `sudo`
`deploy`: deprecated key `skip_cleanup` (not supported in dpl v2, use cleanup) | Replaced `skip_cleanup: true` by `cleanup: false`
`deploy`: key `api_key` is an alias for `token`, using `token` | Replaced `api_key` by `token`
`root`: key `matrix` is an alias for `jobs`, using `jobs` | Replaced `matrix` by `jobs`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,6 @@ matrix:
       rust: 1.36.0
       env: TARGET=x86_64-apple-darwin
 
-sudo: required
-
 before_install:
   - ci/before_install.bash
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

Issue #642, opened 2020-08-21
Cannot build for Android arm64: undefined reference to getgrnam_r


kheaactua:
**Describe the bug you encountered:**

Cannot build for aarch64-linux-android21-clang with cargo-ndk

**Describe what you expected to happen:**

Building

**What version of `fd` are you using?**
v8.1.1

**Which operating system / distribution are you on?**

Host: `Linux sync-android 5.3.0-62-generic #56-Ubuntu SMP Tue Jun 23 11:20:52 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
Target:  `aarch64-linux-android21-clang`

Build command:
```
cargo ndk --platform 21 --target aarch64-linux-android install fd-find
```

Error:
```
...
   Compiling jemallocator v0.3.2
   Compiling fd-find v8.1.1
error: linking with `/opt/android-sdk/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android21-clang` failed: exit code: 1
  |
[...]

PR #662, opened 2020-10-08
Disable jemalloc on Android (fixes #636 and #642)

### LINKED ISSUE #636: Fd does not build inside termux on Android (aarch64) due to jemalloc usage
fd fails to build on aarch64

jemalloc not being used, so the build does not crash

8.1.1

Unix: Linux 4.4.226-lineageos aarch64 and *no lsb_release* (Android)

Appended logfile.

What kind of performance boost does jemalloc offer? Do you plan on supporting Android-like platforms?

fornwall:
Disabling jemalloc for Android fixes a build failure since jemalloc does not build out of the box for that platform on older API versions.

Android started using jemalloc as the system allocator around 2014 for most devices. In the latest Android version (11) the system allocator has been switched to Scudo, which while not being jemalloc is modern and actively maintained and should be used for most software on Android.

See:
- https://android-developers.googleblog.com/2020/06/system-hardening-in-android-11.html
- https://blog.nsogroup.com/a-tale-of-two-mallocs-on-android-libc-allocators-part-2-jemalloc/
- https://stackoverflow.com/questions/2266762/android-libc-version-and-malloc-implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Improved the usability of the time-based options, see #624 and #645 (@gorogoroumaru)
 
 ## Bugfixes
+
+- Disable jemalloc on Android, see #662
+
 ## Changes
 ## Other
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ libc = "0.2"
 # FIXME: Re-enable jemalloc on macOS
 # jemalloc is currently disabled on macOS due to a bug in jemalloc in combination with macOS
 # Catalina. See https://github.com/sharkdp/fd/issues/498 for details.
-[target.'cfg(all(not(windows), not(target_os = "macos"), not(target_env = "musl")))'.dependencies]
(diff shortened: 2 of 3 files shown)


fornwall:
@sharkdp For CI builds of Android, would you be ok to start using GitHub CI for that? Personally I find it easier and more flexible to use that instead of Travis, and many projects have started going in that direction.

sharkdp:



> @sharkdp For CI builds of Android, would you be ok to start using GitHub CI for that? Personally I find it easier and more flexible to use that instead of Travis, and many projects have started going in that direction.

Indeed. We have already switched my `bat` and `pastel` projects to GitHub Actions. If we start to use GitHub Actions, I'd like to move to it completely (and disable Travis and Appveyor). This would also mean that we have to port the whole deployment code to GitHub Actions.

I'll probably not work on this myself, but I would definitely be in favor of this change. The PR on the `bat` repository would probably be a good resource, as the CI system is quite similar:

https://github.com/sharkdp/bat/pull/1023 and https://github.com/sharkdp/bat/pull/1121

sharkdp:
fix released in [v8.2.0](https://github.com/sharkdp/fd/releases/tag/v8.2.0).

Issue #622, opened 2020-06-29
Document regex syntax


nu8:
I was looking to do a word boundary search, so I started with something like
this:

~~~
\<Sunday\>
~~~

as that is how Vim does it. That did not work, so I was inclined to open an FD
issue about it. It turns out that FD uses the Regex crate:

https://docs.rs/regex

which does support word boundary via this syntax:

~~~
\bSunday\b
~~~

my issue is that I feel the FD regex could be better documented. I dont believe
this repo even has any links to the Regex crate, but I could be mistaken.

sharkdp:
from the man page:
```
PATTERN SYNTAX
       The regular expression syntax used by fd is documented here:

       ⟨https://docs.rs/regex/1.0.0/regex/#syntax⟩

       The glob syntax is documented here:

       ⟨https://docs.rs/globset/#syntax⟩
```

But documentation can always be improved. I'm  happy for any suggestions!

nu8:
I mean, just adding a link here would be helpful:

https://github.com/sharkdp/fd#fd-doesnt-seem-to-interpret-my-regex-pattern-correctly

tmccombs:
> as that is how Vim does it.

vim's regex syntax is actually quite different from the more prevalent  "perl-compatible regular ecpressions" (PCRE), and I believe rust's regex syntax, which fd uses, is mostly a subset of PCRE.

PR #663, opened 2020-10-09
Add documentation around regex syntax

chills42:
This is to address https://github.com/sharkdp/fd/issues/622

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,9 @@ X11/xinit/xinitrc
 X11/xinit/xserverrc
 ```
 
-The regular expression syntax used is that of the [rust regex library](https://docs.rs/regex/1.0.0/regex/#syntax).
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much. A minor comment:

Users of `bat` will not really care (and maybe not know) if `bat` was written in Rust. This is how we phrase it in the man page:

```
PATTERN SYNTAX
       The regular expression syntax used by fd is documented here:

       ⟨https://docs.rs/regex/1.0.0/regex/#syntax⟩

       The glob syntax is documented here:

       ⟨https://docs.rs/globset/#syntax⟩
```

The glob-search option is not yet explained in the README, so we don't need the second part, for now.

sharkdp (approved):
Thank you!

PR #664, opened 2020-10-09
Dependency dirs is no longer mantained, use dirs-next

Spaceface16518:
All this does is change the manifest entry and usage of `dirs` to `dirs-next`. `dirs` is no longer maintained, so moving to `dirs-next` is encouraged. There are no user-facing changes (the crate is a drop-in replacement), so I didn't make a changelog entry, but I can if it's requested in a review. Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,20 +137,21 @@ version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "dirs"
-version = "2.0.2"
+name = "dirs-next"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cbcf9241d9e8d106295bd496bbe2e9cffd5fa098f2a8c9e2bbcbf09773c11a8"
 dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "dirs-sys 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "dirs-sys-next 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
-name = "dirs-sys"
-version = "0.3.4"
+name = "dirs-sys-next"
+version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c60f7b8a8953926148223260454befb50c751d3c50e1c178c4fd1ace4083c9a"
 dependencies = [
- "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.70 (registry+https://github.com/rust-lang/crates.io-index)",
  "redox_users 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


Spaceface16518:
I had to update the lockfile manually because the manifest format has slightly changed since the file was last updated in the repository. I don't think I messed anything up though.

sharkdp:
> `dirs` is no longer maintained

Could you please point me to information about this?

Do the dirs-next  functions really return the exact same directories on all platforms?

Spaceface16518:
> Could you please point me to information about this?

It was [announced on Reddit](https://www.reddit.com/r/rust/comments/gli4pc/directoriesnext_dirsnext_the_new_home_for/) about five months ago. There are several advantages to using `dirs-next` over `dirs`. `dirs-next` is more actively maintained (in fact, there was a new release yesterday), while the [`dirs` repository](https://github.com/dirs-dev/dirs-rs) has been archived. Also, there are many contributors to `dirs-next` (including the creator of `dirs`) and the crate is "owned" by multiple individuals so it is less likely to stagnate.

> Do the dirs-next functions really return the exact same directories on all platforms?

I'm not sure what you mean by this. The behavior of the functions is the same as `dirs` (as of now). The config dir will be `~/.config` on Linux, `{FOLDERID_RoamingAppData}` on Windows, and `~/.config` on MacOS (because of the correction). What exactly are you asking about?

tmccombs:
> What exactly are you asking about?

Can `dirs_next` replace `dirs` in a fully backwards-compatible way? Would merging this PR change the behavior of `fd` in any observable way?

Spaceface16518:
No, it will not change the existing behavior. The only change is that `fd` will have easier access to future bug fixes/features that come from `dirs-next`.

sharkdp:
@Spaceface16518 Thank you for the detailed explanation and references.

@paride Would this be fine from a Debian-packaging perspective?

paride:
@sharkdp fine in principle, but `dirs-next` is not packaged in Debian yet, and it will require to be before `fd` can start using it. Unfortunately this can be a lengthy process as new packages in Debian are manually vetted by a small team before being accepted as part of the distribution. Anyway if `dirs-next` is the way forward, then Debian will need to move on too. I appreciate a lot that you're taking the needs of downstream distributions into account, but in this case I'd say: just do what's best for `fd`.

What's not optimal here is probably that `dirs-next` wasn't able to takeover the `dirs` name, perhaps doing a major version bump in the library, but without requiring an actual change in the library name. But I guess this is how things are.

Cc: @sylvestre

sylvestre:
@sharkdp just like @paride 
Please ping me too in the future as I have been doing most of the packaging work on fd-find lately :)

Spaceface16518:
> What's not optimal here is probably that `dirs-next` wasn't able to takeover the `dirs` name, perhaps doing a major version bump in the library, but without requiring an actual change in the library name. But I guess this is how things are.

@paride yes, this was quite the frustration for the community in general. Basically, maintainer of `dirs` abandoned the crate and couldn't be contacted to transfer ownership of the crate, so the new maintainers decided to fork `dirs` and give ownership to an organization as well as some individuals. Then suddenly, the `dirs` maintainer released a new major version and everyone thought that we could move back to `dirs`, but then the `dirs` family repositories was archived and no new versions were released. As of now, `dirs-next` is the only crate that seems like it had any sort of future.

sharkdp (approved):

sharkdp:
@Spaceface16518 Thank you very much.

PR #666, opened 2020-10-10
add --base-directory, --path-separator, and --search-path to the man page

aswild:
These options are in `fd --help` but missing from the man page.
Also fix a typo in the help text for --base-directory.


sharkdp:
Thank you very much!

Issue #661, opened 2020-10-08
Exclude not working: `-E arg`


mcint:
**Describe the bug you encountered:**
Exclude flag not working. Despite calling the command with a flag to exclude items matching the given string, a file with the given string was included in the results.

I should sanity check for unicode or encoding issue, but I had just created the file by saving a modified file, typing the name myself, moments before.

**Describe what you expected to happen:**
File (or directory) matching pattern to be excluded.

**What version of `fd` are you using?**
`fd 8.0.0`

**Which operating system / distribution are you on?**
`Darwin 18.7.0 x86_64`

### Logs
```
$ fd '2020-10.*5.47' -E redact
Screen Shot 2020-10-07 at 5.47.21 PM redacted.png
Screen Shot 2020-10-07 at 5.47.21 PM.png
```

sharkdp:
Note that `-E` takes a glob pattern, not a regex. `-E '*redact*'` is probably what you want.

sharkdp:
If you think that we should improve the documentation, please let us know.

sharkdp:
@chirsz-ever I'd rather not change/break the current behavior. But I think we could make the documentation clearer.

chirsz-ever:
I think the current documents is clear enough:
https://github.com/sharkdp/fd/blob/99934ab98ce1247b789956eb969ff437ae553da7/src/app.rs#L326-L330
https://github.com/sharkdp/fd/blob/99934ab98ce1247b789956eb969ff437ae553da7/doc/fd.1#L146-L149

They explicitly point out that the pattern is a *glob* pattern. But at most time people instinctively want to use `-E expat` instead of `-E '*expat*'`.

Another improvement without breaking change is to add an `--exclude-regex` option, is this OK? @sharkdp

sharkdp:
> Another improvement without breaking change is to add an `--exclude-regex` option, is this OK? @sharkdp

Hm. That's something that is not going to work that easily with the `ignore`-based exclude mechanism. I think I'd rather like to avoid introducing a new option for this.

> I think the current documents is clear enough:

Maybe we could add an example to make it even better. Something like
```
--exclude '*.pyc' --exclude 'node_modules'
```
or a similar pattern that would match files that people would likely try to exclude.

PR #667, opened 2020-10-15
Add '--exclude-regex' option

chirsz-ever:
The purpose is to provide an alternate option using regex to exclude entries.

The option will affected the "smart case" option.

I don't know how to write man page, please help to complete it.


sharkdp:

As stated in #661, I'd rather not add a new option for this. As described in the README, `fd` is not trying to cover every possible use case. We are rather striving for simplicity, if possible.

Issue #600, opened 2020-05-23
"-c never" and NO_COLOR are not respected for -h and other usage messages


arp242:
When adding `-c never` I would expect it to also apply to the `-h` output, but that's still coloured:

![screenshot_2020-05-23-18-07-09_border](https://user-images.githubusercontent.com/1032692/82727968-56438000-9d20-11ea-8d90-ab13b7f9b4af.png)

This also applies to something like `fd -c none -v`, which is an error but has coloured (hard to read) output:

![screenshot_2020-05-23-18-12-03_border](https://user-images.githubusercontent.com/1032692/82728113-f39eb400-9d20-11ea-94c4-ffa2da620609.png)

In general, I'd really recommend to be careful with coloured terminal stuff, because what's helpful for one personal might be next-to-unreadable for the next (especially if it's non-configurable).

This is fd 8.1.0 on Void Linux.

sharkdp:
Thank you for reporting this. That should be fixed, if possible (I'm not entirely sure if clap allows us to read the value of `--color=…` when using `-h`, but I would hope so). Things could certainly be changed for `NO_COLOR`.

tpiekarski:
@sharkdp If you did not yet take a look, I'd like to try to debug and offer a PR

sharkdp:
> @sharkdp If you did not yet take a look, I'd like to try to debug and offer a PR

sounds great!

sharkdp:
@tpiekarski It's probably not completely trivial. We might need to set [`AppSettings::DisableHelpFlags`](https://docs.rs/clap/2.33.1/clap/enum.AppSettings.html#variant.DisableHelpFlags) and then implement `-h`/`--help` ourselves via [`App::write_help`](https://docs.rs/clap/2.33.1/clap/struct.App.html#method.write_help) after checking the `--color` value and setting [`ColorAlways`](https://docs.rs/clap/2.33.1/clap/enum.AppSettings.html#variant.ColorAlways)/[`ColorNever`](https://docs.rs/clap/2.33.1/clap/enum.AppSettings.html#variant.ColorNever) appropriately.

But maybe I'm missing something.

tpiekarski:
@sharkdp Thanks for taking a closer look. 
However I'll take my chances today/tomorrow and try to solve this issue.
Should I have questions I know I can ask them here.

tpiekarski:
[...]

PR #668, opened 2020-10-16
Support colorless help

xanonid:
Closes: #600

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,10 +1,17 @@
 use clap::{crate_version, App, AppSettings, Arg};
 
 pub fn build_app() -> App<'static, 'static> {
+    let clap_color_setting =
+        if std::env::var_os("NO_COLOR").is_none() {
+            AppSettings::ColoredHelp
+        } else {
+            AppSettings::ColorNever
+        };
+
     let mut app = App::new("fd")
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -1,10 +1,17 @@
 use clap::{crate_version, App, AppSettings, Arg};
 
 pub fn build_app() -> App<'static, 'static> {
+    let clap_color_setting =
+        if atty::is(atty::Stream::Stdout) && std::env::var_os("NO_COLOR").is_none() {
```
sharkdp, line 5: I think `clap` does an `isatty` check internally, because `fd --help | cat` is not colored at the moment. So I guess we could remove the `atty::is(…)` part here (as well as the build dependency)

sharkdp (changes requested):
Thank you very much for your contributioin!

src/app.rs
xanonid: isatty check removed (and rebased).

sharkdp (approved):
Thank you for the update.

Could you please add a CHANGELOG entry (see `CONTRIBUTING.md` section that I just added).

sharkdp:
Nevermind, I will add it.

PR #677, opened 2020-10-25
Add --no-extension flag

### LINKED ISSUE #673: Can not find file without extension

Though the folder has few files without extension, but the following command returns nothing.

`$ fd -t f -e ''`

I expected to have a output ~similar to the output of `find . -type f ! -name "*.*"`

fd 8.1.1

Linux 5.4.0-52-generic x86_64
No LSB modules are available.
Distributor ID:	Linuxmint
Description:	Linux Mint 20
Release:	20
Codename:	ulyana

jacobmischka:
Add shorthand flag to only show results with no file extension (including directories and other commonly extensionless files).

Fixes #673

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Add new --no-extension flag, see #673 (@jacobmischka)
 - Improved the usability of the time-based options, see #624 and #645 (@gorogoroumaru)
 - Add new `--prune` flag, see #535 (@reima)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,137 +4,156 @@
 name = "aho-corasick"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
 dependencies = [
- "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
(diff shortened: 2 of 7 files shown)


sharkdp:

I'd rather not add a new option for this niche use case (see `fd` goals in README: "While it does not seek to mirror all of find's powerful functionality, it provides sensible (opinionated) defaults for 80% of the use cases.").

I'd be great to update the `--help` text, though (similar to what I wrote in #673)

mailsanchu:
Is this --no-extension removed? I am getting an error when I use

tavianator:
@mailsanchu `--no-extension` was never added I think.  But https://github.com/sharkdp/fd/issues/673#issuecomment-715914285 gives some examples for how to achieve the same thing.

mailsanchu:
Thank you very much. Actually I want to delete all files with an extension. How we can achieve it using fd
Using find we can do this 
find "$PWD" -type f -not -name "*.*" -exec rm -rvf {} \;

tavianator:
> Actually I want to delete all files with an extension

Do you mean *without* an extension?

Your `find` command should use `-not -name '*.*'` I think, otherwise it will delete everything not named `.`.  (And `rm -r` isn't needed with `-type f`.)

For `fd`,

```
fd '^[^.]*$' --type f
```

will match all files without extensions, so

```
fd '^[^.]*$' --type f -X rm -vf
```

will delete them.

mailsanchu:
Do you mean without an extension?
No I mean files with extension ( wit dot in the file name)

mailsanchu:
I think I found a way to do it 
` fd '[.]' --type f -d1`

PR #682, opened 2020-10-30
Better defaults for LsColors

### LINKED ISSUE #604: Color Presets
_Originally posted by @sharkdp in https://github.com/sharkdp/fd/issues/469#issuecomment-633611703_
> Another option would be to ship `fd` with a more "complete"/colorful version of `LS_COLORS` for cases when it has not been set by the user. But that should be discussed in a separate ticket.

You could change `--color` to accept a color preset rather than always or never. `--color=rich` could be the preset in the demo, `--color=lite` could be the current default, `--color=none` could be no color, and `--color=default` could be the `LS_COLORS` variable with a fallback to the `lite` preset (or `none` if it doesn't pass the Windows-related checks). And it could read other presets from files, making sure to ignore carriage returns for compatibility purposes.

mjsir911:
Fixes #604

There might be a better place for the default constant


sharkdp:

Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? (see new section in `CONTRIBUTING.md` that I just added).

sharkdp:
Nevermind, I will do it.

Issue #669, opened 2020-10-18
exact values in file size filter


Rogach:
Currently `fd --size 0b` gives an error: `[fd error]: '0b' is not a valid size constraint.`

Documentation seems to require `+` or `-` prefixes (greater than or less than). Maybe we should handle cases without prefix as meaning "this exact size"? `find` does it this way - `--size 0` will find files with size of zero bytes.

(if such change is acceptable, I would be willing to provide a pull request)

sharkdp:
That sounds great, yes! Thank you.

sharkdp:
released in [v8.2.0](https://github.com/sharkdp/fd/releases/tag/v8.2.0).

dezza:
Why not just allow `--size=0` ? It makes no sense.

PR #696, opened 2020-12-02
Add exact size support to the --size filter (#669)

Rogach:
This will allow using `--size=100b` or `--size 100b` filters to search for files of that exact size.


sharkdp:

Issue #695, opened 2020-12-01
Documentation improvements


sharkdp:
It's always a good idea to work on documentation. I think it would also be good to restructure the README:

- [x] Add a quick link section at the top of the readme (see https://github.com/sharkdp/bat)
- [x] Extend the tutorial (rename "How to use"?)
- [x] Move the "Parallel command execution" section inside the "How to use" section). See also: #605 
- [x] Move "Installation" section below the "How to use" section

There are a few open issues [labeled with `documentation`](https://github.com/sharkdp/fd/labels/documentation):

- [ ] #692 
- [x] #684
- [x] #673
- [x] #661
- [x] #627
- [x] #605
- [ ] #374 
- [ ] #354 
- [x] #645

tstewart-writer:
I'd be happy to work on the documentation. I can start with the changes to the README, then see what else I can do.

sharkdp:
Sounds great. Several smaller PRs would be appreciated!

tstewart-writer:
Probably a good idea to do this in small PR. I'm pretty new to git/GitHub, so it will be easier for me as well.

sharkdp:
I think this can be closed. All major points have been addressed. We should - however - continue to work on the open `documentation` tickets. I have added all of them to the "fd 9" milestone.

PR #701, opened 2020-12-06
Move Installation/Development sections below the Tutorial section

tstewart-writer:
I moved the Installation and Development sections below the Tutorial section, as mentioned in #695. I thought I would do this before renaming and modifying the Tutorial.


sharkdp:
Thank you!

Issue #605, opened 2020-05-25
Improve the '--exec' help text


sharkdp:
I see a lot of `fd` users that seem to think that they need to add `"{}"` to all commands. Some users also seem to think that the closing semicolon (which needs to be escaped) `\;` is required:
```
fd … -x command "{}" \;
```

In reality, 90% of the `--exec` use cases can be written without the `"{}"` part (which `fd` adds, if it is ommitted) because the file entry often comes last. The semicolon is never required if we teach users to put the `--exec` argument last.

I would like to improve several things in the documentation:
- [x] the README section ["Parallel command execution"](https://github.com/sharkdp/fd#parallel-command-execution) should be merged with the tutorial section.
- [x] We should put less focus on teaching the `{}`/`{.}`/`{/}`/… syntax and show more simple examples like `fd … -x wc -l` or `fd … -X vim` first.
- [x] The `--help` text and the man page should also be updated accordingly. I think two or three simple examples should also go inside the `--help` text and definitely into the man page.
- [x] Clarify that `-x` does not take a string argument (see below)

ericonr:
Do you think the size of the explanations for `{}` and friends should be smaller? I quite like them as they are, tbh.

The examples should be the same for the man pages and the help output right?

sharkdp:
> Do you think the size of the explanations for `{}` and friends should be smaller? I quite like them as they are, tbh.

I would not make the text on this less detailed. I just would like it to be a little bit less in the spotlight. It's not the first thing users should learn when reading about `--exec`. I think they should see a few easy examples (and most common use cases) first.

> The examples should be the same for the man pages and the help output right?

Yes

Piraty:
the manpage definitely needs more attention regarding `-x`, it doesn't even mention the semicolon to end the command. I thought enclosing the command in quotes `"` would make sense as well?

sharkdp:
> the manpage definitely needs more attention regarding `-x`, it doesn't even mention the semicolon to end the command.

:+1: 



> I thought enclosing the command in quotes `"` would make sense as well?

Hm, not necessarily. `fd … -x wc -l` is correct. `fd … -x "wc -l"` will not work. `fd … -x "wc" "-l"` is fine, but not really needed.

Piraty:
[...]

PR #703, opened 2020-12-07
Move and update Parallel command execution section

tstewart-writer:
I moved the "Parallel command execution" section into the tutorial. Additionally, I made some changes based on the discussion in #605 to clarify a few of the things people were confused about, and moved examples above the syntax section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,34 +100,6 @@ for alternative, more complete (or more colorful) variants, see [here](https://g
 
 `fd` also honors the [`NO_COLOR`](https://no-color.org/) environment variable.
 
-## Parallel command execution
-If the `-x`/`--exec` option is specified alongside a command template, a job pool will be created
-for executing commands in parallel for each discovered path as the input. The syntax for generating
-commands is similar to that of GNU Parallel:
-
-- `{}`: A placeholder token that will be replaced with the path of the search result
-  (`documents/images/party.jpg`).
-- `{.}`: Like `{}`, but without the file extension (`documents/images/party`).
-- `{/}`: A placeholder that will be replaced by the basename of the search result (`party.jpg`).
-- `{//}`: Uses the parent of the discovered path (`documents/images`).
-- `{/.}`: Uses the basename, with the extension removed (`party`).
-
-``` bash
-# Convert all jpg files to png files:
-fd -e jpg -x convert {} {.}.png
-
-# Unpack all zip files (if no placeholder is given, the path is appended):
-fd -e zip -x unzip
-
-# Convert all flac files into opus files:
-fd -e flac -x ffmpeg -i {} -c:a libopus {.}.opus
-
-# Count the number of lines in Rust files (the command template can be terminated with ';'):
-fd -x wc -l \; -e rs
-```
-
-The number of threads used for command execution can be set with the `--threads`/`-j` option.
-
 ## Command-line options
 ```
 USAGE:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -314,6 +286,50 @@ If we want to run a command on all search results, we can pipe the output to `xa
 Here, the `-0` option tells *fd* to separate search results by the NULL character (instead of
 newlines). In the same way, the `-0` option of `xargs` tells it to read the input in this way.
 
+### Parallel command execution
+If the `-x`/`--exec` option is specified alongside a command template, a job pool will be created
+for executing commands in parallel for each discovered path as the input. The number of threads 
+used for command execution can be set with the `--threads`/`-j` option.
```
sharkdp, line 292: Maybe we can move the last sentence about threads to the bottom of this section. The default setting will be fine for 99% of use cases.

README.md
```diff
@@ -314,6 +286,50 @@ If we want to run a command on all search results, we can pipe the output to `xa
...
+
+*fd* takes the command template as a series of arguments rather than as a string. If you want to 
+add additional options after the command template, you can terminate it with a `\;`. This is 
+useful when you want to repeat a command with new options. For example, to remove write and 
+execute permissions from all directories, run:
```
sharkdp, line 297: It would be great if we could mention this whole `\;` business at the end of the section (before the threads option). I use `fd -x` and `fd -X` all the time and I never use `\;`. I'd rather teach the most common use cases first.

README.md
```diff
@@ -314,6 +286,50 @@ If we want to run a command on all search results, we can pipe the output to `xa
...
+The syntax for generating commands is similar to that of GNU Parallel:
+
+- `{}`: A placeholder token that will be replaced with the path of the search result
+  (`documents/images/party.jpg`).
+- `{.}`: Like `{}`, but without the file extension (`documents/images/party`).
+- `{/}`: A placeholder that will be replaced by the basename of the search result (`party.jpg`).
+- `{//}`: Uses the parent of the discovered path (`documents/images`).
+- `{/.}`: Uses the basename, with the extension removed (`party`).
+
+If you do not include a placeholder, *fd* automatically adds `{}`.
```
sharkdp, line 331: This is fine here, but maybe we should explain above what will actually be executed if you run `fd -e zip -x unzip` or `fd -e zip -X unzip`.

sharkdp (commented):
Thank you very much!

README.md
tstewart-writer: Sure, that makes sense. As long as it is there is some clarification anywhere in the section, I think it will be helpful.

README.md
tstewart-writer: Honestly, I didn't even think of the --exec-batch option. Should that be included in the Parallel command execution section, or would it be better to have that as a separate section?

tstewart-writer: Also, when you "say maybe we should explain above what will actually be executed" do you mean in the existing block of examples? After the examples? Or just above the line "If you do not include a placeholder, *fd* automatically adds `{}`." Sorry, I'm not quite sure what you have in mind.

tstewart-writer: I guess I was wondering if you meant something a bit like this; a short example which explicitly says which commands are run.

--------------
If the `-x`/`--exec` option is specified alongside a command template, a job pool will be created
for executing commands in parallel for each discovered path as the input.

In the following example, `fd` searches the current directory recursively for any file with the 
`zip` extension, and it runs `unzip` separately for each search result.
```bash
fd -e zip -x unzip
```
Thus, if there are two `zip` files, `file1.zip` and `example-dir/file2.zip`, the example runs `unzip file1.zip`
and `unzip example-dir/file2.zip`.

More Examples:
``` bash
# Convert all jpg files to png files:
fd -e jpg -x convert {} {.}.png

# Unpack all zip files (if no placeholder is given, the path is appended):
fd -e zip -x unzip

# Convert all flac files into opus files:
fd -e flac -x ffmpeg -i {} -c:a libopus {.}.opus

# Count the number of lines in Rust files (the command template can be terminated with ';'):
fd -x wc -l \; -e rs
```

sharkdp: > Honestly, I didn't even think of the --exec-batch option. Should that be included in the Parallel command execution section, or would it be better to have that as a separate section?

To me, the `-X`/`--exec-batch` option is so useful that it should be discussed in the same section, I think. We can simply rename it to "Command execution" maybe, because with `-X` there is no parallelism.

> I guess I was wondering if you meant something a bit like this; a short example which explicitly says which commands are run.

Yes, exactly! 

> Thus, if there are two `zip` files, `file1.zip` and `example-dir/file2.zip`, the example runs `unzip file1.zip`
> and `unzip example-dir/file2.zip`.

:+1: Maybe we could add: "The two `unzip` processes run in parallel (if the files are found fast enough)."

sharkdp:
@tstewart-writer Do you want to continue working on this or should somebody else take over?

sharkdp:
I'll actually just merge this and we can work on updates later.

Issue #697, opened 2020-12-03
--exec option doesn't honor --path-separator in Git Bash


LoopinFool:
In Git Bash, I use the alias `fd='MSYS_NO_PATHCONV=1 fd --path-separator=/'` so that everything is generally done with bash-style path names. This doesn't work with the --exec feature. This makes the output of "grep -H" annoying. The good news is that it functions - grep must accept either file separator. The bad news is that I get a path string I can't  usefully copy/paste in the shell to pass to programs like vim.

Examples:
`fd --type f ^filename.ext$`
Prints the full path of filename.ext with forward slashes

`fd --type f ^filename.ext$ --exec echo {}`
Prints the full path of filename.ext with backward slashes

I would expect --path-separator to apply everywhere, including paths passed to programs launched via the --exec argument.

Using fd version 8.1.1 - installed via chocolatey

Using the Git Bash that comes with Git for Windows version 2.28.0.windows.1

sharkdp:
Thank you for reporting this.

PR #711, opened 2021-01-03
use custom --path-separator in arguments to --exec

aswild:
When --path-separator is used, pass it on to commands run with
--exec(-batch) and --list-details.

Fixes: https://github.com/sharkdp/fd/issues/697

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -3,8 +3,9 @@ mod input;
 mod job;
 mod token;
 
-use std::ffi::OsString;
-use std::path::{Path, PathBuf};
+use std::borrow::Cow;
+use std::ffi::{OsStr, OsString};
+use std::path::{Component, Path, PathBuf, Prefix};
 use std::process::{Command, Stdio};
 use std::sync::{Arc, Mutex};
 
@@ -37,23 +38,24 @@ pub enum ExecutionMode {
 pub struct CommandTemplate {
     args: Vec<ArgumentTemplate>,
     mode: ExecutionMode,
+    path_separator: Option<String>,
 }
 
 impl CommandTemplate {
-    pub fn new<I, S>(input: I) -> CommandTemplate
+    pub fn new<I, S>(input: I, path_separator: Option<String>) -> CommandTemplate
     where
         I: IntoIterator<Item = S>,
         S: AsRef<str>,
     {
-        Self::build(input, ExecutionMode::OneByOne)
+        Self::build(input, ExecutionMode::OneByOne, path_separator)
     }
 
-    pub fn new_batch<I, S>(input: I) -> Result<CommandTemplate>
+    pub fn new_batch<I, S>(input: I, path_separator: Option<String>) -> Result<CommandTemplate>
     where
         I: IntoIterator<Item = S>,
         S: AsRef<str>,
     {
-        let cmd = Self::build(input, ExecutionMode::Batch);
+        let cmd = Self::build(input, ExecutionMode::Batch, path_separator);
         if cmd.number_of_tokens() > 1 {
             return Err(anyhow!("Only one placeholder allowed for batch commands"));
         }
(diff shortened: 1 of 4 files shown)


aswild:
There's definitely some corner cases (especially on Windows) that need to be tested here, especially when dealing with "absolute" paths. [`std::path::Prefix`](https://doc.rust-lang.org/std/path/enum.Prefix.html) can be six different things on Windows and I'm not sure how nicely they'll play with the new logic here.

Working with `OsStr(ing)` in Rust is as painful as ever, so I cast back to a `Path` and use its iterator over the components to re-join them with the custom separator.

I'll do some testing in MSYS when I get a chance.

sharkdp:
Awesome - thank you for your contribution. Happy to review this once the tests are finished.

aswild:
Refactored, tested on Windows, and force-pushed. I learned a lot more about Windows paths today.

Relative paths (originally with either `/` or `\` separators) work as expected, as do drive-letter absolute paths. UNC paths (`\\server\share\path`) also work, but only when using backslashes in `\\server\share` portion (Rust itself doesn't recognize `//server/share` as a UNC path).

Windows support may not be perfect for all the obscure corner cases, but if users are doing something weird, they can post-process fd's output using a tool like sed. Relative and drive-letter paths are surely the common use case, and here they work as expected.

src/exec/mod.rs
```diff
@@ -201,21 +207,33 @@ impl ArgumentTemplate {
...
+    pub fn generate(&self, path: impl AsRef<Path>, path_separator: Option<&str>) -> OsString {
         use self::Token::*;
+        let path = path.as_ref();
 
         match *self {
             ArgumentTemplate::Tokens(ref tokens) => {
                 let mut s = OsString::new();
                 for token in tokens {
+                    // note: the s.push() and replace_separator() calls can't be pulled out of this
+                    // match or else we get errors about references to temporary values
```
sharkdp, line 222: True, but I think that's to be expected. The alternative would be to copy or move the strings, but I think this is clearer.

If it's up to me, we can remove this comment.

src/exec/mod.rs
```diff
@@ -224,6 +242,61 @@ impl ArgumentTemplate {
...
+    /// interpreted as a Path and its components are iterated through and re-joined into a new
+    /// OsString.
+    fn replace_separator<'a>(path: &'a OsStr, path_separator: Option<&str>) -> Cow<'a, OsStr> {
+        // fast-path - no replacement necessary
+        if path_separator.is_none() {
+            return Cow::Borrowed(path);
+        }
+
+        let path_separator = path_separator.unwrap();
+        let mut out = OsString::with_capacity(path.len());
```
sharkdp, line 257: will only be enough if the path separator is on character, right? But okay, that's a strange corner case where we can live additional allocations, I guess.

src/exec/mod.rs
```diff
@@ -315,26 +393,62 @@ mod tests {
...
+                assert_eq!(arg.generate($input, Some("#")), OsString::from($expected));
+            };
+        }
+
+        check!("foo", "foo");
+        check!("foo/bar", "foo#bar");
+        check!("/foo/bar/baz", "#foo#bar#baz");
+
+        // windows-specific absolute path prefixes
+        if cfg!(windows) {
```
sharkdp, line 435: maybe move these to a separate `#[cfg(windows)]` test?

sharkdp:
Very nice!

Thank you for your effort with testing this on Windows as well!

src/exec/mod.rs
aswild: Yeah, I think that expected, but I fought with the borrow checker a bit here so I tend to document that. I can remove the comment though if you prefer.

src/exec/mod.rs
aswild: Yeah, I think the Vec behind OsString grows by doubling in size so there's probably only one or two extra allocations for multi-character separators. The only way to know exactly how big out needs to be is by iterating & counting the path components beforehand, so just letting OsString automatically resize seems like a reasonable thing to do.

src/exec/mod.rs
aswild: sure, I'll do that

sharkdp:
Thank you for the updates.

Issue #712, opened 2021-01-07
How fd behave with --threads 0 and --threads 1?


stephane-archer:
I don't want my `-x command` to be executed in parallel but **sequentially**.
I saw the option `--threads` but what number shoud I put?
is 0 and 1 the same? or 0 has a different meaning? is it "additional thead" in addition to the main thread?

tmccombs:
1 uses a single thread. 

From [this line](https://github.com/sharkdp/fd/blob/6a3aae52fdc60c6bc4b08d4122c7d91b2578c648/src/main.rs#L352), it appears that 0 is not a valid argument for the `--threads` option.

sharkdp:
Exactly. `-j 1` will launch one search thread and one thread for command execution. It ensures sequential execution (but note that the order is not deterministic).

You can test this with a script like
``` bash
#!/bin/bash

(
    flock -n 9 || (echo FAILED && exit 1)

    echo "sequential: $*"
    sleep 0.1
) 9> /tmp/sequential.lock
```

It will show a lot of "FAILED" messages is you run it with j > 1, but run fine with j == 1.

sharkdp:
Please let us know how the documentation could be improved. I guess we could expand the help text for `-j / --threads`?

stephane-archer:
Thanks a lot for your replies,
I was using `fd 8.1.1` and I was able to launch `fd . --threads 0 --extension JPG ./MyFolder -x echo {}`
without any error witch really confused me.
I just upgraded to `8.2.1` and I got the error message `[fd error]: Number of threads must be positive.` which is clearly the behavior I was expected in the first place.
I was using this feature for the first time and in the current version, I think it's really clear.
I didn't realize I was outdated and this issue was already fixed.
My apologies.

sharkdp:
>  I was using `fd 8.1.1` and I was able to launch `fd . --threads 0 --extension JPG ./MyFolder -x echo {}`
> without any error witch really confused me.

Ah, yes. That was only fixed recently.

>  I didn't realize I was outdated and this issue was already fixed.
> My apologies.

No worries. I think we could still improve the documentation. Both for the `--exec` option as well as for the `--threads` option. Running commands sequentially is probably a rather common use case. Let's leave it open until then.

stephane-archer:
I didn't expect the first time I use -x to be in parallel.
I think just adding this detail in the doc will be great because coming from `find` your script and you maybe didn't prepare for that case (I was not).

```
-x, --exec command
[...]

PR #715, opened 2021-01-09
Try to clarify the documentation for -x command

stephane-archer:
see issue #712

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -261,7 +261,7 @@ Provide paths to search as an alternative to the positional \fIpath\fR argument.
 .BI "\-x, \-\-exec " command
 Execute
 .I command
-for each search result. The following placeholders are substituted by a path derived from the current search result:
+for each search result in parallel (see --thread 1 for sequential order). The following placeholders are substituted by a path derived from the current search result:
 .RS
 .IP {}
 path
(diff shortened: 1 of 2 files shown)


tmccombs (approved):

stephane-archer:
What do you think about this change? Do I need to do something to help the review or merge process?

sharkdp:
Looks good - thank you very much. It would be great if we could also modify the `--help` text (`src/app.rs`).

doc/fd.1
```diff
@@ -261,7 +261,7 @@ Provide paths to search as an alternative to the positional \fIpath\fR argument.
 .BI "\-x, \-\-exec " command
 Execute
 .I command
-for each search result. The following placeholders are substituted by a path derived from the current search result:
+for each search result in parallel (see --thread 1 for sequential order). The following placeholders are substituted by a path derived from the current search result:
```
sharkdp, line 264: It should be `--threads 1` instead of `--thread 1`. Maybe:

`(use --threads=1 for sequential command execution)`

stephane-archer: I think you are right

stephane-archer: I think you are right

stephane-archer:
Do you have any feedback?

sharkdp:
Looks good - thank you very much!

stephane-archer:
@sharkdp no problem, I'm glad if I help. Please feel free to do what feels right for you.

Issue #728, opened 2021-02-11
Ctrl-T not working on bash in Ubuntu


chhajedji:
`fd --version`: `fd 8.1.1`

One suggestion:

Adding `FZF_DEFAULT_COMMAND='fd --type f` does not help in Debian even if aliased, as the package name is `fdfind`. Rather, `export FZF_DEFAULT_COMMAND='fdfind --type f` helped me. Maybe you want to update README with this.

Original query:

As suggested in the [tips](https://github.com/junegunn/fzf#tips), I wanted to map my ctrl-t in bash shell to do fuzzy search but exporting the variables `FZF_CTRL_T_COMMAND` and `FZF_DEFAULT_COMMAND` didn't help. But when ctrl-t is pressed, it only swaps last and second last letters and not what it's expected to do. What could I be missing?

![rec-2021-02-11_19 20 17](https://user-images.githubusercontent.com/42715380/107645465-b909ec80-6c9e-11eb-8313-263a5ae5a389.gif)

tmccombs:
That is because Debian calls the command `fdfind` instead of `fd`, on other distributions, or if you build it yourself, the command is `fd`.

chhajedji:
Yeah, got that. It is better to soft link `fd` rather than making an alias.
```sh
ln -s $(which fdfind) ~/.local/bin/
```
After this I can use `FZF_DEFAULT_COMMAND='fd --type f` on Debian also. If using an alias, I have to use `fdfind` instead of `fd`.

Also got the ctrl-t and other completions wokring by [sourcing key-bindings.bash from fzf](https://github.com/junegunn/fzf/blob/master/shell/key-bindings.bash)

tmccombs:
ah, yes. an alias won't work in thinks like `FZF_DEFAULT_COMMAND` because that command won't be run in a context that has access to bash aliases.

PR #729, opened 2021-02-12
README update: Use soft link instead of alias for Debian.

chhajedji:
As export command will not work with fd because fd is just an alias to
fdfind, create a soft link instead of alias.

Addresses https://github.com/sharkdp/fd/issues/728

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,8 +437,10 @@ If you run Ubuntu 19.04 (Disco Dingo) or newer, you can install the
 sudo apt install fd-find
 ```
 Note that the binary is called `fdfind` as the binary name `fd` is already used by another package.
-It is recommended that you add an `alias fd=fdfind` to your shells initialization file, in order to
-use `fd` in the same way as in this documentation.
+It is recommended that after installation, you add a link to `fd` by executing command
+`ln -s $(which fdfind) ~/.local/bin/fd`, in order to use `fd` in the same way as in this documentation.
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

mizlan:
Shouldn't you link it to `~/.local/bin/fd`? Only providing the bin directory and not the filename will create a symlink with the name `fdfind` instead of `fd` since the target file is `fdfind`.

chhajedji:
Ah right.. actually I already did it so didn't verify updated command in updated README ! My bad.

README.md
```diff
@@ -437,8 +437,10 @@ If you run Ubuntu 19.04 (Disco Dingo) or newer, you can install the
 sudo apt install fd-find
 ```
 Note that the binary is called `fdfind` as the binary name `fd` is already used by another package.
-It is recommended that you add an `alias fd=fdfind` to your shells initialization file, in order to
-use `fd` in the same way as in this documentation.
+It is recommended that after installation, you add a link to `fd` by executing command
+`ln -s $(which fdfind) ~/.local/bin/fd`, in order to use `fd` in the same way as in this documentation.
+You can also add `alias fd=fdfind` to your shells initialization file but this wouldn't help if you
+export variables using `fd`. You will have to use `fdfind` in export also.
```
sharkdp, line 443: I think we can even remove this part. Also below. Let's rather try to keep the install instructions short.

Issue #537, opened 2020-02-20
Option `--path-separator` failing on MSys (Git Bash)


fawick:
I have fd 7.4.0 installed on a Windows machine (Rust 1.41.0, stable-x86_64-pc-windows-msvc) and run it from a bash that is provided by [Git for Windows](https://gitforwindows.org/), which itself is based on Msys2 (according to https://github.com/git-for-windows/git/wiki/Technical-overview):

```
$ uname -a
MINGW64_NT-10.0-18363 SOMEHOST 3.0.7-338.x86_64 2019-07-03 15:16 UTC x86_64 Msys

$ fd
bar
bar\baz
bar\baz\bla
bar\baz\bla\some.txt
bar\yetanother.txt
other.txt
```

fd is using backslashes as it is supposed to do in Windows. #429 added the `--path-separator` option, which could be useful for using the results as paths within Git Bash. However, in my environment, the option fails, and every backslash is substituted with the absolute path to the Git for Windows installation directory:

```
$ fd --path-separator="/"
bar
barC:/Program Files/Git/baz
barC:/Program Files/Git/bazC:/Program Files/Git/bla
barC:/Program Files/Git/bazC:/Program Files/Git/blaC:/Program Files/Git/some.txt
barC:/Program Files/Git/yetanother.txt
other.txt

$ ls "C:/Program Files/Git/"
bin/  cmd/  dev/  etc/  git-bash.exe*  git-cmd.exe*  LICENSE.txt  mingw64/  ReleaseNotes.html  tmp/  unins000.dat  unins000.exe*  unins000.msg  usr/
```
Using single-quotes around the slash character make no difference. Nor does leaving out the quotes all together.

sharkdp:
It looks to me like this is a problem of Msys2 or Git Bash. Could you please try these suggestions: https://stackoverflow.com/a/54924640/704831 ?

fawick:
Thank you, that was a great hint, @sharkdp! I am very grateful to you for taking the time to dig into this.

For the record, one needs to set the environment variable `MSYS_NO_PATHCONV=1`:

```bash
$ fd --path-separator='/'
bar
barC:/Program Files/Git/baz
barC:/Program Files/Git/bazC:/Program Files/Git/bla
barC:/Program Files/Git/bazC:/Program Files/Git/blaC:/Program Files/Git/some.txt
barC:/Program Files/Git/yetanother.txt
other.txt

[...]

PR #730, opened 2021-02-14
set default path separator to '/' in MSYS

aswild:
MSYS and MSYS2 environments (such as Git Bash) have a UNIX like
filesystem which uses '/' as the path separator rather than '\', but
Rust doesn't know about this by default.

On Windows, check the MSYSTEM environment variable and set the default
value of the --path-separator option to '/' for convenience.

There is no similar detection of Cygwin because there seems to be no way
for Rust (and any native Win32) programs to detect that they're being
called from a Cygwin environment. Cygwin users can use a shell
alias/function/script to wrap fd.

Fixes: https://github.com/sharkdp/fd/issues/537

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,7 +173,9 @@ fn run() -> Result<ExitCode> {
         _ => ansi_colors_support && env::var_os("NO_COLOR").is_none() && interactive_terminal,
     };
 
-    let path_separator = matches.value_of("path-separator").map(|str| str.to_owned());
+    let path_separator = matches
+        .value_of("path-separator")
+        .map_or_else(filesystem::default_path_separator, |s| Some(s.to_owned()));
 
     let ls_colors = if colored_output {
         Some(LsColors::from_env().unwrap_or_else(|| LsColors::from_string(DEFAULT_LS_COLORS)))
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-use std::env::current_dir;
+use std::env;
 use std::ffi::OsStr;
 use std::fs;
 use std::io;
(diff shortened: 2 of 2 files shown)


aswild:
bah, MSRV strikes again - no Result::as_deref() until 1.47. Should be fixed now

src/main.rs
```diff
@@ -49,6 +49,23 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+    if cfg!(windows) {
+        let msystem = env::var("MSYSTEM").ok()?;
+        match msystem.as_str() {
+            "MINGW64" | "MINGW32" | "MSYS" => Some("/".to_owned()),
+            _ => None,
+        }
+    } else {
+        None
+    }
+}
```
sharkdp, line 67: Maybe we could move this function to `src/filesystem.rs`?

sharkdp:
Thank you very much!

This looks great. I added a minor inline comment.

> bah, MSRV strikes again - no Result::as_deref() until 1.47. Should be fixed now

I'm usually also very open to bumping the MSRV, but it doesn't seem to bad here to still support 1.40.

src/main.rs
aswild: done

aswild:
> I'm usually also very open to bumping the MSRV, but it doesn't seem to bad here to still support 1.40.

Yeah, it's no trouble to support 1.40 in this case, I'm just consistently surprised at how new of Rust features I use without even noticing (until CI fails)

sharkdp:
> until CI fails

If you want, rustup/cargo make it rather easy to have multiple Rust chains in parallel:
``` rust
rustup toolchain install 1.40
```
and then:
```
cargo +1.40 test
```

PR #733, opened 2021-02-24
Issue 624 newer older help text

### LINKED ISSUE #624: Improve time option usage

mitch292:
Only a documentation change so I didn't update the changelog. 

Resolves #624 by adding some additional help text for the new aliases that were originally implemented with PR #645 . 

Also completes the last checkbox in #695 . Let me know if you want me to make any changes!

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -462,7 +462,7 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
-                     '--change-older-than' or '--older' can be used as an alias.\n\
+                     '--change-older-than' or '--older' can be used as aliases.\n\
                      Examples:\n    \
                          --changed-before '2018-10-27 10:00:00'\n    \
                          --change-older-than 2weeks\n    \
--- a/src/app.rs
+++ b/src/app.rs
@@ -442,7 +442,7 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
-                     By default, for point in time, if no time is specified (HH:MM:SS), it defaults to 00:00:00. \
+                     By default, for point in time if no time is specified (HH:MM:SS) it defaults to 00:00:00. \
                      '--change-newer-than' or '--newer' can be used as an alias.\n\
                      Examples:\n    \
                          --changed-within 2weeks\n    \
--- a/src/app.rs
+++ b/src/app.rs
@@ -442,10 +442,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

src/app.rs
```diff
@@ -442,10 +442,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
-                     '--change-newer-than' can be used as an alias.\n\
+                     By default, for point in time if no time is specified (HH:MM:SS) it defaults to 00:00:00. \
```
sharkdp, line 445: I don't really understand that sentence. Maybe simply "If the time is not specified, it defaults to 00:00:00."

src/app.rs
```diff
@@ -442,10 +442,12 @@ pub fn build_app() -> App<'static, 'static> {
...
+                     '--change-newer-than' or '--newer' can be used as an alias.\n\
```
sharkdp, line 446: "can be used as aliases"?

src/app.rs
```diff
@@ -460,10 +462,11 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
-                     '--change-older-than' can be used as an alias.\n\
+                     '--change-older-than' or 'older' can be used as an alias.\n\
```
sharkdp, line 465: Should be `--older`

src/app.rs
mitch292: Yeah I like that better.

src/app.rs
mitch292: Will change

src/app.rs
mitch292: Ah sorry about that

mitch292:
Updated the PR!

src/app.rs
```diff
@@ -460,10 +462,11 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Filter results based on the file modification time. The argument can be provided \
                      as a specific point in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). \
-                     '--change-older-than' can be used as an alias.\n\
+                     '--change-older-than' or '--older' can be used as an alias.\n\
```
sharkdp, line 465: maybe here also "can be used as aliases", for consistency. Thank you for the updates!

mitch292: Ofc np, just pushed up the changes.

sharkdp (approved):

PR #747, opened 2021-03-14
doc: add link to GNU parallel

Delapouite:
Hello

A handy link for people wanting to dig more into GNU parallel

Thanks


sharkdp:
Thank you

Issue #714, opened 2021-01-08
Is there a thing like `--and`, `--or` or `--not`?


x3rAx:
I would like to combine queries like so:

```
fd --type symlink --and --owner --not root
```
or
```
fd --type symlink --or --type file
```

Is this somehow possible with `fd`?

x3rAx:
Ok, I just found `fd --help` (I used `fd -h` before which gave me a shortened help page).

So I found some things that work but they are not as consistent as I would wish:

It seems like `fd` automatically uses `and` or `or` based on what *"makes sense"*.

For example
```
fd --type symlink --owner root
```
is treated like an `and` whereas
```
fd --type symlink --type file
```
is treated like an `or`. This does not mean that using the same option twice is always treated like `or`:
```
fd --type directory --type file --type executable
```
is treated like `(directory OR file) AND executable` (only executable files and directories with the executable bit set are found).

I also found that searching files that are **not** owned by a specific user is possible using
```
fd --owner \!root
```
(or with quotes `fd --owner '!root'`).

This is nice, however, there seems to be no `not` for the file type:
```
fd --type \!symlink
```
gave an error:
```
error: '!symlink' isn't a valid value for '--type <filetype>...'
        [possible values: d, directory, e, empty, executable, f, file, l, p, pipe, s, socket, symlink, x]
```
(it should however be possible to use `fd -t directory -t file -t symlink -t pipe  -t socket` to get the same results, but I'm not sure if there is something missing)

## So the question remains:

Is there a way to explicitly combine options with `and`, `or` or `not`?

sharkdp:
> I would like to combine queries like so:
> 
> ```
> fd --type symlink --and --owner --not root
> ```

You can do this by running
```
fd --type symlink --owner '!root'
```
Different filters (type/owner/size/…) are combined using *AND*.

> or
> 
> ```
> fd --type symlink --or --type file
> ```

You can do this by running
```
fd --type symlink --type file
```
Multiple filters of the same kind are combined using *OR*.

> Ok, I just found `fd --help` (I used `fd -h` before which gave me a shortened help page).

There is a note at the end of the short `fd -h` help text: *"`fd -h` prints a short and concise overview while `fd --help` gives all details."*. Do you think it should be put somewhere else to make it better visible? You seem to have missed it, and you are probably not the only one.

> It seems like `fd` automatically uses `and` or `or` based on what _"makes sense"_.

Yes.

> This does not mean that using the same option twice is always treated like `or`:
[...]

PR #750, opened 2021-03-24
Detailed help visibility

x3rAx:
This PR aims to make it more visible that the `-h` and `--help` flags do different things.

As described in https://github.com/sharkdp/fd/issues/714#issuecomment-805752730, with the changes of this PR, the `-h` and `--help` flags are separated and moved to the beginning of the `FLAGS` section. The note at the bottom has been updated and is colorized. Also an empty line is inserted after the note:

For comparison, this is how it currently looks like:

2021-03-24--12-46-23--853__Code

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,8 @@ path = "src/main.rs"
 [build-dependencies]
 clap = "2.31.2"
 version_check = "0.9"
+lazy_static = "1.1.0"
+colored = "2.0.0"
 
 [dependencies]
 ansi_term = "0.12"
@@ -47,6 +49,7 @@ lscolors = "0.7"
 globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
+colored = "2.0.0"
 
 [dependencies.clap]
 version = "2.31.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ path = "src/main.rs"
 clap = "2.31.2"
 version_check = "0.9"
 lazy_static = "1.1.0"
-colored = "2.0.0"
+ansi_term = "0.12"
 
 [dependencies]
 ansi_term = "0.12"
@@ -49,7 +49,6 @@ lscolors = "0.7"
 globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
-colored = "2.0.0"
 
 [dependencies.clap]
 version = "2.31.2"
--- a/src/app.rs
+++ b/src/app.rs
@@ -16,6 +16,18 @@ pub fn build_app() -> App<'static, 'static> {
             "Note: `fd -h` prints a short and concise overview while `fd --help` gives all \
                  details.",
         )
+        // Help: Short and detailed
+        .arg(
+            Arg::with_name("help")
+                .short("h")
+                .help("Prints a short and concise overview")
+        )
+        .arg(
+            Arg::with_name("help_long")
+                .long("help")
+                .help("Prints detailed help information")
+        )
+        // Other flags
         .arg(
             Arg::with_name("hidden")
                 .long("hidden")
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,15 +1,16 @@
+use ansi_term::Colour::{Purple, Yellow};
 use clap::{crate_version, App, AppSettings, Arg};
-use colored::Colorize;
 use lazy_static::lazy_static;
 
 lazy_static! {
     static ref AFTER_HELP: String = format!(
-        "{} {} {} {} {} {}",
-        "Note:".yellow(),
-        "Use `fd",
-        "--help".purple(),
-        "` to get",
-        "detailed help".purple(),
+        "{} {} `{} {}` {} {} {}",
+        Yellow.paint("Note:"),
+        "Use",
+        "fd",
+        Purple.paint("--help"),
+        "to get",
+        Purple.paint("detailed help"),
         "information. `fd -h` only prints a short and concise overview.\n",
     );
 }
--- a/src/app.rs
+++ b/src/app.rs
@@ -10,7 +10,7 @@ lazy_static! {
         "--help".purple(),
         "` to get",
         "detailed help".purple(),
-        "information. `fd -h` only prints a short an concise overview.\n",
+        "information. `fd -h` only prints a short and concise overview.\n",
     );
 }
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,4 +1,18 @@
 use clap::{crate_version, App, AppSettings, Arg};
+use colored::Colorize;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref AFTER_HELP: String = format!(
+        "{} {} {} {} {} {}",
+        "Note:".yellow(),
+        "Use `fd",
+        "--help".purple(),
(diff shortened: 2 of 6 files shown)


sharkdp:
Thank you very much for your contribution. This looks great. Might take me a few more days until I find the time to properly review it.

x3rAx:
Hi, there seems to be no progress on this PR. If you didn't find time to review it yet, this is totally fine. I just wanted to bring it to your attention in case it went out of mind 🙂

Cargo.toml
```diff
@@ -47,6 +49,7 @@ lscolors = "0.7"
 globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
+colored = "2.0.0"
```
sharkdp, line 52: we already use the `ansi_term` dependency for colored output. It would be great if we could use it for this change as well.

src/app.rs
```diff
@@ -1,4 +1,18 @@
...
+
+lazy_static! {
+    static ref AFTER_HELP: String = format!(
+        "{} {} {} {} {} {}",
+        "Note:".yellow(),
+        "Use `fd",
+        "--help".purple(),
+        "` to get",
+        "detailed help".purple(),
+        "information. `fd -h` only prints a short an concise overview.\n",
```
sharkdp, line 13: typo: "short *and* concise"

src/main.rs
```diff
@@ -50,7 +50,37 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+        }
+    };
+
+    // Print short and long help
+    if matches.is_present("help") {
+        let _ = app.print_help();
+        // HACK: It seems like `clap` does not add a newline after the help
+        //       output when using `print_help`. Therefore add a newline here.
+        //       Note, that the `after_help` string already adds a newline to generate
+        //       an empty line before the next prompt
```
sharkdp, line 70: Thanks for the explanation, but I don't think we need to add it here. Especially not twice. And I don't consider it a "HACK". `print_help` doesn't print a newline, so we add one.

src/main.rs
```diff
@@ -50,7 +50,37 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
 ";
 
 fn run() -> Result<ExitCode> {
-    let matches = app::build_app().get_matches_from(env::args_os());
+    let mut app = app::build_app();
+
+    // When encountering parsing errors, print error message and exit
+    let matches = match app.get_matches_from_safe_borrow(env::args_os()) {
```
sharkdp, line 56: Could you explain why that change is necessary? get_matches_from => get_matches_from_safe_borrow

src/app.rs
x3rAx: Thanks, I fixed it

src/main.rs
x3rAx: From the [Clap documentation](https://docs.rs/clap/2.31.0/clap/struct.App.html#method.print_help):

> Prints the full help message to `io::stdout()` using a `BufWriter` using the same method as if someone ran `--help` to request the help message

So if I handle `-h | --help` manually and then call `print_help()`, I'd expect it to print the help message with a newline at the end, just. Because the documentation says that it would *use the same method as if someone ran `--help`* (which is obviously not the case), I'd consider this a bit as a hack to achieve what the `print_help` method has promised to do.

I can totally understand that you don't want to have a comment that says `HACK` in your tool. I use such comments to underline that this is a piece of code that works around some issue (mostly with external libraries) that might cause issues in the future, when the called functionality is fixed. So in this case, `clap` might decide to add the missing newline which would result in two empty lines in `fd`.

If you still want this comment to be removed, I'll of course do that. It is your tool after all :)

src/main.rs
x3rAx: Calling `get_matches_from` consumes `app` and thus `app` can not be used to call `print_help`:

Cargo.toml
x3rAx: Sure, I've changed it to use `ansi_term` and removed `colored`

src/main.rs
sharkdp: > using the same method

Well, okay. Then it's a question of whether or not this is a bug in `clap`. According to [this](https://github.com/clap-rs/clap/issues/1536), it looks like it is. Apparently it will be fixed in v3, but that has not been released yet.

In that case, the additional newline is a workaround for some upstream bug. I always considered a "hack" to be something else: a really ugly piece of code that does the right thing in principle, but in a horrible way. I don' think that's the case here. But now we're just down to nomenclature :smile:. I would personally prefer "workaround", but I'm also okay with the current wording, if it's really a bug in clap.

src/main.rs
sharkdp: Ok, thanks.

src/main.rs
x3rAx: Workaround is reasonable, I've changed it 😉

x3rAx:
It seems like the CI has failed because of low memory or so:

```
Build started
git clone -q https://github.com/sharkdp/fd.git C:\projects\fd
git fetch -q origin +refs/pull/750/merge:
git checkout -qf FETCH_HEAD
Running Install scripts
curl -sSf -o rustup-init.exe https://win.rustup.rs
rustup-init.exe --default-host %TARGET% --default-toolchain %CHANNEL% -y
info: profile set to 'default'
info: setting default host triple to i686-pc-windows-gnu
info: syncing channel updates for 'stable-i686-pc-windows-gnu'
info: latest update on 2021-05-10, rust version 1.52.1 (9bc8c42bb 2021-05-09)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-mingw'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: using up to 500.0 MiB of RAM to unpack components
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-mingw'
info: installing component 'rust-std'
info: installing component 'rustc'
memory allocation of 16777216 bytes failed     <--------------------------------- this line --------------
Command exited with code -1073740791
```

can you manually re-run the job?

Cargo.toml
x3rAx: I now remember why I added `colored`:

Using `ansi_term` I was unable to make the output colored **only** when the terminal supports it. This can for example be testet when piping to `cat`:

```
cargo run -- -h | \cat
```

At the current state, the last line containing the note for `--help` is still colored, which could lead to problems when the terminal does not support colors.

`colored` can somehow detect that and will disable coloring automatically.

Do you know a way to achieve the same with `ansi_term`?

x3rAx:
The tests are currently failing because in the output, the `NOTE: ...` line is colored by `ansi_term` and this somehow results in a missing `\n` (although it is not missing when running from a terminal...). When removing the colors, the tests succeed.

This is related to [my other comment](https://github.com/sharkdp/fd/pull/750#discussion_r633350441).

sharkdp:
Rebased and fixed conflicts

sharkdp:
Builds are currently failing...

sharkdp:

Issue #753, opened 2021-03-28
GUI for fd ?


Golddouble:
Is there a GUI for fd-search somewhere on the internet?
I have not found any.

Thank you.

sergmister:
I doubt a GUI would be of much benefit to the average user. If that is something your looking, for look for other applications such as kFind.

wolfisberg:
I am using `fd` a lot in conjunction with `rofi`, basically the gui version of piping `fd` into `fzf`.

Just pipe the `fd` output into `rofi`'s dmenu mode (or have a look at dmenu itself) and create whatever searches and lists you might need, your imagination is the limit. What I am using frequently is a PDF search:

      fd --type f -e pdf . ~ | rofi -monitor primary -keep-right -dmenu -i -p PDFS -multi-select | xargs -I {} xdg-open {}

I bound that to a shortcut. Hope this provides some inspiration.

sharkdp:
@wolfisberg I'm also using rofi, but I hadn't thought of this `fd`-`rofi` combination so far. I think it would make a great addition to the ["Integration with other programs"](https://github.com/sharkdp/fd#integration-with-other-programs) list.

wolfisberg:
@sharkdp Should I open a pull request with an added section to the "integration with other programs" list and provide a couple examples?

Golddouble:
Thank you @wolfisberg.

No support for regex?
I have tried this command:
`fd --type f -HI --regex  . ~ | rofi -monitor -5 -keep-right -dmenu -i -p PDFS -multi-select | xargs -I {} xdg-open {}
`

![FD](https://user-images.githubusercontent.com/13977359/114700977-845bf500-9d22-11eb-9405-45b370206d3e.gif)

sharkdp:
@wolfisberg That would be great!

@Golddouble If you want your search *in rofi* to support regexes, you need to enable regex search in rofi (with `-matching regex`), not in `fd`. In these examples, `fd` is basically just used to list all files.

Golddouble:
Unfortunately exactly the same behaviour with:
`fd --type f -H  . ~ | rofi -monitor -5 -matching regex -keep-right -dmenu -i -p PDFS -multi-select | xargs -I {} xdg-open {}`

wolfisberg:
@Golddouble Well, the `^book` regex would have to match against a **path** that starts with *book*, which can't exist. In this example, the matching is done against the entire path, that's why you don't get results.

Golddouble:
[...]

PR #758, opened 2021-04-14
Add rofi to integration with other programs in README.md

wolfisberg:
As discussed in [ISSUE #753](https://github.com/sharkdp/fd/issues/753).

I did not add an entry to the change log, since I think it's not noteworthy and there is already a general documentation improvement entry under *other*.

Let me know if it's too long or not explicit enough. I tried to mimic the style of the other integration entries.


sharkdp (approved):

Issue #692, opened 2020-11-23
 How to search for files based on a "whole-word" name


longcp:
I just want to find sepolicy files like the image, not others, can  fd do this ?
![image](https://user-images.githubusercontent.com/10282683/99968906-b94bc780-2dd4-11eb-92cf-467ce0dc445e.png)

tavianator:
`fd '^selinux$'`

sharkdp:
... or use a glob-based search:
```
fd -g selinux
```

sharkdp:
I think it would be a good idea to add this example to the README. https://github.com/sharkdp/fd#tutorial

PR #779, opened 2021-06-22
Added example for searching exactly a filename

joseph62:
Hello,

This PR is meant to address issue #692.

Let me know if any tweaks would make the example more clear. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,15 @@ src/lscolors/mod.rs
 tests/testenv/mod.rs
 ```
 
+### Searching for a particular file name
+
+ To find files with exactly the provided search pattern, use the `-g` (or `--glob`) option:
+``` bash
+> cd /bin
+> fd -g sh
+sh
+```
+
 ### Hidden and ignored files
 By default, *fd* does not search hidden directories and does not show hidden files in the
 search results. To disable this behavior, we can use the `-H` (or `--hidden`) option:
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ tests/testenv/mod.rs
 
  To find files with exactly the provided search pattern, use the `-g` (or `--glob`) option:
 ``` bash
-> cd /bin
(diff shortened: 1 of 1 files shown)


sharkdp:

It would be great if we could find a more realistic example where the file is not directly located in the search folder. And where the filename contains a file-extension. This makes the glob option interesting, as the `.` would be e special character in a regex. Maybe:
```
fd -g httpd.conf
```

joseph62:
That is a good point about the lack of realism in the example I used. I was trying to find one that hit these points:

1. The desired filename is a subset of other available filenames
2. The example is reproducible on the user's own machine

I wanted the example's output to be different than what a user might see if they drop the `-g` flag. The output of `fd sh` would include shell executables like `bash` and `ksh`.

Looking at this on my laptop now instead of my WSL installation I'm seeing an example that potentially hits both the criteria that I used as well as your points. 

```bash
$ cd /etc
$ fd -g console.perms
security/console.perms
$ fd console.perms # what it looks like without -g on my laptop
security/console.perms
security/console.perms.d
```

I did look at the proposed `fd httpd.conf` example on my machine and found that the results were the same with and without the `-g` flag. 

Let me know if this will work or if you have any other concerns.

sharkdp:
Thank you for your comment. I agree with all points. Unfortunately, I can not reproduce your example on my machine. Another proposal:
```bash
▶ fd -g libc.so /usr
/usr/lib32/libc.so
/usr/lib/libc.so

 ▶ fd libc.so /usr   
/usr/lib32/libc.so
/usr/lib32/libc.so.6
/usr/lib/libc.so.6
/usr/lib/libc.so
```

But I'm also okay with your example.

joseph62:
I love that! Applied.

sharkdp (approved):

sharkdp:
thank you for the update!

PR #785, opened 2021-06-30
Adds path separator checking and hint for automatic `/` expansion 

kalebo:
Posix compatibility shells on windows like msys2 treat `\` as an escape character which makes pasting the paths from `fd`s output annoying when using the platform default path separator. The immediate solution would be to set an alias for `fd` in these shells to use `/` as the path separator using `--path-separator`. However these shells will automatically expand `/` to be the absolute path to the shell's installation directory.  This leads to some surprising results you see paths output from `fd --path-separator /` in these shells.

This patch copies ripgrep's [approach](https://github.com/BurntSushi/ripgrep/blob/9f924ee187d4c62aa6ebe4903d0cfc6507a5adb5/crates/core/args.rs#L1348) of preventing multi-byte path separators entirely and provides a hint about how to avoid the expansion (by supplying `//` instead).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -177,18 +177,21 @@ fn run() -> Result<ExitCode> {
         .value_of("path-separator")
         .map_or_else(filesystem::default_path_separator, |s| Some(s.to_owned()));
 
-    if let Some(ref sep) = path_separator {
-        if sep.len() > 1 {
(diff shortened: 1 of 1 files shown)


sharkdp:

I made a few small adjustments. Hope that's fine for you.

PR #797, opened 2021-07-20
Add comment to README about making sure .local/bin is on PATH

### LINKED ISSUE #791: Command 'fd' not found - Ubuntu 20.04.2

After installing fd-find either via apt on the command line or through Synaptic, I get:

Command 'fd' not found, but can be installed with:

sudo apt install fdclone

...which appears to be some other program entirely.

I did, also, try linking the fd command per the README file for installing on Ubuntu, but I get the same result.

Expected the 'fd' command to work after installing fd-find

Synaptic says, 7.4.0-2build1

Linux 5.11.0-7614-generic x86_64
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal

tmccombs:
To make it more clear how to set up the symlin on debian/ubuntu.

Fixes #791


sharkdp (approved):
Thanks!

PR #801, opened 2021-08-02
Add dependabot config for cargo updates

dbast:
Creates update PRs similar to https://github.com/dbast/fd/pulls with history etc for review.

PRs can be merged if CI is green or closed to ignore the dependency regarding auto updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you! I think I would prefer a monthly update only. At least that's how we set up dependabot on some of my other projects.

dbast:
Ok, changed.

sharkdp:
Thank you!

Issue #436, opened 2019-05-07
Display directories with trailing slashes


rockandska:
HI,

It will be great if there will be a possibility to have directories displayed with a trailing "/" (as an option or not).
When you don't use colors, there is no way to distinct directories from files.

Regards,

sharkdp:
Thank you for the feedback.

I don't think there is any precedent for this (`ls` or `find`). I'd rather not add a trailing slash to directory names.

tavianator:
There is precedent in `ls -p`

sharkdp:
> There is precedent in `ls -p`

Thanks, I didn't know that.

My main point is that I'd rather not add a command-line option for something minor like this, as I really strive to keep the CLI small. So the question becomes: Do we want to add a trailing slash by default? (I'd rather not). If we don't, should there be some other way to make this possible (e.g. a more generic `--output-format` option)? I currently don't see the need for this.

Silly workaround :smile::
```
fd … -x ls -p
```

rockandska:
In my opinion, if `fd` should be an alternative to find (gnu?), it should be able to reproduce the same behaviour at the maximum :) and if you would like to add a trailing slash to directories with `find` (gnu) you could by chaining `-type` / `-printf`.
For example:
```
find . \( -type d -printf '%p/\n' -or -printf '%p\n' \)
```

And parsing `ls` ouput is indeed a silly workaround ^^

sharkdp:
> And parsing `ls` ouput is indeed a silly workaround ^^

`fd … -x ls -p` does not parse the output of `ls`. It executes `ls -p` for every search result and directly shows the `ls` output. It's not a great solution because it spawns a subprocess for every search result. But it works perfectly fine.

Arguably, using `find . \( -type d -printf '%p/\n' -or -printf '%p\n' \)` is worse. It's not only much longer to type, it can also not be used as an `alias` (I believe), because you can not simply append filtering criteria like `-name '…'`.

> In my opinion, if `fd` should be an alternative to find (gnu?), it should be able to reproduce the same behaviour at the maximum :)

I'm assuming you mean *"… at the minimum"*.

The very first paragraph in the README explicitly says: *"fd […] does not seek to mirror all of find's powerful functionality […]"*. So no, even if `find` would support this out of the box, we are not necessarily aiming to support this as well.

sharkdp:
I'd like to close this. Let me know if there is anything left that should be discussed.

blueForestIcarus:
[...]

PR #812, opened 2021-08-08
Append trailing path separators to directories

yyogo:
Solves #436.

Add a trailing slash (or custom path separator, if provided) to directories.
This only affects output; path args for `exec` are processed normally.


sharkdp:



>  When `--path-separator` is supplied display that instead.

:+1: 

>  Also, pass `-p` to `ls` when using `--list-details`.

Nice! ... but:

> Add a `-T/--trailing-slash` flag to help distinguish files from directories when not using colors.
> Do we go with `-T`?

As my comments in #436 suggest (*"What I would vote against is having a new command line option for this."*), I would rather not add a new command-line option for this, but rather make this the new default, if we decide to go for it. Now that I think about it again, I'm not sure why I was so strictly against adding a new option. Maybe because it makes interaction with other tools potentially inconsistent (if they can't rely on it behaving either way). I would not add a new short flag for this, however. `--trailing-slash` is okay. `--trailing-path-separator` would be more accurate. This seems to be an option that users either use in scripts or in an `alias fd="fd --trailing-path-separator"`, so the length of the name shouldn't matter too much. What do you think?

> To determine if the entry is a directory, we make an additional `stat` call. This shouldn't be that significant performance-wise, but maybe we could cache the result from the colorization and use that instead?

`stat` calls can definitely be the bottleneck. Especially if there are MANY search results. I'm pretty sure we can design a benchmark to measure the difference (unless the OS somehow caches the results?). We should definitely look at this.

tmccombs:
We already have to know if it is a directory to know if it should be recursively searched right? Could that be propogated to where we print the path to avoid an additional stat call?

yyogo:
@sharkdp - renaming the flag and dropping the short flag seems fine IMO, I mostly added it as a placeholder anyway.

yyogo:
I've refactored WorkerResult to contain a DirEntry instead of just the path, this seems to slightly improve performance since we don't need to reallocate the path buffer, and makes the additional stat call unnecessary.

yyogo:
another consideration: GNU ls allows setting an arbitrary "indicator" for directories rather than just a path separator. Do we want to support this feature as well?

sharkdp:
> I've refactored WorkerResult to contain a DirEntry instead of just the path, this seems to slightly improve the performance of `spawn_senders()` since we don't need to reallocate the path buffer, and makes the additional stat call unnecessary.

Thank you very much. Are you still available to work on this? Can we get rid of the new command-line option and make this the new default?

yyogo:
Hey, yeah I can fix the conflicts and flip the flag so it can be disabled, would that be ok?

sharkdp:
> Hey, yeah I can fix the conflicts and flip the flag so it can be disabled, would that be ok?

sounds great

sharkdp:
I would actually still prefer to completely remove the command-line option. We don't need everything to be configurable in `fd`. It's an "opinionated" tool with a (hopefully) good set of defaults. And I don't really see why someone would want to remove the trailing slash except for aesthetic reasons. Do you?

wookayin:
I second this. Removing trailing slashes is fairly easy using regex, but the opposite (adding trailing slashes only if a directory) is hard so why we are implementing this feature.

However, one possible scenario I could imagine is the backward compatibility. For example, we ship this feature from some particular version, say fd v8.3. From fd 8.3+ it will show trailing spaces by default. Some scripts or other tools written in the future would assume this default behavior of trailing slashes on directories when manipulating paths, but when it runs with *older versions of fd* (which can happen in many circumstances, e.g., a user haven't updated fd yet) it will behave differently. For example,`somedirectorysomefile` instead of `somedirectory/somefile`. If we had an explicit command-line option to be fail-proof, it would instead throw an error. I am not saying that the default shouldn't change --- I agree directories should have trailing slashes appended.

That said, this would be just a rare corner cases so I'm happy with changing the behavior to as we wish in #436. Having command-line options doesn't completely eliminate the possibility of such cases and users can rely on version checking or something like that..

tmccombs:
It looks like there are now some merge conflicts. @yyogo would you be able to resolve those?

yyogo:
I'm on it

yyogo:
there are a bunch of tests I have to fix, that might take a bit

sharkdp:
The test changes will likely clash with the ones in https://github.com/sharkdp/fd/pull/861. So let's maybe wait for that to be merged?

yyogo:
there's a pretty severe regression with `--color=never` (~2x time), I assume it's because we now have to stat each file to check if it's a folder

yyogo:
@sharkdp ready to merge

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 
 ## Changes
 
+- Directories are now printed with an additional path separator at the end: `foo/bar/`
```
sharkdp, line 52: I have released a new version in the meantime, so it would be great if we could move this line up into the "unreleased" section.

src/exec/job.rs
```diff
@@ -25,8 +25,8 @@ pub fn job(
 
         // Obtain the next result from the receiver, else if the channel
         // has closed, exit from the loop
-        let value: PathBuf = match lock.recv() {
-            Ok(WorkerResult::Entry(path)) => path,
+        let value: DirEntry = match lock.recv() {
+            Ok(WorkerResult::Entry(val)) => val,
```
sharkdp, line 29: In the meantime, this has been changed on `master`. Could we change this here accordingly (`path` instead of `val`)?

sharkdp: well no. Actually now it's a `dir_entry`. So let's change `val` to `dir_entry`. And also `value` => `dir_entry`.

src/exec/job.rs
```diff
@@ -53,7 +53,7 @@ pub fn batch(
     limit: usize,
 ) -> ExitCode {
     let paths = rx.iter().filter_map(|value| match value {
-        WorkerResult::Entry(path) => Some(path),
+        WorkerResult::Entry(val) => Some(val.path().to_owned()),
```
sharkdp, line 56: Hm. Can we prevent the cloning here (`.to_owned()` essentially clones the input path)? We might be able to use `into_iter` instead of `iter`?

src/entry.rs
```diff
@@ -0,0 +1,62 @@
+use std::{
```
sharkdp, line 1: Great that you moved this to a new module. Maybe it should be called `dir_entry`?

yyogo:
Hey @sharkdp I fixed the conflicts with the recent prs. Can you merge?

sharkdp:
I'm very sorry, but I haven't been fast enough and there have been other merges to master that introduced new conflicts in the meantime.

I would also appreciate some feedback from the other contributors (@tmccombs @tavianator) on this change.

rkulla:
I really needed this today when piping non colorized output to a file that then does some syntax folding based on directories ending with a /. Hope we get this soon since I will need to generate reports like this a lot.

For now I am manually adding them by separately running `ls -aF | grep /` to identify the files and appending them manually.

src/exec/job.rs
```diff
@@ -39,7 +41,7 @@ pub fn job(
         // Drop the lock so that other threads can read from the receiver.
         drop(lock);
         // Generate a command, execute it and store its exit code.
-        results.push(cmd.execute(&value, Arc::clone(&out_perm), buffer_output))
+        results.push(cmd.execute(&dir_entry, Arc::clone(&out_perm), buffer_output))
```
tmccombs, line 44: I messed this up when I resolved conflicts. This should be:
```suggestion
        results.push(cmd.execute(dir_entry.path(), Arc::clone(&out_perm), buffer_output))
```

src/output.rs
```diff
@@ -38,15 +45,43 @@ pub fn print_entry<W: Write>(stdout: &mut W, entry: &Path, config: &Config) {
...
+    stdout: &mut W,
+    entry: &DirEntry,
+    config: &Config,
+    style: Option<&Style>,
+) -> io::Result<()> {
+    if entry.file_type().map_or(false, |ft| ft.is_dir()) {
+        let separator = config
+            .path_separator
+            .as_ref()
+            .unwrap_or(&MAIN_SEPARATOR_STR);
```
tmccombs, line 62: I believe this will have to perform an atomic operation every time we print an entry. It would probably be more performant to store the actual path seperator once in config rather than in a `Lazy`. 

Although looking at it some more, it might actually be better to change `path_separator` to be an `Option<char>` instead of an `Option<String>`, since we require it to be a single byte anyway.

tmccombs (commented):
I few minor comments, but nothing major.

src/output.rs
tmccombs: I'll look more into changing this to use `char` in a seperate branch.

tmccombs: hmm, actually it is only required to be a single byte on windows. Although I don't know if there is much of a use case for it to be more than one character.

tmccombs: yeah, I don't think it's worth changing this to use a char. 

However, I do think having a `config.actual_path_separator` that is either the wrapped value of `config.path_separator` or the default path separator.

src/output.rs
```diff
@@ -3,28 +3,35 @@ use std::io::{self, Write};
...
 use lscolors::{Indicator, LsColors, Style};
+use once_cell::sync::Lazy;
 
 use crate::config::Config;
+use crate::dir_entry::DirEntry;
 use crate::error::print_error;
 use crate::exit_codes::ExitCode;
 use crate::filesystem::strip_current_dir;
 
+static MAIN_SEPARATOR_STR: Lazy<String> = Lazy::new(|| std::path::MAIN_SEPARATOR.to_string());
```
tmccombs, line 14: should this use `filesystem::default_path_separator instead of `std::path::MAIN_SEPARATOR`?

src/walk.rs
```diff
@@ -36,7 +35,7 @@ enum ReceiverMode {
 
 /// The Worker threads can result in a valid entry having PathBuf or an error.
 pub enum WorkerResult {
-    Entry(PathBuf),
+    Entry(DirEntry),
```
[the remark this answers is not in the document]
yyogo: the boxed version seems to be slightly slower:
<details>
  <summary>benchmark</summary>

## `fd` regression benchmark

### No pattern

|:---|---:|---:|---:|---:|
| `./fd-no-box-entry --hidden --no-ignore '' '../linux'` | 152.1 ± 4.6 | 143.2 | 162.3 | 1.00 |
| `./fd-box-entry --hidden --no-ignore '' '../linux'` | 155.9 ± 3.2 | 151.4 | 161.9 | 1.02 ± 0.04 |

### Simple pattern

|:---|---:|---:|---:|---:|
| `./fd-no-box-entry '.*[0-9]\.jpg$' '../linux'` | 152.3 ± 3.4 | 147.1 | 160.0 | 1.00 |
| `./fd-box-entry '.*[0-9]\.jpg$' '../linux'` | 170.1 ± 4.3 | 161.2 | 175.4 | 1.12 ± 0.04 |

### Simple pattern (-HI)

|:---|---:|---:|---:|---:|
| `./fd-no-box-entry -HI '.*[0-9]\.jpg$' '../linux'` | 157.0 ± 4.7 | 147.6 | 165.7 | 1.00 |
| `./fd-box-entry -HI '.*[0-9]\.jpg$' '../linux'` | 160.2 ± 3.8 | 154.3 | 166.7 | 1.02 ± 0.04 |

### File extension

|:---|---:|---:|---:|---:|
| `./fd-no-box-entry -HI --extension jpg '' '../linux'` | 168.3 ± 3.4 | 161.9 | 174.5 | 1.00 |
| `./fd-box-entry -HI --extension jpg '' '../linux'` | 171.4 ± 3.2 | 165.0 | 176.6 | 1.02 ± 0.03 |

### File type

|:---|---:|---:|---:|---:|
| `./fd-no-box-entry -HI --type l '' '../linux'` | 158.4 ± 2.9 | 153.4 | 164.8 | 1.00 |
| `./fd-box-entry -HI --type l '' '../linux'` | 162.5 ± 3.5 | 157.8 | 169.3 | 1.03 ± 0.03 |

</details>

<details>
  <summary>diff</summary>

```diff
diff --git a/src/exec/job.rs b/src/exec/job.rs
index 304792b..93946d7 100644
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -25,7 +25,7 @@ pub fn job(
 
         // Obtain the next result from the receiver, else if the channel
         // has closed, exit from the loop
-        let dir_entry: DirEntry = match lock.recv() {
+        let dir_entry: Box<DirEntry> = match lock.recv() {
             Ok(WorkerResult::Entry(dir_entry)) => dir_entry,
             Ok(WorkerResult::Error(err)) => {
                 if show_filesystem_errors {
diff --git a/src/walk.rs b/src/walk.rs
index 7c22227..d67d0b3 100644
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -35,7 +35,7 @@ enum ReceiverMode {
 
 /// The Worker threads can result in a valid entry having PathBuf or an error.
 pub enum WorkerResult {
-    Entry(DirEntry),
+    Entry(Box<DirEntry>),
     Error(ignore::Error),
 }
 
@@ -180,7 +180,7 @@ struct ReceiverBuffer<W> {
     /// The deadline to switch to streaming mode.
     deadline: Instant,
     /// The buffer of quickly received paths.
-    buffer: Vec<DirEntry>,
+    buffer: Vec<Box<DirEntry>>,
     /// Result count.
     num_results: usize,
 }
@@ -545,7 +545,7 @@ fn spawn_senders(
                 }
             }
 
-            let send_result = tx_thread.send(WorkerResult::Entry(entry));
+            let send_result = tx_thread.send(WorkerResult::Entry(Box::new(entry)));
 
             if send_result.is_err() {
                 return ignore::WalkState::Quit;

```
</details>

src/output.rs
tmccombs: oh, nevermind, in that case `path_separator` should be set to a `Some`.

sharkdp:
I ran some regression benchmarks and it looks like this makes `fd` a few percent slower, but I guess I'm fine with merging it anyway. Thank you very much for this implementation and for your patience :smile:

tavianator:
@sharkdp I have some ideas on how to recover the performance, assuming it's caused by the `PathBuf -> DirEntry` change in `WorkerResult`.  I was experimenting with that the other day trying out `Colorable`.

sharkdp:
@tavianator Cool! I hope it was okay to merge this... I believe we fixed the merge conflicts often enough by now :smile:

tavianator:
@sharkdp Haha yeah it's fine, I can build on top of this

reuleaux:
Sorry for coming in late to this issue (I have only just started to use _fd_): I don't want trailing slashes for directories - in fact they get in my way such that I cannot use _fd_ and have to stick  to good old _find_:

I am using _fd_ (or _find_) to find all the directories in my home dir, like so:
```
fdfind -d 1 -td . /home/rx
```
output:
```
/home/rx/cfg/
/home/rx/channel/
/home/rx/class/
...
```
as input for my _rsync_ backup (I call _rsync_ then directory by directory) - but for _rsync_ trailing slashes (or not) make a huge difference (as I am sure you now).

Thus for now, I have to stick to good old find:
```
find /home/rx -maxdepth 1 -type 
```
output
```
/home/rx/cfg
/home/rx/channel
/home/rx/class
```
Please let me have an option to switch off trailing slashes (or better yet: don't even turn them on by default).
Thanks.
  -A

yyogo:
@reuleaux you should open an issue instead of commenting on the PR, regardless you can use `--exec` which passes the directory names without the trailing separator

`fd -td -x echo`

reuleaux:
OK, thanks for getting back to me (and yes: maybe if I have more time,
and care enough about fd, then I open a separate issue):

For now I can only say: trailing slashes (and no means to turn
them off) render fd unsuitable as a find replacement for me,
and no: -x echo is not the solution: I want full path names
including my home directory:

with gold old find:


***@***.*** ~ $ find /home/rx -maxdepth 1 -type d | sort
/home/rx
/home/rx/bin
/home/rx/.cache
/home/rx/cfg
/home/rx/channel
...

with fd wo/ -x echo


***@***.*** ~ $ fdfind -d 1 -td . /home/rx | sort
/home/rx/bin/
/home/rx/cfg/
/home/rx/channel/
...


with -x echo:
***@***.*** ~ $ fdfind -d 1 -td -x echo . /home/rx | sort
. /home/rx ./bin
. /home/rx ./cfg
. /home/rx ./channel


(which is kind of weird: - a two column output!). Just -x echo maybe:


***@***.*** ~ $ fdfind -d 1 -td -x echo | sort
./bin
./cfg
./channel
./class

but now the /home/rx part is missing.

Sorry: I don't care enough about fd (but you should!):

(Experience in software development suggests:)

before adding a ton of niceties to fd, please make sure
the basic things work.

Thanks,
  A

yyogo:
> OK, thanks for getting back to me (and yes: maybe if I have more time, and care enough about fd, then I open a separate issue): For now I can only say: trailing slashes (and no means to turn them off) render fd unsuitable as a find replacement for me, and no: -x echo is not the solution: I want full path names including my home directory: with gold old find: ***@***.*** ~ $ find /home/rx -maxdepth 1 -type d | sort /home/rx /home/rx/bin /home/rx/.cache /home/rx/cfg /home/rx/channel ... with fd wo/ -x echo ***@***.*** ~ $ fdfind -d 1 -td . /home/rx | sort /home/rx/bin/ /home/rx/cfg/ /home/rx/channel/ ... with -x echo: ***@***.*** ~ $ fdfind -d 1 -td -x echo . /home/rx | sort . /home/rx ./bin . /home/rx ./cfg . /home/rx ./channel (which is kind of weird: - a two column output!). Just -x echo maybe: ***@***.*** ~ $ fdfind -d 1 -td -x echo | sort ./bin ./cfg ./channel ./class but now the /home/rx part is missing. Sorry: I don't care enough about fd (but you should!): (Experience in software development suggests:) before adding a ton of niceties to fd, please make sure the basic things work. Thanks, A Jonathan ***@***.***> writes:
> […](#)
> @reuleaux you should open an issue instead of commenting on the PR, regardless you can use `--exec` which passes the directory names without the trailing separator `fd -td -x echo`

Did you look at the help section of fd? your third example (`fdfind -d 1 -td -x echo . /home/rx | sort`) actually runs `echo . /home/rx {filaname}` for each entry. Also for printing absolute paths there is already a flag (`-a`)

Quoting from the readme:

> `fd` [...] does not aim to support all of find's powerful functionality, it provides sensible (**opinionated**) defaults for a **majority** of use cases.

So, it's OK if `fd` doesn't fit your particular usecase (although in this case I think it does). There are plenty of other tools that can do the job.  That said, If you did care, an issue explaining why `fd` fails for you would be appreciated.

luckman212:
This shoud fork less than calling echo for each result, also preserves sort order
```
fd <whatever> --exec-batch printf '%s\n' {}
```

reuleaux:
Ah, OK, thanks a lot, this works indeed!

vegerot:
This breaks the fd vs. find benchmarks in https://github.com/sharkdp/fd-benchmarks

Issue #794, opened 2021-07-16
--change-newer-than requires using UTC time, "newer" check is incorrect


zachriggle:
# Describe the bug you encountered

It looks like there's differences to `--change-newer-than`' s behavior vs. its documentation, as well as having a potential bug.

## Bug 1: Documentation of UTC time usage

It appears that it needs the time to be stated in UTC, not local time.

```
$ touch a

$ sleep 61

$ touch b

$ ls -l a b
-rw-r--r--  1 zachriggle  staff  12168 Jul 15 19:36 a
-rw-r--r--  1 zachriggle  staff      0 Jul 15 19:37 b

$ date -r b '+%Y-%m-%d %H:%M:%S'
2021-07-15 19:37:01

$ fd --change-newer-than '2021-07-15 19:37:01'
a
b
```

However, if we add the `-u` flag to BSD `date` to get UTC time, we get closer to the expected result.

## Bug 2: Files not actually "newer" are printed

Even if we specify in UTC time, file `b` is still printed.

```
$ date -ur b '+%Y-%m-%d %H:%M:%S'
2021-07-16 00:37:01

$ fd --change-newer-than '2021-07-16 00:37:01'
b
```

This is probably technically correct, since the file was created sometime within that second -- but I consider this to be a bug.

However, if I specify the time in seconds, the comparison should be of the same granularity.

## Bug 3: Time comparison seems to be incorrect

Let's try with fractions of a second.  Luckily, GNU `date` does this (macOS default `date` does not) and is available via homebrew.

```
$ gdate -r b +"%Y-%m%-d %H:%M:%S.%N"
2021-0715 19:37:01.866605000

$ fd --change-newer-than '2021-07-16 00:37:01.866605000'
b
```

So it looks like `fd` supports fractions of a second, but is still flagging file `b` as being "newer".  Let's adjust the time by a tiny fraction of a second.

```
$ fd --change-newer-than '2021-07-16 00:37:01.866605001'
<no output>
```

This is the desired output, but it took a long time to get here.

# Describe what you expected to happen:

No files should show up, since the date was extracted from the newer file's timestamp.

# What version of `fd` are you using?

```
$ fd --version
fd 8.2.1
```

**Which operating system / distribution are you on?**

```
$ uname -srm
Darwin 21.0.0 x86_64
```

tmccombs:
Bug 1 is a duplicate of #631.

[...]

PR #814, opened 2021-08-08
Use local times for time functions

### LINKED ISSUE #631: Time functions assume UTC and don't allow for RFC 3339 offsets
`--change-newer-than` assumes given time is in UTC, not local timezone. Furthermore, offsets like '2020-07-19T00:38:00-07:00' throw errors.

fd version 8.1.1 on Arch Linux

jacobmischka:
This patch uses Chrono for explicit date or datetime parsing, only using
humantime for its relative time parsing. The following formats are accepted:

1. Full RFC3339 parsing, requiring an explicit timezone
2. `YY-MM-DD`, defaulting to time `00:00:00` for the given date in the
   local time zone
3. `YY-MM-DD HH:MM:SS` in the local time zone

Fixes #631, #794

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
 normpath = "0.3"
+chrono = "0.4"
 
 [dependencies.clap]
 version = "2.31.2"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - fd cannot search files under a RAM disk, see #752
 - fd doesn't show substituted drive on Windows, see #365
 - Properly handle write errors to devices that are full, see #737
+- Use local time zone for time functions (`--change-newer-than`, `--change-older-than`), see #631 (@jacobmischka)
 
 ## Changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,9 +31,21 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.42"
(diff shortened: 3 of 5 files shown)


tmccombs (approved):

sharkdp:
Awesome - thank you very much!

The code looks great. I don't have any review comments. It would be good if you could [add an entry to the ChangeLog](https://github.com/sharkdp/fd/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog). I wouldn't necessarily announce this as a breaking change (even if it is), but rather as a bugfix. The fact that we can now also specify the date in different formats could be a second entry in the "Features" section of the ChangeLog.

Is there anything in the documentation that needs updating (not necessarily by you)?


FYI: have you seen #794? Just want to mention it as I'm not sure if the resolution of #794 would be compatible with this change. But I don't really see any reason why it shouldn't.

Cargo.toml
```diff
@@ -48,6 +48,7 @@ globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
 normpath = "0.3"
+chrono = "0.4.19"
```
sharkdp, line 51: Do we need to specify `.19` here or would `0.4` be okay? Explicit versions are set in `Cargo.lock` anyways.

sharkdp (approved):

jacobmischka:
I updated the changelog and docs slightly, including a note to resolve #794 and updated `Cargo.toml`, let me know if there's anything else!

sharkdp (approved):

sharkdp:
No - thank you very much!

PR #817, opened 2021-08-08
Add --no-ignore-parent flag

### LINKED ISSUE #787: Supporting "--no-ignore-parent" like ripgrep, or support ignoring specifit .gitignore?
My problem is that I use my home directory as a git repo to manage my dotfiles, so there is a `~/.gitignore` that ignores everything except some files. The problem is `fd` won't work by default at this situation. Although there is a `--no-ignore-vcs`, but I still want `.gitignore` to be respected when I am in a project directory.

Ripgrep has a flag `--no-ignore-parent` that can ignore `.gitignore` in parent directory, which works well enough. A even better solution could introduce a new flag to exclude certain `.gitignore` only, such as `fd --no-ignore-file ~/.gitignore`.

Similar ripgrep issue: https://github.com/BurntSushi/ripgrep/issues/372

will459:
Closes #787 

Tested on my own machine and abides by any .gitignore rules in the current directory while ignoring any rules provided by the parent.  The flag works for both .gitignore and .ignore files rather than being specific to VCS or non-VCS files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Add new `--no-ignore-parent` flag, see #787 (@will459)
+
 ## Bugfixes
 
 - Set default path separator to `/` in MSYS, see #537 and #730 (@aswild)
--- a/src/app.rs
+++ b/src/app.rs
@@ -50,6 +50,16 @@ pub fn build_app() -> App<'static, 'static> {
                 ),
         )
         .arg(
+            Arg::with_name("no-ignore-parent")
+                .long("no-ignore-parent")
+                .overrides_with("no-ignore-parent")
+                .hidden_short_help(true)
+                .long_help(
+                    "Show search results from files and directories that would otherwise be \
+                        ignored by '.gitignore', '.ignore', or '.fdignore' files in parent directories.",
+                ),
+        )
+        .arg(
             Arg::with_name("no-global-ignore-file")
                 .long("no-global-ignore-file")
                 .hidden(true)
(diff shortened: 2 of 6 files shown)


sharkdp:

This looks good.

Would it be possible to write an integration test for this (see `tests/tests.rs`)? We would probably need two nested "Git repositories" in order for the `.gitignore` files to be active (see *"Pretend that this is a Git repository"* section in `tests/testenv/mod.rs`).

will459:
Apologies for the delay.  I wrote two tests to check the --no-ignore-parent flag.  The reasoning for two is that I observed two different behaviours depending on whether the inner directory was a git repo or not.  When the outter is a repo and the inner is not, then the --no-ignore-parent flag being omitted results in the parent ignore being applied and the --no-ignore-flag being present results in the parent ignore not being applied.  When the inner is also a repo the parent ignore is not applied in either case.

In my testing I found ripgrep exhibits the same behaviour when using their --no-ignore-parent flag which under the hood also works by setting WalkBuilder::parent() from the ignore crate so I think this is likely the expected behaviour.

tavianator (approved):
The `cargo fmt` commit should be folded into the previous one(s).  I can do that if you don't get to it first.

will459:
Done :)

sharkdp:
> In my testing I found ripgrep exhibits the same behaviour when using their --no-ignore-parent flag which under the hood also works by setting WalkBuilder::parent() from the ignore crate so I think this is likely the expected behaviour.

Yes, that is expected. Only actual git repositories  are considered for `.gitignore` matching

Thank you!

PR #819, opened 2021-08-09
Use short options for ls command

### LINKED ISSUE #783: fd -l/--list-details doesn't work in busybox

Try to use `fd -l` in my specific case `fd --extension json --list-details`

```
ls: unrecognized option: human-readable
BusyBox v1.32.1 () multi-call binary.

Usage: ls [-1AaCxdLHRFplinshrSXvctu] [-w WIDTH] [FILE]...

List directory contents

	-1	One column output
	-a	Include entries which start with .
	-A	Like -a, but exclude . and ..
	-x	List by lines
	-d	List directory entries instead of contents
	-L	Follow symlinks
	-H	Follow symlinks on command line
	-R	Recurse
	-p	Append / to dir entries
	-F	Append indicator (one of */=@|) to entries
	-l	Long listing format
	-i	List inode numbers
	-n	List numeric UIDs and GIDs instead of names
	-s	List allocated blocks
	-lc	List ctime
	-lu	List atime
	--full-time	List full date and time
	-h	Human readable sizes (1K 243M 2G)
	--group-directories-first
	-S	Sort by size
	-X	Sort by extension
	-v	Sort by version
	-t	Sort by mtime
	-tc	Sort by ctime
	-tu	Sort by atime
	-r	Reverse sort order
	-w N	Format N columns wide
	--color[={always,never,auto}]	Control coloring
```

The correct long list

Installing fd (8.2.1-r0)

node:lts-alpine3.13
  Executing busybox-1.32.1-r6.trigger

GitHub action

tsoutsman:
Closes #783 

I don't think it's possible to create a regression test as that would require complex CI environments with different `ls` versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - fd doesn't show substituted drive on Windows, see #365
 - Properly handle write errors to devices that are full, see #737
 - Use local time zone for time functions (`--change-newer-than`, `--change-older-than`), see #631 (@jacobmischka)
-- Uses short options for ls regardless of the operating system, see #783
+- Support `--list-details` on more platforms (like BusyBox), see #783
 
 ## Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - fd cannot search files under a RAM disk, see #752
 - fd doesn't show substituted drive on Windows, see #365
 - Properly handle write errors to devices that are full, see #737
+- Uses short options for ls regardless of the operating system, see #783
 
 ## Changes
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,8 +20,8 @@ use anyhow::{anyhow, Context, Result};
 use atty::Stream;
 use globset::GlobBuilder;
 use lscolors::LsColors;
-use regex::bytes::{RegexBuilder, RegexSetBuilder};
 use normpath::PathExt;
+use regex::bytes::{RegexBuilder, RegexSetBuilder};
 
 use crate::error::print_error;
 use crate::exec::CommandTemplate;
@@ -212,9 +212,9 @@ fn run() -> Result<ExitCode> {
         let gnu_ls = |command_name| {
             vec![
                 command_name,
-                "-l",               // long listing format
-                "--human-readable", // human readable file sizes
-                "--directory",      // list directories themselves, not their contents
+                "-l", // long listing format
+                "-h", // human readable file sizes
+                "-d", // list directories themselves, not their contents
                 &color_arg,
             ]
         };
--- a/src/main.rs
+++ b/src/main.rs
@@ -210,6 +210,8 @@ fn run() -> Result<ExitCode> {
 
         #[allow(unused)]
         let gnu_ls = |command_name| {
+            // Note: we use short options here (instead of --long-options) to support more
+            // platforms (like BusyBox).
             vec![
                 command_name,
                 "-l", // long listing format


tmccombs (approved):
Please add an entry to the CHANGELOG.  Other than that I think this looks good.

tsoutsman:
@tmccombs done

tsoutsman:
I'm actually stupid. 5th time's the charm.

tsoutsman:
I just tried to run `ls` on an old MacBook I had lying around. It seems like it doesn't support the `--color` option which outputs `ls: illegal option -- -` the same error found in the logs. None of the changes I made should have affected the colour options, but I made a quick and dirty fix to check if that's the issue. The colour should be enabled later on with the `-G` option anyway.

sharkdp:
> I just tried to run `ls` on an old MacBook I had lying around. It seems like it doesn't support the `--color` option

That's why we previously only passed the color arg if we have `gls` (GNU `ls`) available on macOS. I think it was working correctly.

tsoutsman:
Ight, sorry about that whole debacle, it should finally be good now.

peternewman:
> I don't think it's possible to create a regression test as that would require complex CI environments with different `ls` versions.

Thanks @tsoutsman . As mentioned in the issue, this happened on GitHub Actions, so it should be pretty easy to test. Given it looks like there are test runs I'm a bit surprised they haven't caught this before.

tsoutsman:
> Thanks @tsoutsman . As mentioned in the issue, this happened on GitHub Actions, so it should be pretty easy to test. Given it > looks like there are test runs I'm a bit surprised they haven't caught this before.

Given how I handled changing 4 lines of code, I think I should avoid touching the CI with a ten-foot pole.

peternewman:
Interestingly there is a test:
https://github.com/sharkdp/fd/blob/224b7f2354d2c5c18719ec845fa958e13955679e/tests/tests.rs#L1764-L1770

I'm definitely no rust expert, but even when I add a 20.04 run (#820) (which ubuntu-latest was at the time I hit the issue), there's no mention of any of the tests in that file in the test log that I can see:
https://github.com/peternewman/fd/runs/3279423614?check_suite_focus=true

Which might be why it wasn't caught...

sharkdp:
> I'm definitely no rust expert, but even when I add a 20.04 run (#820) (which ubuntu-latest was at the time I hit the issue), there's no mention of any of the tests in that file in the test log that I can see:
> https://github.com/peternewman/fd/runs/3279423614?check_suite_focus=true
> 
> Which might be why it wasn't caught...

These tests do not run on ARM platforms (see `CARGO_TEST_OPTIONS` in CI scripts). They run on all other platforms though (e.g. https://github.com/peternewman/fd/runs/3279423791?check_suite_focus=true)

sharkdp:
@tsoutsman Thank you very much.

> Given how I handled changing 4 lines of code, I think I should avoid touching the CI with a ten-foot pole.

No worries! :smile: If you (still) think that the code is confusing, please let me know. There's certainly a way to improve it.


I think this PR is good to go. @peternewman ?

sharkdp (approved):

peternewman:
> These tests do not run on ARM platforms (see `CARGO_TEST_OPTIONS` in CI scripts). They run on all other platforms though (e.g. https://github.com/peternewman/fd/runs/3279423791?check_suite_focus=true)

Yes of course, you can't run the binary when cross-compiling...

peternewman:
> I think this PR is good to go. @peternewman ?

Yes, agreed. :shipit:

Looking back at my issue, it was within a Docker container I had the issue, which is probably far more of a faff to spin up CI wise...

PR #820, opened 2021-08-09
Add 20.04 Ubuntu builds

peternewman:
I don't know if you want 18.04 dropped from the CI?


sharkdp:
Yes, I think we can drop the 18.04 builds in case everything works. Thank you!

peternewman:
> Yes, I think we can drop the 18.04 builds in case everything works.

Done throughout, including the minimum rust version one.

PR #821, opened 2021-08-09
Change time comparisons to exclusive by default, add `--time-inclusive`

### LINKED ISSUE #794: --change-newer-than requires using UTC time, "newer" check is incorrect

jacobmischka:
As the names imply, `--change-newer-than` and `--change-before` now are
exclusive (`>` or `<`) by default. A new flag `--time-inclusive`
(aliased to simply `--inclusive`) is added to opt into the previous
inclusive (`>=` or `<=`) behavior.

Fixes #794

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -217,7 +217,9 @@ tebibytes
 .TP
 .BI "\-\-changed-within " date|duration
 Filter results based on the file modification time.
-Files with modification times greater than or equal to the argument will be returned.
+Files with modification times greater the argument will be returned.
+If \-\-time-inclusive is also specified, files with modification times greater than or
+equal to the argument will be returned.
 The argument can be provided as a duration (\fI10h, 1d, 35min\fR) or as a specific point
 in time in either full RFC3339 format with time zone, or as a date or datetime in the
 local time zone (\fIYYYY-MM-DD\fR or \fIYYYY-MM-DD HH:MM:SS\fR).
@@ -230,7 +232,9 @@ Examples:
 .TP
 .BI "\-\-changed-before " date|duration
 Filter results based on the file modification time.
-Files with modification times less than or equal to the argument will be returned.
+Files with modification times less than the argument will be returned.
+If \-\-time-inclusive is also specified, files with modification times less than or
+equal to the argument will be returned.
 The argument can be provided as a duration (\fI10h, 1d, 35min\fR) or as a specific point
 in time in either full RFC3339 format with time zone, or as a date or datetime in the
 local time zone (\fIYYYY-MM-DD\fR or \fIYYYY-MM-DD HH:MM:SS\fR).
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -243,7 +243,7 @@ tebibytes
 .TP
 .BI "\-\-changed-within " date|duration
 Filter results based on the file modification time.
-Files with modification times greater the argument will be returned.
+Files with modification times greater than the argument will be returned.
 If \-\-time-inclusive is also specified, files with modification times greater than or
 equal to the argument will be returned.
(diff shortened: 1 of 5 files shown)


doc/fd.1
```diff
@@ -217,7 +217,9 @@ tebibytes
 .TP
 .BI "\-\-changed-within " date|duration
 Filter results based on the file modification time.
-Files with modification times greater than or equal to the argument will be returned.
+Files with modification times greater the argument will be returned.
```
tmccombs, line 246: I don't think the `than` should be removed from this sentence.

```suggestion
Files with modification times greater than the argument will be returned.
```

tmccombs (changes requested):
I'm also not entirely sure if we need the `--time-inclusive` option.

sharkdp:
I haven't followed the discussion in #794 closely. Can the bug be fixed without introducing a new command-line option? @jacobmischka are you still interested to work on this?

tmccombs:
> Can the bug be fixed without introducing a new command-line option? 

I think so. The new command line option is if we want to keep a way to use an inclusive range, but I'm not really sure if that is needed.

jacobmischka:
Sure, though if we don't want to keep the existing behavior with a flag this PR is probably not worth revising, since all that's left then is to remove two `=` characters. I can do that, if that's the route we want to go.

sharkdp:
> Sure, though if we don't want to keep the existing behavior with a flag this PR is probably not worth revising, since all that's left then is to remove two `=` characters. I can do that, if that's the route we want to go.

I would prefer this. Any help would be very much appreciated.

jacobmischka:
I made a new PR since it was frankly just easier to start from master. #886

PR #822, opened 2021-08-09
Add opposing CLI options

### LINKED ISSUE #595: Opposing/negating command-line options (eg. --no-hidden)
Background
-------------
First off, let me just say that I love your projects, `bat` and `fd` are great and I use them both regularly.

What I was looking for was something similar to the `--no-hidden` option as it exists in ripgrep. As in fd, in ripgrep hidden files are not searched by default. However with ripgrep I can create a global config that effectively changes it's behavior to search hidden files by default, and then still be able to pass a `--no-hidden` option in order to turn it back off on a case by case basis when desired. The logic that parses the command-line options would simply let subsequent opposing command-line options override prior ones. When I was looking around to see if this was possible, I found https://github.com/sharkdp/fd/issues/362 where a global config file that was previously proposed, and I agree that aliases adequately satisfy the same use case.

Proposal
----------
If so-called "opposing" command-line options such as `--no-hidden` were supported. Then it would be possible to define aliases such as

```bash
alias fd='fd --hidden'
```
which would effectively change the default options to search hidden files, and then if you wanted to ignore hidden files you could run something like:

```bash
fd --no-hidden
```

Similarly other "opposing" command-line options could be added where applicable, for example the currently supported `--no-ignore` could be complemented with a new `--ignore` option. 

Just briefly looking at the currently supported options, here are some of the ones I feel could possibly be useful:

Currently Supported Option  | Potential "Opposing" Option                
:---:                                          | :---:                                   
--hidden                                  | --no-hidden    
--no-ignore                             | --ignore
--no-ignore-vcs                     | --ignore-vcs
--follow                                    | --no-follow
--fixed-strings                        | --no-fixed-strings
--show-errors                         | --no-show-errors
--absolute-path                     | --relative-path
--full-path                               | --no-full-path

Note that I haven't tested it, but for some currently supported options, this may or may not already be possible:

Currently Supported Option  | "Opposing" Supported Option       
:---:                                          | :---:                                   
--case-sensitive                     | --ignore-case
--glob foo*                              | --exclude foo*

How do you feel about this? If you would be interested in supporting anything like this in fd that would be awesome. If it helped, I could offer to write an initial implementation. I do have a full-time day job so depending on how involved it ended up being I couldn't guarantee how quickly I could finish it, but I would definitely find some time to do it sooner or later. I would expect that the necessary changes would be mostly, if not fully contained within the command-line parsing logic, but I haven't actually looked into it so you never know.

Asha20:
Closes #595

I've implemented the 5 opposing options mentioned in the issue. Those are:
  - `--no-hidden`, which overrides `--hidden`
  - `--ignore`, which overrides `--no-ignore`
  - `--ignore-vcs`, which overrides `--no-ignore-vcs`
  - `--no-follow`, which overrides `--follow`
  - `--relative-path`, which overrides `--absolute-path`


tmccombs:
Looks like builds are failing

Asha20:
Yes, the error is:

```
error: failed to select a version for `version_check`.
Error:     ... required by package `test-case v1.2.0`
    ... which is depended on by `fd-find v8.2.1 (/project)`
versions that meet the requirements `^0.9.2` are: 0.9.3, 0.9.2

all possible versions conflict with previously selected packages.

  previously selected package `version_check v0.9.1`
    ... which is depended on by `fd-find v8.2.1 (/project)`

failed to select a version for `version_check` which could resolve this conflict
```
...which is puzzling to me since `Cargo.toml` has

```toml
version_check = "0.9"
```

while `Cargo.lock` has

```toml
[[package]]
name = "version_check"
version = "0.9.3"
```

so it should be working fine I think? I'm pretty new to Rust so honestly, I don't know how to resolve this error myself.

tmccombs:
Huh... I can take a closer look later today

src/app.rs
```diff
@@ -25,7 +25,17 @@ pub fn build_app() -> App<'static, 'static> {
...
                          hidden files and directories are skipped). Files and directories are \
-                         considered to be hidden if their name starts with a `.` sign (dot).",
+                         considered to be hidden if their name starts with a `.` sign (dot). \
+                         Flag can be overridden with --no-hidden.",
+                ),
+        )
+        .arg(
+            Arg::with_name("no-hidden")
+                .long("no-hidden")
+                .overrides_with("no-hidden")
```
sharkdp, line 35: Shouldn't it also override `--hidden`? We don't want `--hidden` and `--no-hidden` to be present at the same time, right? This way, you might even be able to simplify the logic in `main.rs` (not completely sure).

tests/tests.rs
```diff
@@ -1788,6 +1789,30 @@ fn test_number_parsing_errors() {
...
+    let out_opposing_flags = te.assert_success_and_get_output(".", &flags);
+
+    assert_eq!(
+        out_no_flags,
+        out_opposing_flags,
+        "{} should override {}",
+        opposing_flags.join(" "),
+        flag
+    );
+}
```
sharkdp, line 1814: Nice!

src/app.rs
```diff
@@ -25,7 +25,17 @@ pub fn build_app() -> App<'static, 'static> {
                 .long_help(
                     "Include hidden directories and files in the search results (default: \
                          hidden files and directories are skipped). Files and directories are \
-                         considered to be hidden if their name starts with a `.` sign (dot).",
+                         considered to be hidden if their name starts with a `.` sign (dot). \
+                         Flag can be overridden with --no-hidden.",
```
sharkdp, line 29: Minor: for the long help text, can we please use full sentences ("The flag can be ...").

sharkdp (commented):
Thank you very much for your contribution. Can we please update the man page as well?

src/app.rs
Asha20: I thought we did want to support both of them being present; at least that's the use case mentioned in the original issue [here](https://github.com/sharkdp/fd/issues/595#issue-619793513) with aliasing.

Asha20:
Sure thing, I'll update it a bit later.

src/app.rs
sharkdp: Ah, yes. Sorry for that. My comment was confusing. What I meant was: I think we can let `clap` help us here. If we set `no-hidden` "`overrides_with`" `hidden`, I think that `clap` internally either has "hidden" or "no-hidden" present, never both.

We definitely want the user to be able to pass both of them.

Asha20: Ah, I see what you mean now. Yep, I think that'll clean up `main.rs` a bit.

sharkdp (approved):
Thank you for the updates!

PR #825, opened 2021-08-10
Add documentation and examples of ** in glob mode

### LINKED ISSUE #764: Double star globbing
Hello, I noticed that the `-g` option doesn't seem to support globbing with `**` like with `git grep` or the `-g` option for `rg`. Could you add something like this?

Examples:
```
git grep -l . "**/dir/file*"
rg --files -g "**/dir/file*"
```

tmccombs:
And how it should be used with the `--full-path` option.

Fixes #764.


sharkdp (approved):
Thank you!

PR #826, opened 2021-08-10
Add better documentation of vcs-ignore files.

### LINKED ISSUE #354: Documentation and differentiation of `--no-ignore-vcs` and `--no-ignore`
The man page says:

```
       -I, --no-ignore
              Do not respect files like .gitignore and .fdignore and include ignored files in the search results.

       --no-ignore-vcs
              Do not respect version control ignore files like .gitignore and include the respective entries in the search results.
```

1. Where is the list of `files like...` that are normally respected?  (The wording makes me wonder if there are other files ending in `ignore` which are usually looked at)
2. Is `-I` a strict superset of `--no-ignore-vcs`?

tmccombs:
Explicitly list which files are used for gitignore.

Fixes #354

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -32,7 +32,7 @@ Include hidden files and directories in the search results
 .TP
 .B \-I, \-\-no\-ignore
 Show search results from files and directories that would otherwise be ignored by
-.IR .gitignore ,
+any of the files ignored by the '\-\-no\-ignore\-vcs' option,
 .IR .ignore ,
 .IR .fdignore ,
 or the global ignore file.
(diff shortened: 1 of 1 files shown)


doc/fd.1
```diff
@@ -32,7 +32,7 @@ Include hidden files and directories in the search results
 .TP
 .B \-I, \-\-no\-ignore
 Show search results from files and directories that would otherwise be ignored by
-.IR .gitignore ,
+any of the files ignored by the '\-\-no\-ignore\-vcs' option,
```
sharkdp, line 35: I think I understand your intention here (avoid duplication), but since the `-no-ignore` option is much more prominent than `--no-ignore-vcs`, I would could maybe phrase this a bit differently. Or maybe make it more clear by showing it in a list (I had to read the sentence twice until I understood it - but that might be just me).

```
Show search results from files and directories that would otherwise be ignored by
  - .gitignore (and related Git ignore files, see --no-ignore-vcs)
  - .ignore
  - .fdignore
  - the global ignore file
```

I think we could also add: "the global fd ignore file"


What do you think?

tmccombs: sure, I can do that.

sharkdp (approved):
Thanks!

PR #833, opened 2021-08-21
Some readability refactorings

niklasmohrin:
Hi everyone, thank you all for your work on fd :)

I recently saw the reddit post calling for contributions, so I sat down and toyed around with the codebase a bit. I found some things that I think would improve readability. The main change here is that I had rust-analyzer extract some methods from the long `run` function, because I really couldn't grasp what was happening in the bigger picture. I also changed the ordering of functions in the `main.rs` file, see the commit message for rationale.

There are two more commits, I also put my rationale into the commit messages there.

Of course, this is a big diff for rather little change. If you don't like any particular change (or all of them), feel free to point that out to me :)

TODO:
- [ ] if reviewers like structural changes I did to `run`, I have to _closely_ read all of the code again to check that nothing got lost in the process / during my rebase (tests all pass, but I would rather be sure)

(somewhat related to #382, although this does not include any of the higher level suggestions)


sharkdp:

This is exactly the type of change I wanted to see from a ticket like #382. Tests currently fail on Rust 1.42. It would be great if you could make it work for 1.42. Otherwise, I'm also fine with increasing the min. supported version.

niklasmohrin:
> I always feel like I need to apologize for the current state of that method smile.

It always happens so fast :D

> Tests currently fail on Rust 1.42.

here is how that went:
- `bool::then` was introduced in 1.50, which is a rather big leap from 1.42. I personally really like `bool::then`, but it is not a must have
- `[T; n]: IntoIterator<Item = T>` is from 1.53, which is even later. This one is especially maddening because now the tests for `merge_exitcodes` need to either pass `[...].iter().copied()` or `vec![...]`. The latter is the better of the two in my opinion (performance is not a priority in tests)

I now replaced everything with 1.42 equivalents, but we should keep these backports in mind for whenever msrv is increased again.

While going through everything again, I noticed that I totally reverted #819 in the original draft. Maybe a regression test should be introduced? Maybe we can have a test that spawns a subprocess and has a mocked `ls` in `$PATH`?

Finally, I moved some things around a bit.

sharkdp:
> While going through everything again, I noticed that I totally reverted #819 in the original draft. Maybe a regression test should be introduced? Maybe we can have a test that spawns a subprocess and has a mocked `ls` in `$PATH`?

That's an option (we do something similar in `bat` with mocked pager executables). However, what would there be to test? If we really want to avoid a regression concerning busybox, we would need to run `fd` within a busybox environment somehow. Maybe that can be done by installing something during the CI run... but to be honest, I'm not sure if it's worth the effort.

src/main.rs
```diff
@@ -50,10 +50,48 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
     let matches = app::build_app().get_matches_from(env::args_os());
 
-    // Set the current working directory of the process
+    set_working_dir(&matches)?;
+    let current_directory = Path::new(".");
+    ensure_current_directory_exists(&current_directory)?;
+    let search_paths = extract_search_paths(&matches, current_directory)?;
+
+    let pattern = extract_search_pattern(&matches)?;
+    check_pattern_is_not_path(&matches, pattern)?;
```
sharkdp, line 75: Maybe also use "ensure_"  here, like above?

sharkdp: Also, since we are concerned with readability here, let's go with: `ensure_that_search_pattern_is_not_a_path`?

src/main.rs
```diff
@@ -50,10 +50,48 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+    let pattern_regex = build_pattern_regex(&matches, pattern)?;
+
+    // The search will be case-sensitive if the command line flag is set or
+    // if the pattern has an uppercase character (smart case).
+    let case_sensitive = !matches.is_present("ignore-case")
+        && (matches.is_present("case-sensitive") || pattern_has_uppercase_char(&pattern_regex));
```
sharkdp, line 81: Why is this not part of `construct_options`?

src/main.rs
```diff
@@ -50,10 +50,48 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+
+    let path_separator = matches
+        .value_of("path-separator")
+        .map_or_else(filesystem::default_path_separator, |s| Some(s.to_owned()));
+    check_path_separator_length(path_separator.as_deref())?;
```
sharkdp, line 86: Same here. Move to `construct_options`?

src/main.rs
```diff
@@ -50,10 +50,48 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+
+    let config = construct_options(matches, case_sensitive, path_separator)?;
```
sharkdp, line 88: I know the problem was there before, but we should really use a consistent naming: config or options.

src/main.rs
```diff
@@ -50,10 +50,48 @@ ow=0:or=0;38;5;16;48;5;203:no=0:ex=1;38;5;203:cd=0;38;5;203;48;5;236:mi=0;38;5;1
...
+    check_leading_dot_misuse(&config, &pattern_regex)?;
```
sharkdp, line 89: Maybe: `ensure_use_hidden_option_for_leading_dot_pattern`?

niklasmohrin:
> That's an option (we do something similar in bat with mocked pager executables). However, what would there be to test? If we really want to avoid a regression concerning busybox, we would need to run fd within a busybox environment somehow. Maybe that can be done by installing something during the CI run... but to be honest, I'm not sure if it's worth the effort.

I thought that one could just test that the program only gets called with short flags, but I agree that it is an uncommon regression. In that case, it is just something to look out for, especially with these big diffs (I only found it because I looked at the commits done to main since I first started working on this PR). Normal PRs (that is: ones that don't just move huge chunks around) should not have this problem

src/walk.rs
```diff
@@ -205,12 +207,8 @@ fn spawn_receiver(
...
 
                 // Wait for all threads to exit before exiting the program.
-                let mut results: Vec<ExitCode> = Vec::new();
-                for h in handles {
-                    results.push(h.join().unwrap());
-                }
-
-                merge_exitcodes(&results)
+                let results = handles.into_iter().map(|h| h.join().unwrap());
+                merge_exitcodes(results)
```
sharkdp, line 211: :+1:

src/walk.rs
```diff
@@ -221,16 +219,16 @@ fn spawn_receiver(
...
-                .unwrap_or_else(|| time::Duration::from_millis(100));
+            let max_buffer_time = config.max_buffer_time.unwrap_or(DEFAULT_MAX_BUFFER_TIME);
 
             let stdout = io::stdout();
             let mut stdout = stdout.lock();
 
-            let mut num_results = 0;
+            let rx_iter = rx
+                .into_iter()
+                .take(config.max_results.unwrap_or(std::usize::MAX));
```
sharkdp, line 229: Concerning readability, I think I would prefer a `if` clause here:
``` rs
if let Some(max_results) = config.max_results {
  rx_iter = rx_iter.take(max_results);
}
```
which would get rid of the `usize::MAX` (even if it's not a real problem).

sharkdp: Just to make sure: are we 100% sure that the new `.take(…)` implementation does the same thing as the `break` before? It's certainly more elegant, but this functionality is quite crucial. With `--max-results=<NUM>` (or `-1`), we definitely want `fd` to exit early, *immediately* after finding the `N`*th* (or first) result.

src/walk.rs
niklasmohrin: Actually, I now noticed that this is semantically different: Since `Iterator::any` stops early once any of the values satisfies the predicate, the later threads won't be joined if one of them returned an error. I assume this should not happen.

To avoid the `collect`, I could add one of two things: A `exhaustive_any` extension to iterators that fully consumes the iterator (so every join call happens), or a `ScopeThread` that is joined once the scope of the handle ends / the handle is dropped. I think the first option would be simpler and easier to read.

Ofc, we could also collect into a vector again, but that would not be very satisfying :laughing:

src/walk.rs
niklasmohrin: The conditional `rx_iter = rx_iter.take(...)` doesn't work because the types are different. Only way around that is to have `rx_iter: &mut dyn Iterator`or similar I think.

The `.take` definetely doesn't wait for the rest of the results and just breaks after the specified number of iterations. reading again, I noticed another small deviation though: Originally this only counts the number of successful results, not the `WorkerResult::Error` cases. While this is probably possible to do with an iterator adapter, it is probably easier to just leave it as it was

src/walk.rs
niklasmohrin: Update: Actually, we can also just guard the iterator to make sure it is fully exhausted. Here is an example:

```rust
struct LoudNumber(i32);
impl Drop for LoudNumber {
    fn drop(&mut self) {
        println!("Dropping {}", self.0);
    }
}
fn main() {
    let mut iter = (0..10).map(LoudNumber);
    let _ = iter.any(|x| x.0 == 5);

    println!("---");
    let mut iter = scopeguard::guard((0..10).map(LoudNumber), |remaining_items| {
        remaining_items.for_each(drop)
    });
    let _ = iter.any(|x| x.0 == 5);
}
```

This outputs:
```console
Dropping 0
Dropping 1
Dropping 2
Dropping 3
Dropping 4
Dropping 5
---
Dropping 0
Dropping 1
Dropping 2
Dropping 3
Dropping 4
Dropping 5
Dropping 6
Dropping 7
Dropping 8
Dropping 9
```

What do you think?

tavianator: Personally I'd prefer the explicit collect over the extra dependency.  `for_each(drop)` is misleading IMO, since it's not like the items wouldn't otherwise be dropped.  This is more straightforward IMO:

```rust
                // Wait for all threads to exit before exiting the program.
                let results: Vec<_> = handles
                    .into_iter()
                    .map(|h| h.join().unwrap())
                    .collect();

                merge_exitcodes(results)
```

niklasmohrin: I agree that `for_each(drop)` to force side effects in `map` is more involved than this code should be and new readers probably have to think about the code rather than just seeing that we just want to join all threads. I saw some discussion about an `Iterator::exhaust(self)` or similar in Rust forums and issues that would be equivalent to `for_each(drop)`, but it got turned down because the latter already exists.

We could do two things here:
1. Don't change the code and collect into vector.
2. Add an extension trait with `.exhausted` or similar that returns an adapter that consumes the remaining iterator on drop. This could increase readability in callers code (and you could hover the method to see what it is supposed to do). I think I will check if this option is feasible and maybe push a suggestion here. (we could also have this in another crate if you like)

Honestly, I would be fine with both approaches though. If we go with the first approach, the change to the `merge_exitcodes` signature doesn't make sense until msrv is up to 1.53 with `[T; N]: IntoIterator`. I would just revert and add a `FIXME:` comment referring to this PR.

sharkdp:
@niklasmohrin I know it's been a while but it would be great if we could get this ready for merging. I kind of don't want to merge other things before this is done. It would be fantastic if you could do a final round of updates (see my other comment) and solve the merge conflict. Thank you!

PR #834, opened 2021-08-22
Minor CI improvements

sharkdp:
(this is a result of a CI sync with bat and hyperfine)

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -150,7 +150,7 @@ jobs:
       run: |
         # test only library unit tests and binary for arm-type targets
         unset CARGO_TEST_OPTIONS
-        unset CARGO_TEST_OPTIONS ; case ${{ matrix.job.target }} in arm-* | aarch64-*) CARGO_TEST_OPTIONS="--lib --bin ${PROJECT_NAME}" ;; esac;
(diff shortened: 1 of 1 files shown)


niklasmohrin:
@sharkdp I was about to open a PR about rustfmt :sweat_smile: 

... and another possible improvement: I recently learned that Clippy has a msrv setting _to prevent warnings we can't fix_. This way, one can use the newest lints / fixes without getting impossible suggestions. Here is the PR from tealdeer that does the switch: https://github.com/dbrgn/tealdeer/pull/200

Should I open one here too?

sharkdp:
> ... and another possible improvement: I recently learned that Clippy has a msrv setting _to prevent warnings we can't fix_. This way, one can use the newest lints / fixes without getting impossible suggestions. Here is the PR from tealdeer that does the switch: [dbrgn/tealdeer#200](https://github.com/dbrgn/tealdeer/pull/200)
> 
> (edit: this also embeds the setting into local clippy runs, not only CI; that means that devs don't get the unfixable warnings anymore)
> 
> Should I open one here too?

That sounds great :+1:. We already run clippy on MSRV, to prevent problems like this, but I guess this way, we could profit from newer lints that we *could* fix on MSRV?

niklasmohrin:
Exactly :+1:

PR #836, opened 2021-08-23
CI: Run clippy on stable and use msrv setting in `clippy.toml`

niklasmohrin:
See #834


.github/workflows/CICD.yml
```diff
@@ -33,11 +32,26 @@ jobs:
       with:
         command: fmt
         args: -- --check
-    - name: Run clippy (on minimum supported rust version to prevent warnings we can't fix)
+    - name: Ensure MSRV is set in `clippy.toml`
+      run: grep "^msrv = \"${{ env.MIN_SUPPORTED_RUST_VERSION }}\"\$" clippy.toml
+    - name: Run clippy
```
sharkdp, line 37: Is there a way to pass `msrv` via the command line to clippy? This way, we could get rid of the chek and we wouldn't need to clutter our root directory.

niklasmohrin: I think we can have it as

```rust
#![clippy::msrv = "1.30.0"]
```

niklasmohrin:
It's been a couple of days since I have been active here :sweat_smile: Please see my edited review comment response (don't know if GH pings everyone on comment edit, thus this message :) )

.github/workflows/CICD.yml
sharkdp: Good point. Thank you.

sharkdp (approved):

PR #847, opened 2021-09-05
Split executable and empty file "types" into seperate "prop" flag

### LINKED ISSUE #823: Make `--type empty` its own option
```
-t, --type <filetype>...              
        Filter the search by type (multiple allowable filetypes can be specified):
          'f' or 'file':         regular files
          'd' or 'directory':    directories
          'l' or 'symlink':      symbolic links
          'x' or 'executable':   executables
          'e' or 'empty':        empty files or directories
          's' or 'socket':       socket
          'p' or 'pipe':         named pipe (FIFO)
```
the help `multiple allowable filetypes can be specified` suggests that adding multiple type parameters will have an `OR` effect (if a file is any of the types then it will be included in the results), but this is not always the case. In particular, the `empty` type is inconsistent. When used with `symlink`, `socket`, or `pipe`, it works in accordance with the help, anything that is `empty` `OR` the other type will be included. However, if you use it with `file`, `executable`, or `directory` it will only include results if they are both `empty` `AND` the other type (e.g. `fd --type empty --type directory` only includes empty directories.

`empty` being a type prevents me from searching for empty symlinks, sockets, or pipes. (I'm not sure if these are possible but I digress)

This issue is touched upon in #714 :
> You are right. executable and empty are special in that they mean: only show executable/empty results. Otherwise --type file --type empty would mean: "show me files OR empty directories/files", which would result in ALL files and empty directories. I don't think that's what most users would expect. See also: #246 (comment)

I think `empty` being made its own option would fix this issue, removing any confusion.

tmccombs:
But keep the old file types for backwards compatibility.

Fixes #823

This is a work in progress. I haven't added anything to the changelog yet, and the documentation could probably be a little better. I'm also not entirely sure if this is something we want to do.

Note that the new `--prop` options behave slightly differently than the old file types. In particular:

- If no `--type` option is given, then `empty` defaults to just empty files, not files and directories. That makes it more consistent with the way `executable` works, seems more useful to me, and is simpler to implement, but I could be convinced to change this to default to files and directories.
- The `executables` property no longer always includes `files`, so you can search for just executable directories. I'm not really sure when that would be useful, but it seems more consistent and logical to me.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -267,18 +267,38 @@ pub fn build_app() -> App<'static, 'static> {
                 ])
                 .hide_possible_values(true)
                 .help(
-                    "Filter by type: file (f), directory (d), symlink (l),\nexecutable (x), \
-                         empty (e), socket (s), pipe (p)",
+                    "Filter by type: file (f), directory (d), symlink (l),\n \
+                         socket (s), pipe (p)",
                 )
                 .long_help(
                     "Filter the search by type (multiple allowable filetypes can be specified):\n  \
                        'f' or 'file':         regular files\n  \
                        'd' or 'directory':    directories\n  \
                        'l' or 'symlink':      symbolic links\n  \
-                       'x' or 'executable':   executables\n  \
-                       'e' or 'empty':        empty files or directories\n  \
                        's' or 'socket':       socket\n  \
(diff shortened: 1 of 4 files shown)


sharkdp:
Closing for now, see #823

PR #853, opened 2021-09-16
Faster colors

### LINKED ISSUE #720: --color=always ~10x slower

I understand that outputting the escape sequences for colors is more text, but I didn't expect it to be 10x slower. Is this expected behavior?

Expected `color=always` to not be 10x slower, although maybe I'm simply misguided.

image

`fd 8.2.1` (`fd-8.2.1-1` on the AUR)

```
Linux 5.10.10-arch1-1 x86_64
```

and `lsb_release` but I don't think it's very helpful :)
```
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```
Ryzen 3900x (24 threads) but it seemed during the bench mark only ~80% cpu was being used.

tavianator:
This can give a significant performance benefit:

    $ hyperfine ./fd-{before,after}" --color=always -HI -j1 --search-path=../../linux >/dev/null"
    Benchmark #1: ./fd-before --color=always -HI -j1 --search-path=../../linux >/dev/null
      Time (mean ± σ):     802.3 ms ±   4.8 ms    [User: 360.0 ms, System: 531.7 ms]
      Range (min … max):   795.8 ms … 808.9 ms    10 runs
 
    Benchmark #2: ./fd-after --color=always -HI -j1 --search-path=../../linux >/dev/null
      Time (mean ± σ):     263.3 ms ±   3.0 ms    [User: 166.4 ms, System: 190.0 ms]
      Range (min … max):   259.0 ms … 268.0 ms    11 runs
 
    Summary
      './fd-after --color=always -HI -j1 --search-path=../../linux >/dev/null' ran
        3.05 ± 0.04 times faster than './fd-before --color=always -HI -j1 --search-path=../../linux >/dev/null'

Fixes #720.  On the other hand, it does lose the distinct colors for 
symlinks along the path, as you mentioned in
https://github.com/sharkdp/fd/issues/720#issuecomment-778680554.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,10 +1,11 @@
+use std::borrow::Cow;
 use std::io::{self, StdoutLock, Write};
 use std::path::Path;
 use std::process;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-use lscolors::{LsColors, Style};
+use lscolors::{Indicator, LsColors, Style};
 
 use crate::config::Config;
 use crate::error::print_error;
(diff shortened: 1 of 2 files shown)


src/output.rs
```diff
@@ -51,29 +45,41 @@ fn print_entry_colorized(
...
+    if offset > 0 {
+        let mut parent_str = Cow::from(&path_str[..offset]);
         if let Some(ref separator) = config.path_separator {
-            *path_string.to_mut() = replace_path_separator(&path_string, separator);
+            *parent_str.to_mut() = replace_path_separator(&parent_str, separator);
         }
-        write!(stdout, "{}", style.paint(path_string))?;
 
-        // TODO: can we move this out of the if-statement? Why do we call it that often?
-        if wants_to_quit.load(Ordering::Relaxed) {
```
tavianator, line 71: I think I have misinterpreted the reason for this code and we need to keep this check

tavianator:
I even have an idea of how to bring the per-component colors back, but it would need changes to `ignore`

src/output.rs
```diff
@@ -53,32 +54,51 @@ fn print_entry_colorized(
...
+
+    if let Some(parent) = path.parent() {
+        offset = parent.to_string_lossy().len();
+        for c in path_str[offset..].chars() {
+            if std::path::is_separator(c) {
+                offset += c.len_utf8();
+            } else {
+                break;
+            }
+        }
```
sharkdp, line 69: I don't understand how this code could be triggered?

tavianator: Say the (user supplied) path is `a/b///c`.  `parent` will be `a/b`.  This loop then advances the length to include `a/b///`.

sharkdp: Ah! Thank you.

sharkdp (approved):

PR #861, opened 2021-10-10
show "./" prefix when search path provided

### LINKED ISSUE #760: fd is vulnerable to filenames with leading dashes
**Describe the bug you encountered:**  When piping relative paths from `fd` to utilities like `xargs`, the program does not return paths, but rather file names.  For instance, in a directory containing `foo.txt`, the command `fd -t f -0 | xargs -0 -I% rm %` passes `foo.txt` to xargs rather than `./foo.txt`.  

If the directory contains files with leading dashes, such as the file `-rf`, commands like `rm` interpret this as options rather than a valid path.  Certain commands have the end-of-options `--` parameter, but not all do, and the burden falls on the user to use `--`.  `find` fixes this problem by always returning absolute or relative paths; returning `./-rf` would not pose a problem for further processing.

Filenames with dashes are not limited to deliberately malicious constructs.  I have found program-generated files in `.config` folders full of leading dashes, as well as media files saved by id, whose id happened to have a leading dash.
 
**Describe what you expected to happen:**  when searching by relative paths, `fd` should return paths, i.e. filenames prefixed by `./` to avoid problems like the above; especially when the output of `fd` is in a pipeline.

**What version of `fd` are you using?** `fd 8.2.1`

Linux 5.11.12-arch1-1 x86_64

LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a

jcaplan:
For example

```
➜  /ws/fd (no-strip/760) ./target/debug/fd mod  
src/exec/mod.rs
src/filter/mod.rs
tests/testenv/mod.rs
➜  /ws/fd (no-strip/760) ./target/debug/fd mod .
./src/exec/mod.rs
./src/filter/mod.rs
./tests/testenv/mod.rs
➜  /ws/fd (no-strip/760) ./target/debug/fd mod --search-path=. 
./src/exec/mod.rs
./src/filter/mod.rs
./tests/testenv/mod.rs
``` 

Show prefix by default for -X/-x, adding `{-}` format string:
```
➜  /ws/fd (no-strip/760) ./target/debug/fd mod -x  echo {}
./tests/testenv/mod.rs
./src/exec/mod.rs
./src/filter/mod.rs
➜  /ws/fd (no-strip/760) ./target/debug/fd mod -x  echo {-}
src/exec/mod.rs
src/filter/mod.rs
tests/testenv/mod.rs
``` 

and for parent:
```
➜  /ws/fd (no-strip/760) ./target/debug/fd mod -x  echo {//} 
./tests/testenv
./src/filter
./src/exec
➜  /ws/fd (no-strip/760) ./target/debug/fd mod -x  echo {//-}
tests/testenv
src/exec
src/filter
```

Also show prefix when output is non interactive, unless `-P` is given:

```
➜  /ws/fd (no-strip/760) ./target/debug/fd mod | xargs -i echo "{}"
./src/exec/mod.rs
./src/filter/mod.rs
./tests/testenv/mod.rs
➜  /ws/fd (no-strip/760) ./target/debug/fd -P mod | xargs -i echo "{}"
src/exec/mod.rs
src/filter/mod.rs
tests/testenv/mod.rs
```

This way users have to *opt in* to less secure current behaviour

Fixes #760


jcaplan:
> IMO You can always opt-in to prepend ./ manually when using -x/-X, automatically prepending ./ in -x/-X makes it impossible to opt-out

agreed. just took a  bit of head scratching to find the way to do it with minimal passing around of config. now specifying the search path works the same way for `-X` and `-x`

sharkdp:
Well but #760 is a security problem. Without the prefix, fd is vulnerable.

We don't want opt-in security. We need security in every situation.

sharkdp:
We already have a `isatty` check in `src/main.rs`. Search for `interactive_terminal`.

jcaplan:
> 
> 
> We already have a `isatty` check in `src/main.rs`. Search for `interactive_terminal`.

yeah I'm already using that

tests/tests.rs
```diff
@@ -1897,7 +1919,28 @@ fn test_error_if_hidden_not_set_and_pattern_starts_with_dot() {
...
     te.assert_failure(&["--glob", ".gitignore"]);
 
-    te.assert_output(&["--hidden", "^\\.gitignore"], ".gitignore");
-    te.assert_output(&["--hidden", "--glob", ".gitignore"], ".gitignore");
+    te.assert_output(&["--hidden", "^\\.gitignore"], "./.gitignore");
+    te.assert_output(&["--hidden", "--glob", ".gitignore"], "./.gitignore");
     te.assert_output(&[".gitignore"], "");
 }
+
+/// Show "./" prefix in results if a search path is provided
```
sharkdp, line 1927: Is this comment outdated?

doc/fd.1
```diff
@@ -405,5 +405,8 @@ $ fd -e py
 .TP
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
+.TP
+.RI "Show ./ prefix in results by specifying search path:"
+$ fd pattern ./
```
sharkdp, line 410: To be honest, I don't the we need this is an example. We should rather add the new `--strip-prefix` option in the normal overview of command line options.

src/config.rs
```diff
@@ -109,4 +109,7 @@ pub struct Config {
 
     /// The maximum number of search results
     pub max_results: Option<usize>,
+
+    /// Whether to prefix each relative path with "./"
+    pub no_strip: bool,
```
sharkdp, line 114: There is a mental overhead in all of the code where I need to negate everything in my head, because this is called "no strip". How about calling this: `strip_cwd_prefix` and reversing all the logic in the code.

src/app.rs
```diff
@@ -615,6 +615,17 @@ pub fn build_app() -> App<'static, 'static> {
                          argument. Changes the usage to `fd [FLAGS/OPTIONS] --search-path <path> \
                          --search-path <path2> [<pattern>]`",
                 ),
+        )
+        .arg(
+            Arg::with_name("strip-prefix")
+                .long("strip-prefix")
+                .short("P")
```
sharkdp, line 622: Can we please remove the short option? I don't think a short prefix for this will be needed frequently, so I'd rather reserve it for future use. We should also add `.hidden_short_help(true)` here, as we don't want this to show up when calling `fd -h` (which only shows the most important options).

src/app.rs
```diff
@@ -615,6 +615,17 @@ pub fn build_app() -> App<'static, 'static> {
                          argument. Changes the usage to `fd [FLAGS/OPTIONS] --search-path <path> \
                          --search-path <path2> [<pattern>]`",
                 ),
+        )
+        .arg(
+            Arg::with_name("strip-prefix")
+                .long("strip-prefix")
```
sharkdp, line 621: How about a rename to `--strip-cwd-prefix`? Maybe that would make it a bit clearer what kind of prefix will be fixed.

src/exec/mod.rs
```diff
@@ -99,6 +99,8 @@ impl CommandTemplate {
                     "{/}" => tokens.push(Token::Basename),
                     "{//}" => tokens.push(Token::Parent),
                     "{/.}" => tokens.push(Token::BasenameNoExt),
+                    "{-}" => tokens.push(Token::StripPrefix),
+                    "{//-}" => tokens.push(Token::ParentStripPrefix),
```
sharkdp, line 103: To be honest, I'd like to remove these new placeholders. I don't think there's a need for this and I don't want users to be overwhelmed when looking at the placeholder syntax.

tests/tests.rs
```diff
@@ -1897,7 +1919,28 @@ fn test_error_if_hidden_not_set_and_pattern_starts_with_dot() {
...
+#[test]
+fn test_no_strip() {
+    let te = TestEnv::new(DEFAULT_DIRS, DEFAULT_FILES);
+
+    te.assert_output(
+        &["-P", "."],
```
sharkdp, line 1933: can we please use the long form here?

sharkdp:
Thank you very much for looking into this and for the updates, @jcaplan. We'd like to include this change in the next stable release. Please let us know if you need help with the remaining open points.

src/exec/mod.rs
tmccombs: That seems reasonable to me. If there really is a need for them we can always add them later.

sharkdp:
Thank you @jcaplan and everyone else involved!

jcaplan:
Sorry dropped the ball there on the last changes. My wife gave birth a few weeks ago. Spare time is currently devoted to sleeping and eating 😅

sharkdp:
Congratulations!! :smile_cat: :smile_cat: :smile_cat: 

You are officially excused :smile:. Just kidding: On my projects, there will never be any time (or other form of) pressure.

PR #863, opened 2021-10-11
walk: Add a cache for DirEntry metadata

tavianator:
This is a reimplementation of https://github.com/sharkdp/fd/pull/402, using `once_cell::unsync::OnceCell` instead of a hand-rolled `Lazy` type.  It gives some performance benefits when multiple filters require metadata:

```
$ hyperfine ./fd-{before,after}" -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null"
Benchmark #1: ./fd-before -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null
  Time (mean ± σ):     449.3 ms ±   5.1 ms    [User: 246.9 ms, System: 354.8 ms]
  Range (min … max):   441.1 ms … 455.1 ms    10 runs
 
Benchmark #2: ./fd-after -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null
  Time (mean ± σ):     392.5 ms ±   3.8 ms    [User: 230.1 ms, System: 287.9 ms]
  Range (min … max):   388.0 ms … 399.3 ms    10 runs
 
Summary
  './fd-after -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null' ran
    1.14 ± 0.02 times faster than './fd-before -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null'
```

without much affect on perf in other cases:

```
$ hyperfine ./fd-{before,after}" -j1 --search-path=../../linux --size=+1k >/dev/null"
Benchmark #1: ./fd-before -j1 --search-path=../../linux --size=+1k >/dev/null
  Time (mean ± σ):     383.8 ms ±   5.6 ms    [User: 212.7 ms, System: 296.1 ms]
  Range (min … max):   377.5 ms … 393.3 ms    10 runs
 
Benchmark #2: ./fd-after -j1 --search-path=../../linux --size=+1k >/dev/null
  Time (mean ± σ):     385.2 ms ±   4.1 ms    [User: 228.4 ms, System: 280.5 ms]
  Range (min … max):   378.3 ms … 393.3 ms    10 runs
 
Summary
  './fd-before -j1 --search-path=../../linux --size=+1k >/dev/null' ran
    1.00 ± 0.02 times faster than './fd-after -j1 --search-path=../../linux --size=+1k >/dev/null'
```

`strace` confirms a reduction in `statx()` calls:

```
$ strace -fc ./fd-before -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
...
 18.32    0.927790           3    254795     24015 statx
...
$ strace -fc ./fd-after -j1 --search-path=../../linux --size=+1k --owner=tavianator >/dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
...
 18.92    0.954145           5    181209     24015 statx
...
```

Together with #853 it would be possible to share the `Metadata` with the colorization which would be an additional benefit, but it's not done in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Upcoming release
 
+## Performance improvements
+
+- File metadata is now cached between the different filters that require it (e.g. `--owner`,
+  `--size`), reducing the number of `stat` syscalls when multiple filters are used; see #863
+
 ## Features
 - Don't buffer command output from `--exec` when using a single thread. See #522
 
@@ -15,6 +20,8 @@
 - Properly handle write errors to devices that are full, see #737
 - Use local time zone for time functions (`--change-newer-than`, `--change-older-than`), see #631 (@jacobmischka)
 - Support `--list-details` on more platforms (like BusyBox), see #783
+- The filters `--owner`, `--size`, and `--changed-{within,before}` now apply to symbolic links
+  themselves, rather than the link target, except when `--follow` is specified; see #863
 
 ## Changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,7 @@ dependencies = [
  "lscolors",
  "normpath",
  "num_cpus",
+ "once_cell",
  "regex",
  "regex-syntax",
  "tempdir",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ anyhow = "1.0"
 dirs-next = "2.0"
 normpath = "0.3"
 chrono = "0.4"
+once_cell = "1.8.0"
 
 [dependencies.clap]
 version = "2.31.3"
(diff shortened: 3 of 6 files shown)


tavianator:
Perhaps I can make it build on Windows after 4 tries?  :P

sharkdp:
I have two remarks:
- Can we please add an entry to `CHANGELOG.md`? For the caching itself and also for the bugfix concerning symlink modification times.
- It might be a good idea to wait with merging this until https://github.com/sharkdp/fd/pull/833 has been merged. What do you guys think?

sharkdp:
Thank you. I merged #833.

PR #877, opened 2021-11-02
Update some dependencies

tmccombs:
Combination of dependabot prs.


sharkdp:
> Combination of dependabot prs.

Do you think the dependabot PRs are useless? (honest opinion please)
I realize they can be a bit annoying and maybe spam our notification inboxes, but I like that we can easily see the changelog for each of the dependencies (I usually read it to see if there are important notes or breaking changes that would not show up as a compilation error) and also the individual build status.

sharkdp (approved):

tavianator:
I actually don't mind the dependabot PRs, but I also like these "rollups."  I don't think it's necessary to keep separate commits around for bumping the version of each dependency.  Not a strong opinion though.

tmccombs:
I don't have a strong opinion either way.

sharkdp:
Ok, thank you. Since the dependabot PRs auto-close, I think the current workflow with both the individual PRs and the "rollup" is also fine.

PR #884, opened 2021-11-14
Refactor exit codes and fix #854

### LINKED ISSUE #854: exit(130) is not the same as raise(SIGINT)
fd handles ^C by propagating `ExitCode::KilledBySigint` up to `main()` which does  `process::exit(exit_code.into())`.  But exit code 130 is a lie told by bash (this comment has a good explanation: https://github.com/rust-lang/rust/issues/62569#issuecomment-774546256).

This is observable from programs that check `WIFEXITED()`/`WIFSIGNALED()`.  Here's a crude example:

```
$ bfs -D exec /dev/null -exec fd . / \; >/dev/null & sleep 1; pkill -INT fd; wait %1
bfs: -D exec: -exec: Executing 'fd' ... [2 arguments]
bfs: warning: Command 'fd' terminated by signal 2 (Interrupt)
$ bfs -D exec /dev/null -exec fd --color=always . / \; >/dev/null & sleep 1; pkill -INT fd; wait %1
bfs: -D exec: -exec: Executing 'fd' ... [3 arguments]
bfs: -D exec: -exec: Command 'fd' failed with status 130
```

The right way to pretend to die by SIGINT is something like `signal(SIGINT, SIG_DFL); raise(SIGINT);`.

tavianator:
- Add an exit() method to ExitCode
- Raise SIGINT when dying by ^C

### DIFF
--- a/src/exit_codes.rs
+++ b/src/exit_codes.rs
@@ -1,5 +1,8 @@
 use std::process;
 
+#[cfg(unix)]
+use nix::sys::signal::{raise, signal, SigHandler, Signal};
+
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum ExitCode {
     Success,
@@ -26,6 +29,16 @@ impl ExitCode {
 
     /// Exit the process with the appropriate code.
     pub fn exit(self) -> ! {
+        #[cfg(unix)]
+        if self == ExitCode::KilledBySigint {
+            // Get rid of the SIGINT handler, if present, and raise SIGINT
+            unsafe {
+                if signal(Signal::SIGINT, SigHandler::SigDfl).is_ok() {
+                    let _ = raise(Signal::SIGINT);
+                }
+            }
+        }
+
         process::exit(self.into())
     }
 }
--- a/src/exit_codes.rs
+++ b/src/exit_codes.rs
@@ -1,3 +1,5 @@
+use std::process;
+
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum ExitCode {
     Success,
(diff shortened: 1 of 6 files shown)


src/walk.rs
```diff
@@ -137,11 +136,9 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> R
     if config.ls_colors.is_some() && config.command.is_none() {
         let wq = Arc::clone(&wants_to_quit);
         ctrlc::set_handler(move || {
-            if wq.load(Ordering::Relaxed) {
+            if wq.fetch_or(true, Ordering::Relaxed) {
```
sharkdp, line 139: :+1:

src/exit_codes.rs
```diff
@@ -21,6 +26,21 @@ impl ExitCode {
...
+    /// Exit the process with the appropriate code.
+    pub fn exit(self) -> ! {
+        #[cfg(unix)]
+        if self == ExitCode::KilledBySigint {
+            // Get rid of the SIGINT handler, if present, and raise SIGINT
+            unsafe {
+                if signal(Signal::SIGINT, SigHandler::SigDfl).is_ok() {
+                    let _ = raise(Signal::SIGINT);
+                }
+            }
```
sharkdp, line 39: Hm, it's unfortunate that `ctrlc` doesn't provide a way to unregister the signal handler yet (see https://github.com/Detegr/rust-ctrlc/issues/28). If unregistering is that easy, do you think this could be upstreamed to `ctrl` somehow?

sharkdp (approved):
Thank you!

src/exit_codes.rs
tavianator: I'll look in to this.  I like the suggested `ctrlc::Channel` API.

PR #885, opened 2021-11-15
Add buffering to stdout when it's not a terminal.

tmccombs:
Follow-up of #736

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,8 +1,7 @@
 use std::borrow::Cow;
-use std::io::{self, StdoutLock, Write};
+use std::io::{self, Write};
 use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
 
 use lscolors::{Indicator, LsColors, Style};
 
@@ -16,11 +15,11 @@ fn replace_path_separator(path: &str, new_path_separator: &str) -> String {
 }
 
 // TODO: this function is performance critical and can probably be optimized
-pub fn print_entry(
-    stdout: &mut StdoutLock,
+pub fn print_entry<W: Write>(
+    stdout: &mut W,
     entry: &Path,
     config: &Config,
-    wants_to_quit: &Arc<AtomicBool>,
+    wants_to_quit: &AtomicBool,
 ) {
     let path = if config.strip_cwd_prefix {
         strip_current_dir(entry)
@@ -46,12 +45,12 @@ pub fn print_entry(
 }
 
 // TODO: this function is performance critical and can probably be optimized
-fn print_entry_colorized(
-    stdout: &mut StdoutLock,
+fn print_entry_colorized<W: Write>(
+    stdout: &mut W,
     path: &Path,
     config: &Config,
     ls_colors: &LsColors,
-    wants_to_quit: &Arc<AtomicBool>,
+    wants_to_quit: &AtomicBool,
 ) -> io::Result<()> {
     // Split the path between the parent and the last component
     let mut offset = 0;
@@ -94,6 +93,7 @@ fn print_entry_colorized(
     }
 
     if wants_to_quit.load(Ordering::Relaxed) {
+        stdout.flush()?;
         ExitCode::KilledBySigint.exit();
     }
 
@@ -101,8 +101,8 @@ fn print_entry_colorized(
 }
 
 // TODO: this function is performance critical and can probably be optimized
-fn print_entry_uncolorized_base(
-    stdout: &mut StdoutLock,
+fn print_entry_uncolorized_base<W: Write>(
+    stdout: &mut W,
     path: &Path,
     config: &Config,
 ) -> io::Result<()> {
@@ -116,8 +116,8 @@ fn print_entry_uncolorized_base(
 }
 
 #[cfg(not(unix))]
-fn print_entry_uncolorized(
-    stdout: &mut StdoutLock,
+fn print_entry_uncolorized<W: Write>(
+    stdout: &mut W,
     path: &Path,
     config: &Config,
 ) -> io::Result<()> {
@@ -125,8 +125,8 @@ fn print_entry_uncolorized(
 }
 
 #[cfg(unix)]
-fn print_entry_uncolorized(
-    stdout: &mut StdoutLock,
+fn print_entry_uncolorized<W: Write>(
+    stdout: &mut W,
     path: &Path,
     config: &Config,
 ) -> io::Result<()> {
--- a/src/output.rs
+++ b/src/output.rs
@@ -93,7 +93,8 @@ fn print_entry_colorized<W: Write>(
     }
 
     if wants_to_quit.load(Ordering::Relaxed) {
-        stdout.flush()?;
+        // Ignore any errors on flush, because we're about to exit anyway
+        let _ = stdout.flush();
         ExitCode::KilledBySigint.exit();
     }
 
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -1,13 +1,13 @@
-use std::borrow::Cow;
 use std::ffi::OsStr;
 use std::fs::{FileType, Metadata};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::mpsc::{sync_channel, Receiver, SyncSender as Sender};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time;
+use std::{borrow::Cow, io::Write};
 
 use anyhow::{anyhow, Result};
 use ignore::overrides::OverrideBuilder;
@@ -21,6 +21,12 @@ use crate::exec;
 use crate::exit_codes::{merge_exitcodes, ExitCode};
 use crate::filesystem;
 use crate::output;
+
+pub const WORKER_CHANNEL_DEFAULT_BOUND: usize = 1000;
+
+pub fn make_worker_channel() -> (Sender<WorkerResult>, Receiver<WorkerResult>) {
+    sync_channel(WORKER_CHANNEL_DEFAULT_BOUND)
+}
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
 enum ReceiverMode {
@@ -53,7 +59,7 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> R
     let first_path_buf = path_iter
         .next()
         .expect("Error: Path vector can not be empty");
-    let (tx, rx) = channel();
+    let (tx, rx) = make_worker_channel();
 
     let mut override_builder = OverrideBuilder::new(first_path_buf.as_path());
 
@@ -220,8 +226,6 @@ fn spawn_receiver(
         } else {
             let start = time::Instant::now();
 
-            let mut buffer = vec![];
-
             // Start in buffering mode
             let mut mode = ReceiverMode::Buffering;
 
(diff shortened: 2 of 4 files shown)


sharkdp:
Thank you for bringing this up to date! It would be good to see some benchmark results (`regression.sh` tests and maybe some benchmarks with output activated for TTY behavior) in order to gain confidence in the changes. Let me know if you need help.

src/output.rs
```diff
@@ -94,15 +93,16 @@ fn print_entry_colorized(
     }
 
     if wants_to_quit.load(Ordering::Relaxed) {
+        stdout.flush()?;
```
tavianator, line 96: If `flush()` fails, we probably still want to exit.  Actually I think it would be good to check `wants_to_quit` before writing anything instead of after. 

We could even do it outside of `print_entry_colorized()` now, but that'd be for another PR.

src/walk.rs
```diff
@@ -220,54 +226,60 @@ fn spawn_receiver(
...
                             ReceiverMode::Streaming => {
-                                output::print_entry(&mut stdout, &value, &config, &wants_to_quit);
+                                output::print_entry(&mut stdout, &path, &config, &wants_to_quit);
+                                if is_interactive && stdout.flush().is_err() {
+                                    // Probably a broken pipe. Exit gracefully.
+                                    return ExitCode::GeneralError;
+                                }
                             }
                         }
-
```
tavianator, line 270: Probably looks better to keep this paragraph break

src/walk.rs
```diff
@@ -282,7 +294,6 @@ fn spawn_receiver(
                     }
                 }
             }
-
```
tavianator, line 285: and this one

src/walk.rs
```diff
@@ -22,6 +22,12 @@ use crate::exit_codes::{merge_exitcodes, ExitCode};
 use crate::filesystem;
 use crate::output;
 
+pub const WORKER_CHANNEL_DEFAULT_BOUND: usize = 1000;
+
+pub fn make_worker_channel() -> (Sender<WorkerResult>, Receiver<WorkerResult>) {
+    sync_channel(WORKER_CHANNEL_DEFAULT_BOUND)
```
tavianator, line 28: Why are we switching to a bounded queue here?

tavianator: Ah, from the original PR https://github.com/sharkdp/fd/pull/736

> I also changed from using a `channel` to a `sync_channel` so the program wont use too much memory if the output device is slow.

tavianator:
By the way, you could add

```
Co-authored-by: sourlemon207 <jw1756@protonmail.com>
```

to the commit message to attribute it to both you and the original author.

tmccombs:
I tried running the the rust-lang repository and got:

```
## `fd` regression benchmark

### No pattern

|:---|---:|---:|---:|---:|
| `./fd-master --hidden --no-ignore '' '/home/thayne/bulk-home/devel/rust-lang'` | 245.2 ± 2.7 | 241.6 | 249.7 | 1.00 |
| `./fd-feature --hidden --no-ignore '' '/home/thayne/bulk-home/devel/rust-lang'` | 269.5 ± 4.4 | 263.3 | 278.1 | 1.10 ± 0.02 |

### Simple pattern

|:---|---:|---:|---:|---:|
| `./fd-master '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 343.1 ± 4.6 | 338.4 | 355.0 | 1.00 ± 0.01 |
| `./fd-feature '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 342.6 ± 1.8 | 340.6 | 346.4 | 1.00 |

### Simple pattern (-HI)

|:---|---:|---:|---:|---:|
| `./fd-master -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 242.9 ± 3.7 | 238.7 | 251.2 | 1.00 |
| `./fd-feature -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 248.9 ± 5.3 | 242.5 | 258.1 | 1.02 ± 0.03 |

### File extension

|:---|---:|---:|---:|---:|
| `./fd-master -HI --extension jpg '' '/home/thayne/bulk-home/devel/rust-lang'` | 250.9 ± 4.4 | 246.0 | 258.4 | 1.00 |
| `./fd-feature -HI --extension jpg '' '/home/thayne/bulk-home/devel/rust-lang'` | 253.3 ± 2.9 | 249.3 | 258.6 | 1.01 ± 0.02 |

### File type

|:---|---:|---:|---:|---:|
| `./fd-master -HI --type l '' '/home/thayne/bulk-home/devel/rust-lang'` | 245.2 ± 4.6 | 240.8 | 255.7 | 1.00 |
| `./fd-feature -HI --type l '' '/home/thayne/bulk-home/devel/rust-lang'` | 251.1 ± 4.9 | 246.5 | 261.0 | 1.02 ± 0.03 |

### Cold cache

|:---|---:|---:|---:|---:|
| `./fd-master -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 28.951 ± 0.725 | 28.224 | 29.674 | 1.01 ± 0.03 |
| `./fd-feature -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 28.654 ± 0.318 | 28.339 | 28.975 | 1.00 |

```

interestingly, it looks like this branch is actually a little bit slower, I'm not really sure why.

tmccombs:
I switch back to using `channel` instead of `sync_channel` than it is, at least not any slower:

```
## `fd` regression benchmark

### No pattern

|:---|---:|---:|---:|---:|
| `./fd-master --hidden --no-ignore '' '/home/thayne/bulk-home/devel/rust-lang'` | 250.0 ± 2.4 | 245.9 | 253.7 | 1.00 |
| `./fd-feature --hidden --no-ignore '' '/home/thayne/bulk-home/devel/rust-lang'` | 254.4 ± 1.4 | 251.7 | 255.9 | 1.02 ± 0.01 |

### Simple pattern

|:---|---:|---:|---:|---:|
| `./fd-master '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 342.2 ± 3.1 | 336.3 | 347.9 | 1.00 |
| `./fd-feature '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 342.7 ± 2.7 | 340.3 | 349.0 | 1.00 ± 0.01 |

### Simple pattern (-HI)

|:---|---:|---:|---:|---:|
| `./fd-master -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 242.6 ± 3.1 | 239.2 | 249.0 | 1.00 |
| `./fd-feature -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 244.9 ± 3.8 | 239.2 | 253.7 | 1.01 ± 0.02 |

### File extension

|:---|---:|---:|---:|---:|
| `./fd-master -HI --extension jpg '' '/home/thayne/bulk-home/devel/rust-lang'` | 253.8 ± 3.0 | 249.2 | 259.3 | 1.00 |
| `./fd-feature -HI --extension jpg '' '/home/thayne/bulk-home/devel/rust-lang'` | 254.7 ± 3.2 | 250.6 | 260.4 | 1.00 ± 0.02 |

### File type

|:---|---:|---:|---:|---:|
| `./fd-master -HI --type l '' '/home/thayne/bulk-home/devel/rust-lang'` | 243.4 ± 4.5 | 238.9 | 255.0 | 1.00 |
| `./fd-feature -HI --type l '' '/home/thayne/bulk-home/devel/rust-lang'` | 244.8 ± 3.0 | 240.9 | 249.1 | 1.01 ± 0.02 |

### Cold cache

|:---|---:|---:|---:|---:|
| `./fd-master -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 29.176 ± 0.388 | 28.729 | 29.432 | 1.00 ± 0.02 |
| `./fd-feature -HI '.*[0-9]\.jpg$' '/home/thayne/bulk-home/devel/rust-lang'` | 29.120 ± 0.375 | 28.754 | 29.504 | 1.00 |

```

sharkdp:
Ok, thank you. And do we have a significant improvement for searches that need to write a lot of results to stdout? Something like `fd -HI`?

sharkdp:
Have you played around with different buffer sizes? https://doc.rust-lang.org/stable/std/io/struct.BufWriter.html#method.with_capacity

tmccombs:
> Have you played around with different buffer sizes?
 
I have not

I'm wondering if it might be that the bottleneck, at least on my system,  is reading from disk, so even if it writes to stdout faster, it doesn't impact the overall run time. 

So I tried changing it to write the output to a file, instead of whatever hyperfine does with the output by default, and got much better results:

```
❯ hyperfine --warmup 5 './fd-master -HI "" ~/bulk-home/devel/rust-lang/ > test1' './fd-feature -HI "" ~/bulk-home/devel/rust-lang/ > test2' --shell bash                                                at 00:14:26 
Benchmark 1: ./fd-master -HI "" ~/bulk-home/devel/rust-lang/ > test1
  Time (mean ± σ):     559.2 ms ±  91.6 ms    [User: 813.3 ms, System: 878.3 ms]
  Range (min … max):   511.7 ms … 817.3 ms    10 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: ./fd-feature -HI "" ~/bulk-home/devel/rust-lang/ > test2
  Time (mean ± σ):     326.5 ms ±  17.4 ms    [User: 781.8 ms, System: 681.6 ms]
  Range (min … max):   301.9 ms … 347.3 ms    10 runs
 
Summary
  './fd-feature -HI "" ~/bulk-home/devel/rust-lang/ > test2' ran
    1.71 ± 0.30 times faster than './fd-master -HI "" ~/bulk-home/devel/rust-lang/ > test1'
```

sharkdp:
> So I tried changing it to write the output to a file, instead of whatever hyperfine does with the output by default, and got much better results:

Using hyperfine, it should write to `/dev/null` by default (related: https://github.com/sharkdp/hyperfine/issues/377). I would have naively assumed that this wouldn't change performance!?

sharkdp:
I ran some benchmarks myself. I see quite significant improvements when a lot of results (1.6 million) are printed:
```
Benchmark 1: ./fd-master --hidden --no-ignore '' '/home/shark'
  Time (mean ± σ):      1.392 s ±  0.025 s    [User: 4.129 s, System: 4.597 s]
  Range (min … max):    1.372 s …  1.454 s    10 runs
 
Benchmark 2: ./fd-feature --hidden --no-ignore '' '/home/shark'
  Time (mean ± σ):      1.094 s ±  0.037 s    [User: 3.769 s, System: 4.273 s]
  Range (min … max):    1.072 s …  1.193 s    10 runs
 
Summary
  './fd-feature --hidden --no-ignore '' '/home/shark'' ran
    1.27 ± 0.05 times faster than './fd-master --hidden --no-ignore '' '/home/shark''
```

In combination with the changes by @tavianator, this will make colorized searches almost 5x faster (compared to 8.2.1) :partying_face: Very cool!
```
Benchmark 1: ./fd-8.2.1 --color=always '' '/home/shark'
  Time (mean ± σ):      2.736 s ±  0.034 s    [User: 2.001 s, System: 2.149 s]
  Range (min … max):    2.706 s …  2.820 s    10 runs
 
Benchmark 2: ./fd-feature --color=always '' '/home/shark'
  Time (mean ± σ):     595.9 ms ±   5.8 ms    [User: 1085.6 ms, System: 916.4 ms]
  Range (min … max):   584.9 ms … 606.8 ms    10 runs
 
Summary
  './fd-feature --color=always '' '/home/shark'' ran
    4.59 ± 0.07 times faster than './fd-8.2.1 --color=always '' '/home/shark''
```

sharkdp (approved):

tavianator:
A `write()` to `/dev/null` [doesn't even look at the data](https://github.com/torvalds/linux/blob/1bff7d7e8c487b9b0ceab70b43b781f1d45f55eb/drivers/char/mem.c#L456), so reducing the number of `write()`s just reduces syscall overhead which may be hidden by parallelism anyway.

sharkdp:
> A `write()` to `/dev/null` [doesn't even look at the data](https://github.com/torvalds/linux/blob/1bff7d7e8c487b9b0ceab70b43b781f1d45f55eb/drivers/char/mem.c#L453-L457), so reducing the number of `write()`s just reduces syscall overhead which may be hidden by parallelism anyway.

Interesting. I was secretly hoping that you would explain what's going on :smile: - thanks!

So I guess this means that we would get slightly more realistic benchmarking results if we pipe the output to a file?

sharkdp:
> > Have you played around with different buffer sizes?
> 
> I have not

I tried out various buffer sizes and it doesn't make a huge difference as long as the buffer size is larger than a few 100 B. The default is 8 KiB which seems reasonable.

sharkdp:
> So I guess this means that we would get slightly more realistic benchmarking results if we pipe the output to a file?

If the output goes to an actual file (on `tmpfs`), the advantage is even greater:
```
Benchmark 1: ./fd-master --hidden --no-ignore '' '/home/shark' > /tmp/test.out
  Time (mean ± σ):      1.863 s ±  0.005 s    [User: 4.058 s, System: 5.137 s]
  Range (min … max):    1.857 s …  1.875 s    10 runs
 
Benchmark 2: fd --hidden --no-ignore '' '/home/shark' > /tmp/test.out
  Time (mean ± σ):      1.087 s ±  0.005 s    [User: 3.716 s, System: 4.357 s]
  Range (min … max):    1.078 s …  1.098 s    10 runs
 
Summary
  'fd --hidden --no-ignore '' '/home/shark' > /tmp/test.out' ran
    1.71 ± 0.01 times faster than './fd-master --hidden --no-ignore '' '/home/shark' > /tmp/test.out'
```

Even more so, if we are writing to an SSD:
```
Benchmark 1: ./fd-master --hidden --no-ignore '' '/home/shark' > /home/shark/test.out
  Time (mean ± σ):      3.376 s ±  0.064 s    [User: 4.138 s, System: 6.629 s]
  Range (min … max):    3.318 s …  3.525 s    10 runs
 
Benchmark 2: fd --hidden --no-ignore '' '/home/shark' > /home/shark/test.out
  Time (mean ± σ):      1.480 s ±  0.299 s    [User: 3.748 s, System: 4.469 s]
  Range (min … max):    1.184 s …  2.137 s    10 runs
 
Summary
  'fd --hidden --no-ignore '' '/home/shark' > /home/shark/test.out' ran
    2.28 ± 0.46 times faster than './fd-master --hidden --no-ignore '' '/home/shark' > /home/shark/test.out'
```

sharkdp:
I see no reason to delay merging this.

tavianator:
> So I guess this means that we would get slightly more realistic benchmarking results if we pipe the output to a file?

Depends what you're trying to be realistic about :).  Writing to `/dev/null` is a pretty good proxy for the "maximum possible" throughput of a program, which is a reasonable thing to benchmark.

You might also try writing to a file, but that complicates things.  Depending on how much you write, the data might only hit the page cache, or the flash cache of a hard drive, etc.  The filesystem, hardware, etc. will all affect the performance and make it harder to interpret the results.

The right thing to do depends on what you actually want to measure.  I think mostly people care about interactive use, where writing to the TTY is probably the biggest bottleneck.  The current benchmarks don't really capture this well.

Another common use is probably piping `fd` to some other program.  For this case it might be better to do something like `fd | cat >/dev/null` instead of `fd >/dev/null` so that the writes actually go through a pipe.

tavianator (approved):

sharkdp:
I see no reason to delay merging this. Thank you @sourlemon207 @tmccombs @tavianator.

I'll create a ticket to address the shortcomings of the current benchmark set.

PR #891, opened 2021-11-20
Add an example of the `-X` option using `rg`

### LINKED ISSUE #862: Combining fd and rg
fd 8.2.1 on Windows 10

`fd --changed-within "2021-10-09" | rg -i xxx`
won't return an result.

`fd --changed-within "2021-10-09" --exec  rg -i xxx`
returns the lines macthing xxx but the output is missing the filename where xxx is found.
Any idea how to get the filename where xxx is found?

`rg -i xxx` 
alone returns machting lines including filename

tacoda:
Closes #862

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,9 +194,9 @@ fd … -X ls -lhd --color=always
 This pattern is so useful that `fd` provides a shortcut. You can use the `-l`/`--list-details`
 option to execute `ls` in this way: `fd … -l`.
 
-Another example of the `-X` option can be shown with `rg`, using all the search results as arguments:
+The `-X` option is also useful when combining `fd` with [ripgrep](https://github.com/BurntSushi/ripgrep/) (`rg`) in order to search within a certain class of files, like all C++ source files:
 ```bash
-fd … -X rg -i target
+fd -e cpp -e cxx -e h -e hpp -X rg 'std::cout'
 ```
 
 Convert all `*.jpg` files to `*.png` files:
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

README.md
```diff
@@ -194,6 +194,11 @@ fd … -X ls -lhd --color=always
 This pattern is so useful that `fd` provides a shortcut. You can use the `-l`/`--list-details`
 option to execute `ls` in this way: `fd … -l`.
 
+Another example of the `-X` option can be shown with `rg`, using all the search results as arguments:
+```bash
+fd … -X rg -i target
```
sharkdp, line 199: Maybe we could try to find a realistic example here to make it easier to understand what's going on? I would leave out the `-i` option for ripgrep. `target` also doesn't seem like a "typical" search pattern for me. Maybe something like:
```bash
fd -e cpp -e cxx -e h -e hpp -X rg 'std::cout'
```

README.md
```diff
@@ -194,6 +194,11 @@ fd … -X ls -lhd --color=always
 This pattern is so useful that `fd` provides a shortcut. You can use the `-l`/`--list-details`
 option to execute `ls` in this way: `fd … -l`.
 
+Another example of the `-X` option can be shown with `rg`, using all the search results as arguments:
```
sharkdp, line 197:
```suggestion
The `-X` option is also useful when combining `fd` with [ripgrep](https://github.com/BurntSushi/ripgrep/) (`rg`) in order to search within a certain class of files, like all C++ source files:
```

sharkdp:

PR #892, opened 2021-11-25
Remove `lazy_static` from dependencies

sorairolake:
This is because `once_cell` already exists on the dependencies.


sharkdp:

PR #895, opened 2021-11-26
walk: Encapsulate the buffering behavior in a struct

### LINKED ISSUE #868: fd waits for the whole traversal if the only matches arrive within max_buffer_time
If all the matches are received very quickly, within `max_buffer_time`, then they'll be added to the buffer without being printed.  The first match to arrive after the time limit will kick it into streaming mode, but if we never get any more matches then that never happens and we'll wait for the whole search to complete before printing anything.

I guess the fix is to use `rx.recv_timeout()` instead of `for worker_result in rx` [here](https://github.com/sharkdp/fd/blob/17dd2a6dfec1ca2086c3d279f5c458758bc9a71b/src/walk.rs#L239).

tavianator:
The new ReceiverBuffer struct allows us to factor out the receiver
implementation into a number of helper methods.  The new implementation
uses rx.{recv,recv_timeout} instead of a for loop, which enables us to
switch to streaming mode at the right time without waiting for more
results.

Fixes #868.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -1,12 +1,13 @@
 use std::ffi::OsStr;
 use std::fs::{FileType, Metadata};
 use std::io;
+use std::mem;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::mpsc::{channel, Receiver, RecvTimeoutError, Sender};
 use std::sync::{Arc, Mutex};
 use std::thread;
-use std::time;
+use std::time::{Duration, Instant};
 use std::{borrow::Cow, io::Write};
 
 use anyhow::{anyhow, Result};
@@ -23,6 +24,7 @@ use crate::filesystem;
 use crate::output;
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
+#[derive(PartialEq)]
 enum ReceiverMode {
     /// Receiver is still buffering in order to sort the results, if the search finishes fast
     /// enough.
@@ -41,7 +43,7 @@ pub enum WorkerResult {
 /// Maximum size of the output buffer before flushing results to the console
 pub const MAX_BUFFER_LENGTH: usize = 1000;
 /// Default duration until output buffering switches to streaming.
-pub const DEFAULT_MAX_BUFFER_TIME: time::Duration = time::Duration::from_millis(100);
+pub const DEFAULT_MAX_BUFFER_TIME: Duration = Duration::from_millis(100);
 
 /// Recursively scan the given search path for files / pathnames matching the pattern.
 ///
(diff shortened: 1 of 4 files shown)


src/walk.rs
```diff
@@ -160,6 +162,161 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> R
...
+    /// Receive the next worker result.
+    fn recv(&self) -> Result<WorkerResult, RecvTimeoutError> {
+        match self.mode {
+            ReceiverMode::Buffering => {
+                // Wait at most until we should switch to streaming
+                let now = Instant::now();
+                self.deadline
+                    .checked_duration_since(now)
+                    .ok_or(RecvTimeoutError::Timeout)
+                    .and_then(|t| self.rx.recv_timeout(t))
```
tmccombs, line 226: If/once [`recv_deadline`](https://doc.rust-lang.org/stable/std/sync/mpsc/struct.Receiver.html#method.recv_deadline) (Tracked by https://github.com/rust-lang/rust/issues/46316) is stabilized, that would probably be a better fit here.

src/walk.rs
```diff
@@ -160,6 +162,161 @@ pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> R
...
+            }
+            ReceiverMode::Streaming => {
+                // Wait however long it takes for a result
+                Ok(self.rx.recv()?)
+            }
+        }
+    }
+
+    /// Wait for a result or state change.
+    fn poll(&mut self) -> Result<(), ExitCode> {
```
tmccombs, line 236: It's a little weird that this function is `poll` but it isn't actually part of a Future. I don't have a better name though. It's just a little unfortunate that it happens to share a name with the most important function of a `Future`.

tmccombs, line 236:
```suggestion
    fn poll(&mut self) -> ControlFlow<ExitCode> {
```

Would probably be a better return type here, but that would require increasing the MSRV to 1.55.

tmccombs (approved):

tmccombs:
Have you done any benchmarks on this?

src/walk.rs
tavianator: Yeah for sure.  (Do you know what the process for requesting stabilization is?)

Funnily enough `recv_timeout()` is [implemented in terms of `recv_deadline()`](https://github.com/rust-lang/rust/blob/68b554e6af18726fe6fa8de2134c59c441e0b019/library/std/src/sync/mpsc/mod.rs#L1276) so here I'm emulating `recv_deadline()` with `recv_timeout()` which is emulated by `recv_deadline()`.

src/walk.rs
tavianator: I didn't realize `ControlFlow` was stabilized.  I'll leave it up to @sharkdp whether that MSRV bump would be acceptable.

tavianator:
@tmccombs Benchmarks seem totally neutral

src/walk.rs
tmccombs: Probably not worth bumping MSRV just for this, but may be worth changing when we do bump it.

src/walk.rs
tmccombs: I don't know exactly what the process for that is. From the tracking issue it looks like the biggest open question is the name of the function.

tavianator: Huh well here's the news on that: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Stabilizing.20.60recv_deadline.28.29.60

Kinda amazed that https://github.com/rust-lang/rust/issues/39364 is still open.  I guess we should switch to `crossbeam-channels` before merging this PR.

tavianator: Switched to `crossbeam-channel`

tmccombs (approved):

sharkdp:
Thank you!

I'm always usually fine with bumping MSRV, because `fd` is not a library. And it feels like most people use `rustup` with a recent version of Rust anyways (instead of distribution-package).

sharkdp:
This PR (specifically a4bb734482c7bf3fd4d30081e7d3ce7786a3775e) introduces a pretty severe performance regression (20-30%) for me. For example:

```
Simple pattern

Benchmark 1: ./fd-8.3.0 '.*[0-9]\.jpg$' '/home/shark/Informatik/'
  Time (mean ± σ):     179.6 ms ±   1.7 ms    [User: 685.0 ms, System: 653.2 ms]
  Range (min … max):   176.5 ms … 181.6 ms    16 runs
 
Benchmark 2: ./fd-a4bb734482c7b '.*[0-9]\.jpg$' '/home/shark/Informatik/'
  Time (mean ± σ):     229.4 ms ±  14.8 ms    [User: 788.6 ms, System: 930.8 ms]
  Range (min … max):   206.9 ms … 253.0 ms    11 runs
 
Summary
  './fd-8.3.0 '.*[0-9]\.jpg$' '/home/shark/Informatik/'' ran
    1.28 ± 0.08 times faster than './fd-a4bb734482c7b '.*[0-9]\.jpg$' '/home/shark/Informatik/''
```

I found this with `git bisect`, i.e. none of the other commits before showed that kind of slowdown. I would appreciate if we could roll this back.

@tmccombs: I think you have also tried crossbeam-channel in the past? I vaguely remember that I tried it as well at some point and also saw a performance regression.

tavianator:
@sharkdp Interesting, I did re-do the benchmarks after switching to crossbeam and found there was still no measurable difference:

```
Benchmark 1: ./fd-std . ~/code/linux
  Time (mean ± σ):     126.4 ms ±   1.9 ms    [User: 913.5 ms, System: 3832.9 ms]
  Range (min … max):   121.4 ms … 129.4 ms    23 runs
 
Benchmark 2: ./fd-crossbeam . ~/code/linux
  Time (mean ± σ):     126.3 ms ±   1.5 ms    [User: 971.0 ms, System: 3819.8 ms]
  Range (min … max):   123.7 ms … 128.8 ms    23 runs
 
Summary
  './fd-crossbeam . ~/code/linux' ran
    1.00 ± 0.02 times faster than './fd-std . ~/code/linux'
```

The pattern doesn't seem to affect it, whether there's many matches or few.

I wonder if there is a platform-specific thing going on, either OS or hardware.  What platform(s) do you see the regression on?  My tests were on a Threadripper 3960X running Linux.

tavianator:
I also see no perf regression on an i7-6700HQ.  So it's not that crossbeam is only fast on new hardware :)

tavianator:
Anyway I can prepare a revert until we understand what's going on so it doesn't block a new release

tmccombs:
I never did benchmarks for crossbeam. I'm actually a little surprised if that is it, since crossbeam is supposed to be faster than std mpsc (although that might depend on how it is used)

sharkdp:
> I wonder if there is a platform-specific thing going on, either OS or hardware. What platform(s) do you see the regression on? My tests were on a Threadripper 3960X running Linux.

Hmm. Strange. Im also on Linux, Intel Core i7-4700MQ. Maybe it depends on the target folder instead? I will try to do some more comprehensive benchmarks.

Thank you for opening the revert PR.

tavianator:
@sharkdp I'd be curious to compare `cargo flamegraph` output between our machines

sharkdp:
Ok. Let's continue the discussion in #933.

PR #902, opened 2021-12-05
Quit senders more aggressively

### LINKED ISSUE #867: --max-results=1 does not actually quit after the first result

I am running `fd` on a slow FUSE filesystem ([goofys](https://github.com/kahing/goofys)). It seems like `-1`/`--max-results=1` doesn't do what it says on the tin.

I expected `fd` to print the result and then quit. There is only one file on the path that matches my pattern.

I ran this command: `fd -F --max-results=1 PATTERN PATH`

One result is printed ~5 seconds later, but `fd` continues to run for a long time, at least a minute, until I interrupted it. FWIW, `find` can search the entire path in <25 seconds.

```
fd 8.2.1
```

```
Linux 5.8.0-63-generic x86_64
Distributor ID:	Ubuntu
Description:	Ubuntu 20.10
Release:	20.10
Codename:	groovy
```

tavianator:
Fixes #867

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,7 +1,6 @@
 use std::borrow::Cow;
 use std::io::{self, Write};
 use std::path::Path;
-use std::sync::atomic::{AtomicBool, Ordering};
 
 use lscolors::{Indicator, LsColors, Style};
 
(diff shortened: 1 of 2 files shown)


tmccombs (approved):

sharkdp:
Thank you very much!

PR #903, opened 2021-12-07
A couple of minor recommendations from the Code Quality action

tmccombs:
use derive for a Default impl, and use a function directly instead of
needlessly wrapping it in a closure


sharkdp (approved):
Thanks

PR #904, opened 2021-12-14
fix: disable jemalloc on risc-v target

rapiz1:
`jemallocator` does not support risc-v and fails the compilation.
This fixes build for risc-v.

Ping @xctan as the original patch author

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ libc = "0.2"
 # FIXME: Re-enable jemalloc on macOS
 # jemalloc is currently disabled on macOS due to a bug in jemalloc in combination with macOS
 # Catalina. See https://github.com/sharkdp/fd/issues/498 for details.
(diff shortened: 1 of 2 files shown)


xctan:
The crate `jemallocator` seems to be outdated while `jemalloc` has support for risc-v. It would seem that adding support for risc-v in `jemallocator` is more reasonable, so this is just a temporary fix.

felixonmars:
Another alternative is tikv-jemallocator, which is used by rustc itself now: https://github.com/meilisearch/MeiliSearch/pull/1692

rapiz1:
@tavianator Do you have any comments on this PR? Should it be replaced by `tikv-jemallocator`? Let's see what we can do to push the PR forward and it will be great to have `fd` build on risc-v

sharkdp:
If we can show that:
* this does not lead to any performance regressions
* fixes the problems on risc-v
* still builds on all other supported platforms
…I'm fine with moving to `tikv-jemallocator`

PR #906, opened 2021-12-20
add filters for creation and access time

wfraser:
This generalizes the `TimeFilter` type to operate on different time attributes, and adds support for creation and access times, in addition to the modification time already supported.

Creation time is especially nice because it's not supported by GNU Find on Linux -- the option is there, but only works on BSD systems. This implementation does, and will work on whatever systems and filesystems the [`std::fs::Metadata::created()`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.created) function supports. :)

### DIFF
--- a/src/filter/time.rs
+++ b/src/filter/time.rs
@@ -2,16 +2,19 @@ use chrono::{offset::TimeZone, DateTime, Local, NaiveDate};
 
 use std::time::SystemTime;
 
-/// Filter based on time ranges.
 #[derive(Debug, PartialEq)]
-pub enum TimeFilter {
+pub enum TimeRange {
     Before(SystemTime),
     After(SystemTime),
 }
 
-impl TimeFilter {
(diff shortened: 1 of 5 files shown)


src/filter/time.rs
```diff
@@ -26,21 +29,66 @@ impl TimeFilter {
                     })
                     .or_else(|| Local.datetime_from_str(s, "%F %T").ok())
                     .map(|dt| dt.into())
-            })
+            })?;
+        Some(variant(time))
```
tmccombs, line 33:
```suggestion
            }).and_then(variant)
```

tmccombs (changes requested):
What if any is the impact on performance with `--changed-within` and `--changed-before`? Can you run some benchmarks comparing this to master with searches using those options (using hyperfine). 

Also, before this is merged you'll need to add an entry to the CHANGELOG and update the _fd zsh completions and the man page. Although maybe we should wait on the benchmarks for that.

sharkdp:

Issue #907, opened 2021-12-21
Specifying `--no-ignore-vcs` option also turns on `--no-ignore-parent` option


phongnh:
**What version of `fd` are you using?**
`fd 8.3.0`

Here is my use case, I don't use `.gitignore` at project but I use custom `.ignore` or `.fdignore` at project's parent repo.

```sh
$ tree -L 2 -a ~/projects
/home/phongnh/projects
├── repo
│   ├── changelog.md
│   ├── ...
│   ├── .git
│   └── .gitignore
└── .ignore
```

```sh
cd ~/projects/repo
fd --type file --no-ignore-vcs # ==> this does not respect ~/projects/.ignore
```

Based on my check, it seems this comes from this PR https://github.com/sharkdp/fd/pull/817.

Please help to check this, thanks.

sharkdp:
Fixed in https://github.com/sharkdp/fd/releases/tag/v8.3.1

Issue #901, opened 2021-12-02
[BUG] ~/.ignore & ~/.fdignore are not respected


ahmedelgabri:
(_Bug template gave me 400 a couple of times, that's why I didn't use it_)

`~/.ignore` & `~/.fdignore` are not respected

```shell
echo .git > ~/.ignore # or ~/.fdignore
cd ~/repo
fd --hidden --no-ignore-vcs # will show .git in the results
```

while works with `~/.config/fd/ignore`

```shell
mkdir -p ~/.config/fd
echo .git > ~/.config/fd/ignore
cd ~/repo
fd --hidden --no-ignore-vcs # will NOT show .git in the results
```

fd 8.3.0 on macOS Monterey 12.0.1

tavianator:
Is `~/.ignore` or `~/.fdignore` documented?  As far as I can tell, only `~/.config/fd/ignore` is documented to apply globally.

ahmedelgabri:
@tavianator 

From this comment https://github.com/sharkdp/fd/issues/485 by @sharkdp 

> A `.fdignore` or `.ignore` file will always just affect directories below it (just like `.gitignore` files).

So, technically if I have `~/.ignore` or `~/.fdignore` it should work, and it was working like that.

I'm fine if the behaviour changed now, but at least document that or explicitly mention how it works now. Because I was surprised when I started to see `.git` & `node_modules` popping up in my search.

tavianator:
Ohh right I see.  `~/.ignore`/`~/.fdignore` are not global ignore files but they ought to apply to your whole home directory, including `~/repo`.  And indeed they do in my testing, but `--no-ignore-vcs` breaks it.

This was broken by https://github.com/sharkdp/fd/commit/f8ae334ca9416aae06eed600fec0eb94118a04ea, aka https://github.com/sharkdp/fd/pull/817.  [This snippet](https://github.com/sharkdp/fd/pull/817) turns off `read_parent_ignore` a little too aggressively.

phongnh:
Same problem on my side for fd 8.3. I use `fd` for vim fuzzy finder plugins. In my case I don't use `.gitignore` in project because it ignores config files that I needed like `config/database.yml` in Rails project. Instead, I use `.ignore` (`.fdignore`) at parent dir to ignore custom files that I never open. For example:

```
$ tree -L 1 ~/projects/
.
├── rails-app1
└── rails-app2
```

```
# ~/projects/.ignore
vendor/
*.scss
*.js
```

[...]

PR #908, opened 2021-12-21
Fix logic for --no-ignore-parent

tmccombs:
Make sure that using `--no-ignore-vcs` or `--no-ignore` don't also
enable `--no-ignore-parent`. So that if `--no-ignore-vcs` is enabled, it
continues to respect .fdignore and .ignore in the parent directories.

Fixes: #907
Fixes: #901


tavianator (commented):
Looks good to me (but needs `cargo fmt` apparently)

tavianator (approved):

tmccombs:
Yeah haven't gotten a chance to run that yet

sharkdp:
Thank you! I think we should probably publish a new release once the perf. regression on current master (see #895) has been fixed.

tmccombs:
Agreed

tmccombs:
Alternatovely, we could chery pick this fix onto a branch from the last release and release that

PR #909, opened 2021-12-23
Allow use to disable jemalloc

eatradish:
jemalloc will require special flags on devices where page size is not 4K. With default options fd will not work on a Mac Mini M1 where hardware only allow 16K paging size. This commit will allow the user to turn off jemalloc (using --no-default-features) under special circumstances where jemalloc is not usable.


tmccombs (approved):

sharkdp:
Thank you very much.

PR #912, opened 2021-12-24
Update manpage to match --help output

rlue:
Three command line options were added in late 2021
that never made their way into the manpage:

* f8ae334 introduced --no-ignore-parent;
* bf9e6fd introduced --strip-prefix
  (later renamed strip-cwd-prefix); and
* 17dd2a6 introduced --batch-size.

This commit updates the manpage content to reflect
the auto-generated output of `fd --help`.

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -66,6 +66,10 @@ git setting, which defaults to
 .IR $HOME/.config/git/ignore ).
 The flag can be overridden with '--ignore-vcs'.
 .TP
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

PatrickF1:
Thank you! I was just about to PR the same thing for strip-cwd-prefix

sharkdp (approved):
Thank you very much!

PR #913, opened 2021-12-27
Make `--full-path` switch on automatically

xhjkl:
This PR suggests that `-p/--full-path` shall get enabled whenever there is a slash in the search pattern. The same way case sensitivity switches on when there is different casing in the input pattern.

That is, before these changes, `fd src/main` would silently give nothing; and now it just works.
This behavior might be useful for completion setups like fzf's `^T` or suchlike.

Also dropped is the error message when one types a trailing slash in the pattern: `fd src/`. I'm a bit sorry about that — this message was helpful, but since `fd src/` is now a legitimate way to invoke fd, it may be the right thing to do. But suggestions are always welcome.

There are points I thought about, but did not address within this contribution. Namely, since the flag now sets on automatically, we probably need and explicit `--no-full-path`, but I did not want to bloat the command-line interface. This behavior is also easy to trip up with `fd '[^/]'` or similar — this is why it does not get applied to glob patterns. But I believe if one explicitly writes out a `[^/]` in their pattern, they probably mean something by it.

----

Anyway, thank you so much for this brilliant program, and apologies if I have done anything stupid in my PR ✨

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -74,12 +74,12 @@ fn run() -> Result<ExitCode> {
     let search_paths = extract_search_paths(&matches, current_directory)?;
 
     let pattern = extract_search_pattern(&matches)?;
-    ensure_search_pattern_is_not_a_path(&matches, pattern)?;
     let pattern_regex = build_pattern_regex(&matches, pattern)?;
 
     let config = construct_config(matches, &pattern_regex)?;
     ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regex)?;
     let re = build_regex(pattern_regex, &config)?;
+
     walk::scan(&search_paths, Arc::new(re), Arc::new(config))
 }
 
@@ -165,27 +165,6 @@ fn update_to_absolute_paths(search_paths: &mut [PathBuf]) {
     }
 }
 
-/// Detect if the user accidentally supplied a path instead of a search pattern
-fn ensure_search_pattern_is_not_a_path(matches: &clap::ArgMatches, pattern: &str) -> Result<()> {
-    if !matches.is_present("full-path")
-        && pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(pattern).is_dir()
-    {
-        Err(anyhow!(
-            "The search pattern '{pattern}' contains a path-separation character ('{sep}') \
-             and will not lead to any search results.\n\n\
-             If you want to search for all files inside the '{pattern}' directory, use a match-all pattern:\n\n  \
-             fd . '{pattern}'\n\n\
-             Instead, if you want your pattern to match the full file path, use:\n\n  \
-             fd --full-path '{pattern}'",
-            pattern = pattern,
-            sep = std::path::MAIN_SEPARATOR,
-        ))
-    } else {
-        Ok(())
-    }
-}
-
 fn build_pattern_regex(matches: &clap::ArgMatches, pattern: &str) -> Result<String> {
     Ok(if matches.is_present("glob") && !pattern.is_empty() {
         let glob = GlobBuilder::new(pattern).literal_separator(true).build()?;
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you very much for your contribution. I will need to think about this some more. I would also be interested in the feedback of the other contributors.

tmccombs:
Is there any case in which a pattern containing the path separator would actually match anything? On unix operating systems, I'm pretty sure path components are not allowed to contain "/", and "/" isn't a special character in regex. I suppose it could be part of a the name for a named group, although that seems kind of unlikely.  I'm not sure about whether windows path components can contain "\" escaped somehow. And possibly it could be problematic in cygwin and/or WSL? 

At the very least though, the current implementation won't work well on windows, because "\" is used as an escape character in regex, so on windows any regex that uses "\" as a special regex character will result in this behavior. Possibly, looking for "\\" instead would be sufficient?

src/main.rs
```diff
@@ -221,6 +200,13 @@ fn construct_config(matches: clap::ArgMatches, pattern_regex: &str) -> Result<Co
         .map_or_else(filesystem::default_path_separator, |s| Some(s.to_owned()));
     check_path_separator_length(path_separator.as_deref())?;
 
+    // Search full path if the pattern contains a path separator, even if the flag was not set.
+    // But only do so for hand-written patterns.
+    let search_full_path = matches.is_present("full-path")
+        || (cfg!(unix)
+            && !matches.is_present("glob")
+            && pattern_regex.contains(std::path::MAIN_SEPARATOR));
```
tmccombs, line 208: On windows MAIN_SEPARATOR is a backslash "\", which is also a special character in regex, which means any regex that uses "\" will be forced to use full-path matching.

tmccombs (changes requested):

tmccombs:
Oops, I forgot to include my conclusion.

If we are pretty confident that full-path would only be switched on in cases where the pattern wouldn't be able to match anything without full-path being on, then I'm leaning towards automatically enabling it. But maybe we should print a warning saying that we did so, unless the `--full-path` option is explicitly set?

src/main.rs
xhjkl: Understood; will have to think through this.

sharkdp:
I like the general idea. And as you pointed out, there is precedent for this with the smart-case option of `fd`. To summarize, there are the following concerns so far:

- The problem that `\` is also a regex special character, which means that we cannot easily do the same on Windows. This might be solved by actually looking at the regex AST to see if it searches for a `\` character or not. This is similar to what we do for enabling smart-case (see `src/regex_helper.rs`).
- The problem that someone could use a `/` in a regex WITHOUT wanting to do a full-path search. This sounds unlikely, but we should think more about this. As described above, a possible pattern could include something like `[^/]`. We also might be able to look at the AST to solve this somehow.
- The changed behavior when someone runs `fd src/` (which currently shows an error). To be honest, I do something like `fd . src/` (with or without the trailing `/`) quite often. I also sometimes forget the catch-all `.` pattern and type `fd src/`. I guess this happens to a lot of users. Showing an error is obviously annoying, but still better than simply showing no results (treating `src/` as a pattern in a non `--full-path` search), like we did previously. If we now decide to change the behavior to simpliy perform a `--full-path` search, the UX get's a bit worse again, in my opinion. Because instead of showing everything beneath `src/`, we will now search for `src/` in the full path, which might include other results as well. But maybe it's not too bad either.

tavianator:
It's definitely possible to check if a regex can potentially match a `/`, as well as whether it can *only* match strings that contain a `/` (or a `\\`).  This may not be "explicit" in the regex though, e.g. `[!-@]` includes `/`.

The `fd src/` issue is a good point.  I do this *all the time* because I'm used to `find`.

tavianator:
> Is there any case in which a pattern containing the path separator would actually match anything?

As far as I know, the only possibility is the root directory `/`, which is named `/`:

```shell
$ find / -name /
/
```

But `fd` doesn't let you do that:

```shell
$ fd / /
[fd error]: The search pattern '/' contains a path-separation character ('/') and will not lead to any search results.

If you want to search for all files inside the '/' directory, use a match-all pattern:

  fd . '/'

  fd --full-path '/'
```

sharkdp:

PR #916, opened 2021-12-28
Revert "Switch from std::sync::mpsc to crossbeam-channel"

tavianator:
@sharkdp [noticed][1] a quite severe performance degredation due to this
change.  Switch back to std::sync::mpsc until we can fix the performance
regression.

This reverts commit a4bb734482c7bf3fd4d30081e7d3ce7786a3775e.

[1]: https://github.com/sharkdp/fd/pull/895#issuecomment-1002099369

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,16 +100,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


tavianator:
I suppose it's worth mentioning that this makes it possible we'll encounter https://github.com/rust-lang/rust/issues/39364, as we are using `recv_timeout()`

sharkdp:
Merging this for now until the situation has been clarified. I opened a new ticket for this in #933.

PR #929, opened 2022-01-04
Upgrade to clap 3.0

tmccombs:
This wasn't backwards compatible so required some more substantial
changes.


tmccombs:
This supersedes #929

build.rs
```diff
@@ -1,11 +1,13 @@
 use std::fs;
 
-use clap::Shell;
+use clap_complete::{generate_to, Shell};
+use Shell::*;
+//use clap_complete::shells::Shel{Bash, Fish, PowerShell, Elvish};
```
sharkdp, line 5: Can this be removed?

build.rs
```diff
@@ -1,11 +1,13 @@
...
 
 include!("src/app.rs");
 
 fn main() {
-    let min_version = "1.53";
```
sharkdp, line 8: Can we please also update all `README.md` files to include the new version?

src/app.rs
```diff
@@ -758,3 +763,8 @@ pub fn build_app() -> App<'static, 'static> {
 
     app
 }
+
+#[test]
+fn verify_app() {
+    build_app().debug_assert()
```
sharkdp, line 769: what does this do?

tests/tests.rs
```diff
@@ -1420,12 +1420,12 @@ fn test_exec_batch() {
 
         te.assert_failure_with_error(
             &["foo", "--exec-batch", "echo", "{/}", ";", "-x", "echo"],
-            "error: The argument '--exec <cmd>' cannot be used with '--exec-batch <cmd>'",
+            "error: The argument '--exec-batch <cmd>...' cannot be used with '--exec <cmd>...'",
```
sharkdp, line 1423: Hm. This is unexpected. `--exec-batch` can only occur once, IIRC. Maybe it now needs to be marked as `.multiple_occurences(false)`? Not sure why this wasn't required before.

sharkdp (changes requested):
Thank you very much for this!

I'm assuming you went through the migration guide (https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating)?

Cargo.lock
```diff
@@ -85,18 +85,28 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "2.34.0"
+version = "3.0.1"
```
sharkdp, line 88: Can we bump this again. There have been a couple of minor bugfix releases following the 3.0 release (3.0.5 as of writing).

build.rs
tmccombs: ah yes.

tests/tests.rs
tmccombs: It still can't occur more than once:

image

I think the `...` just indicates it takes more than one argument. Although since it is a single word, it may be a little confusing. I can't find a way to improve the way clap presents it though. (maybe we should make a feature request upstream?)

build.rs
tmccombs: done

Cargo.lock
tmccombs: done

src/app.rs
tmccombs: It tests that all the debug asserts in building the `App` succeed. Adding this test was recommded in the migration guide.

tests/tests.rs
sharkdp: ahh, I see. In that case, I think it's fine.

sharkdp (approved):
Thank you!

PR #937, opened 2022-01-07
Less flaky tests

tmccombs:
Change test_opposing to be less flaky.
    
    Make it resilient to the different invocations returning results in
    different orders.


sharkdp:
Thank you!

Issue #939, opened 2022-01-10
Is `fd -L -t l` supposed to work?


timkgh:
**What version of `fd` are you using?**
fd 8.3.1

I understand that the `-L` and `-t l` are kinda at odds with each other, but I would expect that `fd` could follow the symlinks and also print any symlinks it runs into. Is this a bug or desired behavior?

timkgh:
It finds **broken** symlinks, which I suppose _could_ be useful 😄

tmccombs:
I don't know if it is "supposed to work", but the reason it doesn't work is because of the way the [`walkdir`](https://github.com/BurntSushi/walkdir) crate works.

Specifically, if the option to follow symlinks is enabled, then [`metadata()`](https://docs.rs/walkdir/latest/walkdir/struct.DirEntry.html#method.metadata) method will return the metadata for the target of the symlink rather than the symlink itself. And most of the time, that is what you want, so that for example you can know if the symlink points to a file or a directory, or the modification time of the target of the symlink etc. 

Changing the behavior isn't too difficult. I almost have it working on a branch. But it seems to hang for some reason. I haven't figured it out yet.

sharkdp:
I would like to close this for now. @timkgh See reasoning in #940.

PR #940, opened 2022-01-10
Include valid symlinks with both --follow and --type symlink

tmccombs:
This changes it so that if you specify `--type symlink` then it will
check if each entry is a symlink even if we would otherwise follow it.

Fixes: #939

### DIFF
--- a/src/filetypes.rs
+++ b/src/filetypes.rs
@@ -15,10 +15,11 @@ pub struct FileTypes {
 
 impl FileTypes {
     pub fn should_ignore(&self, entry: &walk::DirEntry) -> bool {
+        let is_symlink = entry.is_symlink();
         if let Some(ref entry_type) = entry.file_type() {
-            (!self.files && entry_type.is_file())
-                || (!self.directories && entry_type.is_dir())
-                || (!self.symlinks && entry_type.is_symlink())
+            (!self.files && entry_type.is_file() && !is_symlink)
+                || (!self.directories && entry_type.is_dir() && !is_symlink)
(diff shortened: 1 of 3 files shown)


src/filetypes.rs
```diff
@@ -15,10 +15,11 @@ pub struct FileTypes {
 
 impl FileTypes {
     pub fn should_ignore(&self, entry: &walk::DirEntry) -> bool {
+        let is_symlink = entry.is_symlink();
         if let Some(ref entry_type) = entry.file_type() {
-            (!self.files && entry_type.is_file())
-                || (!self.directories && entry_type.is_dir())
-                || (!self.symlinks && entry_type.is_symlink())
+            (!self.files && entry_type.is_file() && !is_symlink)
```
tmccombs, line 20: It might be better for this (and the following line) to be something like:

```suggestion
            (!self.files && entry_type.is_file() && !(self.symlinks && is_symlink))
```
(maybe factoring that out self.symlinks && is_symlink into a variable?)

so that if you specify only `--type file` it doesn't include symlinks to directories and vice versa.

tavianator:
I'm not sure we should change behaviour here.  RIght now `fd` is consistent with `find`: `find -L -type l` also finds only broken symlinks.

sharkdp:
I also think we should stick with the current behavior and maintain compatibility to `find`. Related discussion: https://github.com/BurntSushi/ripgrep/pull/1503 https://github.com/sharkdp/fd/issues/357

tmccombs:
I'm ok with that.

Issue #943, opened 2022-01-13
Let `--exec` print final commands?


hg-zt:
It is helpful for debugging and showing progress

The idea is
```bash
$ fd .mp4 --threads=1 --exec-show-command some-video-processing-program --input {} --output {.}-output.mp4 
> some-video-processing-program --input 1.mp4 --output 1-output.mp4
(...actual output of the program)
> some-video-processing-program --input 2.mp4 --output 2-output.mp4
> some-video-processing-program --input subdir/1.mp4 --output subdir/1-output.mp4
```
It's like `set -x` for bash.

sharkdp:
Thank you for the feature request. Did you see the discussion in #946? I think it would be good to think about this some more before implementing it. One idea to support this would be to implement #406 (see discussion therein)

tmccombs:
Some other options (mentioned in the PR):

- Add something to documentation mentioning you can put echo in front of the command to get a dry run with warning that it may not be the exact commands that are actually run. 
- add an equivalent of the -ok option in find that asks the user to confirm before actually running each command. Name could be interactive, confirm, verify, etc

sharkdp:
@tmccombs Note that this feature request is slightly different, as it asks for the printing of the command just before it is being executed (not a dry run). So the `echo` approach currently doesn't work. But it would work if we allowed for multiple `-x`/`--exec` instances.

hg-zt:
> Thank you for the feature request. Did you see the discussion in #946? I think it would be good to think about this some more before implementing it. One idea to support this would be to implement #406 (see discussion therein)

It is somehow different from `--dry-run` because one of my use cases is to show the progress for long-running jobs, i.e. which file/directory is under processing. Currently I'm using `-x bash -c` as a workaround.
And yeah I think the request could be solved by implementing #406.

Flimm:
This issue was closed as completed, but I'm confused. What is the fix? How can I make `fd` print the command before running it?

The only solution I found was to resort to using `xargs`:

    fd -0 --glob '*.mp4' | xargs -0 -I {} sh -c 'echo {} ; some-video-processing-program {}'

PR #946, opened 2022-01-22
add --dry-run option to cli and implement it

savente93:
I've implemented a `--dry-run` since it is a need that I've found myself having when using `fd`. I'm aware that I should have created an issue first for discussion, which I only realised after implementing the feature. However, because it is somewhat related to #943 as well as being a relatively minor/simple change I decided to open a PR anyway. If there are things that will need discussion I'm happy to close this and open another PR once all the discussions are resolved. 

Otherwise, there is not that much to say about this change. All it does is take the `dry_run` option from the CLI and pass it down to the commands being executed so they can be printed instead of executed. For this I've mostly just repurposed existing code and tests with slight adjustments for the use case. Hoping this will be an easy merge :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ## Features
 
+- Add "--dry-run" option to print commands instead of executing when used with "--exec" to make tesing a bit easier (@savente93)
 
 ## Bugfixes
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -411,6 +411,11 @@ pub fn build_app() -> App<'static> {
                            fd -e jpg -x convert {} {.}.png\
                     ",
                 ),
+        ).arg(
+            Arg::new("dry-run")
+                .long("--dry-run")
+                .requires("exec")
+                .help("Print the commands to be executed without executing them"),
         )
         .arg(
             Arg::new("exec-batch")
--- a/src/config.rs
+++ b/src/config.rs
@@ -85,6 +85,9 @@ pub struct Config {
     /// If a value is supplied, each item found will be used to generate and execute commands.
     pub command: Option<Arc<CommandTemplate>>,
 
+    /// Whether to print commands instead of executing them for testing
+    pub dry_run: bool,
+
     /// Maximum number of search results to pass to each `command`. If zero, the number is
     /// unlimited.
     pub batch_size: usize,
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -11,26 +11,7 @@ pub fn execute_command(
     mut cmd: Command,
     out_perm: &Mutex<()>,
     enable_output_buffering: bool,
-    dry_run: bool,
 ) -> ExitCode {
-    if dry_run {
-        let _lock = out_perm.lock().unwrap();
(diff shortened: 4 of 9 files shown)


savente93:
I wasn't aware of the minimum rust version. The API used was stabilized in `1.57.0` so there are 3 options:
1. reject this PR/implementation
2. bump the minimal rust version to `1.57.0`
3. add a `#[feature(command_access)]` 
Obviously I'd prefer one of the latter two options, but I'll leave it up to a maintainer to comment on which one is best.

tmccombs:
What benefit does this have over just prepending an echo in front of the command?

savente93:
nothing that I can think of actually, I just didn't think of that. Made the change and ensured it runs and clears tests on 1.54 so should be good now. sorry for that!

tavianator:
I think the question is more like, why is `--dry-run` necessary if the user can just add `echo` themselves?

savente93:
ah, like that. Gotcha. 

Well honestly, not all that much. When I implemented this feature I didn't realise that prepending an echo would still capture most arguments as being associated with the correct command nor that it would still expand the placeholders. 

I had tried the echo method before but in my cases I found that very cumbersome due to the nested " and ' I thought were necessary to associate the commands correctly and actually lost some files due to doing that incorrectly while using `mv`. Having a `--dry-run` option would give (people like) me more security that commands won't be executed which I consider just a QoL feature when I'm working with (critical) files. So I would personally still appreciate the feature, but if you don't think that is worth adopting the additional code over I'd understand too. In that latter case it might be worth considering adding that to the documentation.

sharkdp:
@savente93 Thank you very much for this contribution. I really appreciate the effort and I definitely think we should do something about the problem you mentioned (commands being executed that you didn't intend to).

I usually also use `echo` when using `fd` with `rm` or `mv`. Sometimes you can also use the interactive `-i` option, if there are not too many search results. I agree with the other maintainers that this new option therefore doesn't add  too much value (it's also a bit longer to write). But I'm certainly in favor of updating the documentation with this respect.

There is also another problem here: an official `--dry-run` option might suggest that the shown commands are guaranteed to be executed. It might even suggest that the commands are executed in the shown order. But both assumptions are not true. The execution order is not defined. And the search results could also have changed between running `--dry-run --exec <cmd>` and `--exec <cmd>`. So a official `--dry-run` flag would have to come with a big warning (the `echo` method has the same problem of course).

A way to potentially solve this (although I'm not convinced this is a good idea) would be to add a flag similar to the `--interactive` flags of `rm`/`mv`, where we would ask for confirmation before running each command. Or maybe show a list of to-be-executed commands first and then get the user confirmation.

savente93:
Thanks for the kind words, that really means a lot to me :) I think all the points made are valid. Shall I close this PR and open an issue to discuss this further, or would you like to continue the conversation here?

tavianator:
> A way to potentially solve this (although I'm not convinced this is a good idea) would be to add a flag similar to the `--interactive` flags of `rm`/`mv`, where we would ask for confirmation before running each command.

`find` supports this as `-ok`/`-okdir`

PR #968, opened 2022-02-28
Clippy fixes

tmccombs:


tavianator:
This one also has the crossbeam revert<sup>2</sup>

src/exec/input.rs
```diff
@@ -5,13 +5,13 @@ use crate::filesystem::strip_current_dir;
 
 /// Removes the parent component of the path
 pub fn basename(path: &Path) -> &OsStr {
-    path.file_name().unwrap_or_else(|| path.as_os_str())
+    path.file_name().unwrap_or(path.as_os_str())
```
sharkdp, line 8: It looks like this introduces a clippy warning? Or are you running a different clippy version offline?

tmccombs: My local clippy version is 0.1.59. And the code quality github action passed. What is the warning it introduces?

sharkdp: Strange. I'm seeing this inline annotation on Githubs diff view:

image

sharkdp:
Anyway, let's merge this. Thank you!

src/exec/input.rs
tmccombs: ah! the warning you see is from the minimum supported rust version check.

I found the commit that changed this in clippy: https://github.com/rust-lang/rust-clippy/pull/7639/files#diff-bada3dbc6599bd1411fea3afe759dbb229221e3d62d46bc8c835d5cb731e81b8

So, it now recommends using `unwrap_or` because it thinks `as_os_str()` is cheap (and it is) as of rust 1.59, but on older versions, it warned that it should use `unwrap_or_else`.

PR #985, opened 2022-03-19
Match against reference instead of value

Bankq:
I only recently started learning Rust and `fd` has been a great codebase to learn from!

This patch is more like a question: would matching against reference be equivalent to matching values in this case? It comes with a tiny cosmetic gain that doesn't require `ref` annotation anymore.


tavianator:
Yes it should be equivalent.  The feature is known as [match ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html) and Rust didn't always have it.  But it's quite nice IMO.

tavianator (approved):

Bankq:
TIL, thanks!

tmccombs (approved):

PR #1000, opened 2022-04-13
Update README.md

Bost:
GuixOS installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -634,6 +634,13 @@ Or via [Chocolatey](https://chocolatey.org):
 choco install fd
 ```
 
+### On GuixOS
+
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

README.md
```diff
@@ -634,6 +634,13 @@ Or via [Chocolatey](https://chocolatey.org):
 choco install fd
 ```
 
+### On GuixOS
+
+You can install [the fd package](https://guix.gnu.org/en/packages/fd-8.1.1/) from the official repo:
```
sharkdp, line 639: Is it possible to link to something that doesn't include a particular version number?

Bost: Unfortunately, I don't see any way how to do it.

PR #1002, opened 2022-04-13
sudo is added to README.md for some CentOS installation lines

mperov:
```
$ sudo chown -R root:root fd-v8.3.2-x86_64-unknown-linux-gnu/
$ cd fd-v8.3.2-x86_64-unknown-linux-gnu/
$ sudo cp fd /bin
$ gzip fd.1
gzip: fd.1.gz: Permission denied
$ chown root:root fd.1.gz
chown: changing ownership of `fd.1.gz`: Operation not permitted
```


mperov:
@tavianator hello! Yeah, I agree, deleting `chown -R root:root` is more good solution!

sharkdp:
@mperov Do you want to update the PR to the proposed solution (assuming it has been tested)?

mperov:
@sharkdp  yes, of course! But I need some time because now I'm so busy....

mperov:
Hi @sharkdp! I've done!

tavianator:
Do you need the `sudo su`?  I think the recipe I gave in https://github.com/sharkdp/fd/pull/1002#issuecomment-1098192088 should work fine with fewer commands run as root.

If you do need it, `sudo -i` is a bit better than `sudo su`

mperov:
@tavianator `sudo su` is needed to decrease count of sudo in steps. Of course, `sudo su` and `sudo -i` are equal. However, for me `sudo su` is more familiar. Should I change it to you accept my PR?

Your solution doesn't set root owner for files. In system directories files should have root owner.

tavianator:
> Your solution doesn't set root owner for files

It does, because of `sudo cp`

mperov:
Hello @tavianator 
I removed sudo su and add sudo only to copy to the system directories like you recommended.

sharkdp:

Can we maybe merge the two almost-identical sections? Do we really need the GNU version on one OS and musl on the other? Or is that an entirely independent discussion?

PR #1020, opened 2022-05-19
exec: Execute batches before they get too long

tavianator:
Fixes #410.

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -5,11 +5,14 @@ mod token;
 
 use std::borrow::Cow;
 use std::ffi::{OsStr, OsString};
+use std::io;
+use std::iter;
 use std::path::{Component, Path, PathBuf, Prefix};
-use std::process::{Command, Stdio};
+use std::process::Stdio;
 use std::sync::{Arc, Mutex};
 
 use anyhow::{bail, Result};
+use argmax::Command;
 use once_cell::sync::Lazy;
 use regex::Regex;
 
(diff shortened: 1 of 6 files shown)


tavianator:
Based on https://github.com/sharkdp/argmax/pull/5.

Currently some tests fail because they expect the arguments to be sorted.  That seems hard to do.  If you're okay with it, I'll fix the tests to ignore the order.

src/exec/mod.rs
```diff
@@ -195,43 +294,10 @@ impl CommandTemplate {
     fn generate(&self, input: &Path, path_separator: Option<&str>) -> Command {
         let mut cmd = Command::new(self.args[0].generate(&input, path_separator));
         for arg in &self.args[1..] {
-            cmd.arg(arg.generate(&input, path_separator));
+            cmd.try_arg(arg.generate(&input, path_separator)).unwrap();
```
sharkdp, line 297: I haven't reviewed the full code, but it looks strange that we are forced to use `.unwrap()` here. Is it because we have used `args_would_fit` somewhere before? If so, is this really the best API for `argmax`? Or would it be enough if we had `try_args` that would leave the `Command` untouched in case it returns an error?

tavianator: Oh I don't think I meant to leave this unwrap() here, just hadn't finished converting everything to argmax

sharkdp: Oh, I see. Sorry for the confusion.

tavianator:
I even tried running the exact workflow locally with https://github.com/nektos/act and it passed.  Very confused.

tavianator:
https://github.com/sharkdp/fd/runs/6596619198?check_suite_focus=true#step:5:1

I think I know what's happening: CI is testing the merge with `master`, not the branch itself.  :facepalm:

tests/tests.rs
```diff
@@ -1508,11 +1547,6 @@ fn test_exec_batch_with_limit() {
 
     let te = TestEnv::new(DEFAULT_DIRS, DEFAULT_FILES);
 
-    te.assert_output(
-        &["foo", "--batch-size", "0", "--exec-batch", "echo", "{}"],
-        "./a.foo ./one/b.foo ./one/two/C.Foo2 ./one/two/c.foo ./one/two/three/d.foo ./one/two/three/directory_foo",
-    );
```
sharkdp, line 1514: Why was this removed? Because it could fail on a hypothetical platform with a very low argmax limit?

src/exec/mod.rs
```diff
@@ -89,20 +92,117 @@ impl CommandSet {
...
         }
-        ExitCode::Success
+
+        let arg = self.path_arg.generate(path, separator);
+        if !self
+            .cmd
+            .args_would_fit(iter::once(&arg).chain(&self.post_args))
+        {
+            self.finish()?;
+        }
```
sharkdp, line 189: Very nice :heart:

tests/tests.rs
sharkdp: Ah no, because it makes assumptions about the sort order. Ok. I guess it's not really worth the effort. This path is already tested sufficiently in other tests.

sharkdp (approved):

sharkdp:
Really clear implementation. I also performed some successful local tests. No questions left :smile:.

tmccombs:
Sorry I didn't get to reviewing this sooner, but noce work @tavianator

tavianator:
Thanks @tmccombs!

PR #1024, opened 2022-05-21
remove description of --batch-size flag from EXAMPLES subsection

cvemys:
--batch-size option is already described in OPTIONS subsection and it's description in EXAMPLES is totally irrelevant.
This also removes the indent before "SEE ALSO" to make it a new
subsection.

First image is a screenshot of current man page and the second one is the man page after this change.

Screenshot from 2022-05-21 18-53-14
Screenshot from 2022-05-21 18-53-31

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -454,11 +454,5 @@ $ fd -e py
 .TP
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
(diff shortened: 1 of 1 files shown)


sharkdp (approved):
Oh, not sure how that happened. Thank you!

Issue #1018, opened 2022-05-14
`-x` catch the `<path>...` arguments (if any)


ccqpein:
**What version of `fd` are you using?**
[fd 8.3.1]

`fd -h` shows `[<path>...]` as the ARGS of `fd`. 

> fd [FLAGS/OPTIONS] [<pattern>] [<path>...]

However, for the special option `-x, --exec <cmd>`. There are two things: 

1. for example, if I run `fd -I -t d '^target$' -x echo ./some_path`. It gonna show the `./some_path` first, then the query results followed.
2. even the `-x, --exec <cmd>` is an *option*, but looks like it can only used after `<pattern>`.

For the first situation, I understand that `-x` is special and it is hard to tell if the `path` after it is the arguments of command or the search path. However, do you guys think it should be some warning inside README?

For the second situation, can we update the `fd` command's format like: `fd [FLAGS/OPTIONS] [<pattern>] [-x/-X][<path>...]`? 

Thank you!

tavianator:
`-x` eats everything that comes after it until `;`, so if you want to you can write

```console
$ fd -x echo \; pattern path
```

but it's a lot easier to just write

```console
$ fd pattern path -x echo
```

This is mentioned in the README [here](https://github.com/sharkdp/fd#placeholder-syntax), but isn't obvious from the `fd -h` output.  `fd --help` has some examples, but none with patterns or paths that I see.

sharkdp:
Thank you for reporting this. Let's improve the documentation then.

PR #1025, opened 2022-05-22
Add an example for where to put pattern/path when using -x

merelymyself:
As per #1018, `fd --help` for `-x` is not very helpful at demonstrating that -x should be at the end of `fd`, after paths and patterns. This PR adds an example to make it more obvious.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -395,6 +395,7 @@ pub fn build_app() -> Command<'static> {
                 .long_help(
                     "Execute a command for each search result in parallel (use --threads=1 for sequential command execution). \
                      All positional arguments following --exec are considered to be arguments to the command - not to fd. \
+                     --exec will eat everything that comes after it until ';', even if it is meant to be the path or pattern. \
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you. I'm not super happy with the example. We tried to find realistic use cases so far. And commands that would hopefully be self-explanatory. Maybe we can come up with a better, more realistic example? Or change one of the existing examples instead?

Or just explain it in plain text that it's useful to place `-x` last (and why)?

merelymyself:
@sharkdp - the help already states that it is better to put it at the end, so the why part is probably the part we need to add - will something like this work?

sharkdp:
TBH, I didn't remember that we already have the "All positional arguments following --exec are considered to be arguments to the command - not to fd." sentence in there. In this case, I don't think the new addition adds a lot of value. If we can come up with another good example, I'm happy to add that though.

merelymyself:
I see, then I'll just close this first. Thanks!

Issue #533, opened 2020-02-08
Add -printf formatting like GNU find


avielsh:
Hi
Please add a -printf like feature to fd.
Currently, to get the file information I'm running:
```fd --changed-within=10days -x ls -dl --time-style=long-iso "{}"```
Which spawns ```ls``` for every result. 

This is actually much slower than running:
```find . -mtime 10  -printf "%TY-%Tm-%Td %TH:%TM %p\n" ```
(For example)

Test dataset is around 20,000 files. fd is running for ~21 seconds, find for ~10. 
Of course when more results are found, fd method would become slower.
Thanks !

**Comment by sharkdp** [comment]
>  Currently, to get the file information I'm running:
> `fd --changed-within=10days -x ls -dl --time-style=long-iso "{}"`
> Which spawns `ls` for every result.

That's what `-X`/`--exec-batch` is for which would spawn just a single `ls` process. This has the added advantage that the output columns are aligned (minor comment: the `"{}"` argument is superfluous and can be omitted). Would this work for you?
```bash
fd --changed-within=10days -X ls -dl --time-style=long-iso
```

> This is actually much slower than running:
> `find . -mtime 10 -printf "%TY-%Tm-%Td %TH:%TM %p\n" `

The equivalent command would use `-mtime -10`, right?

> Test dataset is around 20,000 files. fd is running for ~21 seconds, find for ~10.

How do you perform your benchmarks? Do you [account for disk caching effects](https://github.com/sharkdp/hyperfine#warmup-runs-and-preparation-commands)?

avielsh:
> That's what -X/--exec-batch is for, which would spawn just a single ls process. 

My result set for that is too large (~20000) files. I'm getting: 
```[fd error]: Problem while executing command: Argument list too long (os error 7)```
Maybe add a batch count argument to limit arguments per batch ?🤔

I actually tried to pipe to ```xargs``` with ```-P4 -n1500``` for batch processing while maintaining the argument list for ```ls``` short enough, but the parallel execution in xargs causes all sorts of string issues in the output.

> The equivalent command would use -mtime -10, right?

Yes my mistake :)

> How do you perform your benchmarks? Do you account for disk caching effects?

[Hyperfine](https://github.com/sharkdp/hyperfine) is cool. I was using ```time```.

Actually, my example was a simplified version of what I actually use, which is listing the whole file system. ( for instant searching like ```locate```)

Here's my hyperfine result (gfind and gls are GNU versions of the commands)
[...]

PR #1043, opened 2022-06-16
Implement option for printing custom formats

tmccombs:
This is mostly a proof of concept before merging there is still some work that needs to be done for:

- [x]  Write new tests (both unit tests and integration tests)
- [x] Add documentation to man page
- [x] Potentially add support for outputting arbitrary bytes on unix like we do for normal printing (I'm not going to worry about this for now)
- [x] Benchmark, and optimize if necessary
    - one potential optimizaiton could be to write directly to the stream rather than building an intermediary OsString, I started looking into that, but it got kind of messy.

Also, this doesn't currently support colorized output when using format string. I think that for an initial implementation that is probably fine.


sharkdp:
Thank you very much for taking a stab at this!

I didn't have time to properly review it, but I think I am in favor of integrating this new option! That is, unless it leads to a performance penalty for the default case (without `--format`).

Something that I thought about in a different context, but seems fitting here: I'm not a huge fan of the `{}`/`{/}`/`{//}`/`{.}`/… syntax, which we borrowed from GNU parallel, I believe. Would it make sense to think about something a bit more verbatim? I think we could easily support something like `{path}`/`{basename}`/`{parent}`/`{without_extension}`/… in addition. What do you think? I'd much rather type out those words than having to look up the syntax each time.

tmccombs:
I like the more verbose names too.

sharkdp:
Irregardless of the short/long syntax discussion, I'd like to see this being integrated to `fd`!

@tmccombs Are you still interested in finishing this? It would probably be best to focus on the benchmarks first, to make sure that we don't run into performance regressions.

tmccombs:
Sorry it took me so long to get back to this. I am interested in finishing this. Though I wouldn't mind some help with benchmarking it.

tmccombs:
I fixed the merge conflicts.

One question I have is if we would rather have a [--printf option](https://github.com/sharkdp/fd/issues/533) with more format optiions, such as access and modified times, owning user and group, etc. Certainly not as extensive as find, but more than just variants of the path. Although, I  really don't love the idea of having two different format syntaxes.

sharkdp:
> One question I have is if we would rather have a [--printf option](https://github.com/sharkdp/fd/issues/533) with more format optiions, such as access and modified times, owning user and group, etc. Certainly not as extensive as find, but more than just variants of the path. Although, I really don't love the idea of having two different format syntaxes.

I would personally like if we extend the `{…}` syntax. For example, we could have `{basename}` as an alias for `{/}`. This would allow us to use `{access_time}` and similar for an extended printing syntax. If we need more control over *how* a field like a date is printed, we could follow Python/Rust f-strings with something like `{access_time:%Y-%m-%d}`. I don't think that we need to have a find `-printf` compatibility mode here. What do you think?

tmccombs:
That sounds good to me.

cohml:
@tmccombs @sharkdp - Any chance this will be merged soon?

I'm dying for this feature and compulsively checking this PR every few days as a result. Please put me out of my misery lol.

src/cli.rs
```diff
@@ -452,6 +452,20 @@ pub struct Opts {
...
+    ///   '{}': path (of the current search result)
+    ///   '{/}': basename
+    ///   '{//}': parent directory
+    ///   '{.}': path without file extension
+    ///   '{/.}': basename without file extension
+    #[arg(
+        long,
+        value_name = "fmt",
+        help = "Print results according to template",
+        conflicts_with = "list_details"
```
sharkdp, line 465: :+1:

sharkdp (approved):
Thank you very much and sorry for the late review!

tmccombs:
I've run a few benchmarks with hyperfine, and haven't noticed a significant difference in performance.

PR #1048, opened 2022-06-24
Clap 3.2

tmccombs:
This doesn't replace every deprecated api in  clap 3.2, but it does make use of some of the new features in the places where I thought it was most valuable.


sharkdp:
Thank you for working on this. On a first glance, the new code looks quite a bit more complicated than what we had before, or am I mistaken? The build process also seems to be more complicated with a new feature flag?

src/app.rs
```diff
@@ -525,6 +552,7 @@ pub fn build_app() -> Command<'static> {
                 .short('S')
                 .takes_value(true)
                 .number_of_values(1)
+                .value_parser(SizeFilter::from_string)
```
epage, line 555: Personally, I'd recommend splitting up the resolving of deprecations and using new features into separate PRs. 

This way it doesn't get blocked on discussions around items like

> On a first glance, the new code looks quite a bit more complicated than what we had before, or am I mistaken? The build process also seems to be more complicated with a new feature flag?

sharkdp:
@tmccombs Could you please explain how this is related to #1067?

tmccombs:
I'm just going to close this, in favor of #1067. I tried making this first, but then though that using the derive api might be better.

PR #1061, opened 2022-07-12
Implement --type executable on Windows

### LINKED ISSUE #1051: -t executable flags can't show windows executable
### `-t executable` flags can't show windows executable 

图片

tavianator:
Fixes #1051.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- `--type executable`/`-t` now works on Windows, see #1051 and #1061
 
 ## Bugfixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "faccess"
+version = "0.2.4"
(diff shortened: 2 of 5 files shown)


tavianator:
I don't know enough about Windows to actually make a test for this.  Might be nice if someone with Windows can confirm it works.

sharkdp:
Thank you!

> I don't know enough about Windows to actually make a test for this.

We have the `fd.exe` available in the integration test suite, so maybe we could use that as a target for a simple test?

> Might be nice if someone with Windows can confirm it works.

/cc @xBLACKICEx

xBLACKICEx:
> Thank you!
> 
> > I don't know enough about Windows to actually make a test for this.
> 
> We have the `fd.exe` available in the integration test suite, so maybe we could use that as a target for a simple test?
> 
> > Might be nice if someone with Windows can confirm it works.
> 
> /cc @xBLACKICEx

why not

tmccombs (approved):

sharkdp (approved):

sharkdp:
Thank you!

PR #1062, opened 2022-07-12
Enable jemallocator on 64-bit musl builds

### LINKED ISSUE #710: fd is allocation heavy, which can lead to heavy lock contention on some malloc implementations, especially hardened ones

When comparing the speed of `fd` with musl vs glibc, they are very comparable with `-j1`, but when the core count goes above a certain threshold, musl gets diminishing returns, to the point that with `-j12`, on a 6-core system (with hyperthreading), it's actually slower than with `-j4`.

With fd + musl:

```
~ ➜ time fd -j1 freebsd ~/ >/dev/null

________________________________________________________
Executed in    3.17 secs   fish           external
   usr time    1.81 secs    0.00 micros    1.81 secs
   sys time    1.36 secs  599.00 micros    1.36 secs

~ took 3s ➜ time fd -j4 freebsd ~/ >/dev/null

________________________________________________________
Executed in    2.11 secs   fish           external
   usr time    5.40 secs  621.00 micros    5.40 secs
   sys time    2.47 secs   46.00 micros    2.47 secs

~ took 2s ➜ time fd -j8 freebsd ~/ >/dev/null

________________________________________________________
Executed in    2.62 secs   fish           external
   usr time   11.58 secs  476.00 micros   11.58 secs
   sys time    7.39 secs   35.00 micros    7.39 secs

~ took 2s ➜ time fd -j12 freebsd ~/ >/dev/null

________________________________________________________
Executed in    2.98 secs   fish           external
   usr time   16.22 secs    0.00 micros   16.22 secs
   sys time   15.78 secs  559.00 micros   15.78 secs
```

With fd + musl + pre-loaded jemalloc:

```
~ took 2s ➜ time LD_PRELOAD=/usr/lib/libjemalloc.so.2 fd -j1 freebsd ~/ >/dev/null

________________________________________________________
Executed in    2.75 secs   fish           external
   usr time    1.43 secs    0.00 micros    1.43 secs
   sys time    1.32 secs  913.00 micros    1.32 secs

~ ➜ time LD_PRELOAD=/usr/lib/libjemalloc.so.2 fd -j4 freebsd ~/ >/dev/null

________________________________________________________
Executed in  790.87 millis    fish           external
   usr time    1.75 secs  930.00 micros    1.75 secs
   sys time    1.37 secs    0.00 micros    1.37 secs

~ ➜ time LD_PRELOAD=/usr/lib/libjemalloc.so.2 fd -j8 freebsd ~/ >/dev/null

________________________________________________________
Executed in  552.14 millis    fish           external
   usr time    2.40 secs    0.00 micros    2.40 secs
   sys time    1.89 secs  882.00 micros    1.89 secs

~ ➜ time LD_PRELOAD=/usr/lib/libjemalloc.so.2 fd -j12 freebsd ~/ >/dev/null

________________________________________________________
Executed in  539.76 millis    fish           external
   usr time    3.42 secs  828.00 micros    3.42 secs
   sys time    2.76 secs    0.00 micros    2.76 secs
```

With fd + glibc:

```
[chroot] /home/ericonr > time fd -j1 freebsd ~/ >/dev/null

________________________________________________________
Executed in    2,73 secs   fish           external
   usr time    1,28 secs  488,00 micros    1,28 secs
   sys time    1,35 secs  141,00 micros    1,35 secs

[chroot] /home/ericonr > time fd -j4 freebsd ~/ >/dev/null

________________________________________________________
Executed in  854,58 millis    fish           external
   usr time    1,75 secs    0,00 micros    1,75 secs
   sys time    1,61 secs  585,00 micros    1,61 secs

[chroot] /home/ericonr > time fd -j8 freebsd ~/ >/dev/null

________________________________________________________
Executed in  601,37 millis    fish           external
   usr time    2,25 secs  439,00 micros    2,25 secs
   sys time    2,35 secs  144,00 micros    2,35 secs

[chroot] /home/ericonr > time fd -j12 freebsd ~/ >/dev/null

________________________________________________________
Executed in  558,01 millis    fish           external
   usr time    3,10 secs  612,00 micros    3,10 secs
   sys time    3,15 secs    0,00 micros    3,15 secs
```

With `perf`, I can track the issue (apparently) down to `__lock`:

```
  Children      Self  Command  Shared Object     Symbol
+   29.83%    29.58%  fd       libc.so           [.] __lock
+   27.29%     0.00%  fd       libc.so           [.] a_cas (inlined)
```

This matches with what was observed: with more threads, you get more lock contention, and can end up spending way too much time locked. I'd expect a similar effect to be observable with `hardened_malloc` from GrapheneOS and possibly OpenBSD's malloc. Theoretically speaking, rust might not require a hardened malloc, but you usually don't want to carry your own malloc for each application either, so it'd be nice if this could perform better "out of the box", by simply performing less allocations (which helps by removing most of the dependency on malloc performance). That said, I don't know how feasible fixing this is.

I'd expected all versions of the test to show similar growth in execution speed. Note that this is in my `~/` directory, because otherwise execution is short enough that you can't spot the issue (the numbers with a tarball of the linux kernel were too close).

fd 8.2.1

Linux (kernel 5.10.2), with musl 1.2.1

glibc used for testing was 2.30

jemalloc used for `LD_PRELOAD` was 5.2.1

### LINKED ISSUE #980: 3x~10x Performance regression between 7.2.0 and >7.3.0 on large folder
Noticed that some fd commends runs much slower (10x slower) when I upgraded my local fd from 6.2.0 to newest 8.3.2, and did a quick version bisect.

Looks like the regression is between 7.2.0 and 7.3.0, and all version I've tested after 7.3.0 (7.4.0, 7.5.0, 8.0.0, 8.1.1, 8.3.2) are all as about the same speed as 7.3.0.

Reproduce script:
```bash
set -e

wget -q https://github.com/sharkdp/fd/releases/download/v7.2.0/fd-v7.2.0-x86_64-unknown-linux-musl.tar.gz
tar -xf fd-v7.2.0-x86_64-unknown-linux-musl.tar.gz

wget -q https://github.com/sharkdp/fd/releases/download/v7.3.0/fd-v7.3.0-x86_64-unknown-linux-musl.tar.gz
tar -xf fd-v7.3.0-x86_64-unknown-linux-musl.tar.gz

hyperfine --version
hyperfine \
  --warmup 5 \
  './fd-v7.2.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src' \
  './fd-v7.3.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src'
```

(I'm using Chrome OS source tree as an example here, but I can reproduce similar regression on other large source tree, for example, linux source tree)

Result:
* On a VPS without SSD, with 24 cores/96 hyperthreads:
```
hyperfine 1.11.0
Benchmark #1: ./fd-v7.2.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src
  Time (mean ± σ):      2.468 s ±  0.058 s    [User: 90.829 s, System: 115.032 s]
  Range (min … max):    2.402 s …  2.555 s    10 runs
 
Benchmark #2: ./fd-v7.3.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src
  Time (mean ± σ):     25.529 s ±  0.328 s    [User: 222.856 s, System: 1924.844 s]
  Range (min … max):   24.980 s … 26.091 s    10 runs
 
Summary
  './fd-v7.2.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src' ran
   10.34 ± 0.28 times faster than './fd-v7.3.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src'
```

* On my local laptop with SSD, with 4 cores/8 hyperthreads:
```
hyperfine 1.13.0
Benchmark 1: ./fd-v7.2.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src
  Time (mean ± σ):      2.348 s ±  0.101 s    [User: 10.347 s, System: 6.298 s]
  Range (min … max):    2.237 s …  2.527 s    10 runs
 
Benchmark 2: ./fd-v7.3.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src
  Time (mean ± σ):      6.882 s ±  0.090 s    [User: 44.010 s, System: 6.813 s]
  Range (min … max):    6.783 s …  7.065 s    10 runs

Summary
  './fd-v7.2.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src' ran
    2.93 ± 0.13 times faster than './fd-v7.3.0-x86_64-unknown-linux-musl/fd ".*camera_hal.*" ~/chromiumos/src'
```

Also tried adding `--color=never` and the result are similar to this, from the changelog the only other suspect is the `--exec-batch` command?

Happy to provide additional testing / debug info if needed.

tavianator:
Partial fix for #710 and #980.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ faccess = "0.2.4"
 # FIXME: Re-enable jemalloc on macOS
 # jemalloc is currently disabled on macOS due to a bug in jemalloc in combination with macOS
 # Catalina. See https://github.com/sharkdp/fd/issues/498 for details.
-[target.'cfg(all(not(windows), not(target_os = "android"), not(target_os = "macos"), not(target_os = "freebsd"), not(target_env = "musl"), not(target_arch = "riscv64")))'.dependencies]
+[target.'cfg(all(not(windows), not(target_os = "android"), not(target_os = "macos"), not(target_os = "freebsd"), not(all(target_env = "musl", target_pointer_width = "32")), not(target_arch = "riscv64")))'.dependencies]
 jemallocator = {version = "0.5.0", optional = true}
 
 [dev-dependencies]
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,7 +40,7 @@ use crate::regex_helper::{pattern_has_uppercase_char, pattern_matches_strings_wi
     not(target_os = "android"),
     not(target_os = "macos"),
     not(target_os = "freebsd"),
-    not(target_env = "musl"),
+    not(all(target_env = "musl", target_pointer_width = "32")),
     not(target_arch = "riscv64"),
     feature = "use-jemalloc"
 ))]


sharkdp (approved):

tavianator:
Whoops I force pushed `master` instead of this PR branch.  I just undid that, but @sharkdp you might want to enable branch protection for `master` in case I fat finger it again :)

tavianator:
So actually this used to work, but right now the musl version bundled with Rust is too old, leading to https://github.com/tikv/jemallocator/issues/30 https://github.com/tikv/pprof-rs/issues/142

sharkdp:
> Whoops I force pushed `master` instead of this PR branch. I just undid that, but @sharkdp you might want to enable branch protection for `master` in case I fat finger it again :)

Done. No worries.

sharkdp:
How do we want to proceed here? I think it would be great to get this into the next release. Maybe we can just add a special case for the two platforms that are failing (i.e.: disable jemalloc for those) and add a comment that links to the tickets you mentioned above?

That would still give musl+jemalloc on the more important platforms.

tavianator:
@sharkdp That sounds good to me!  I probably won't have time for a few days so feel free to take this over if you do.

sharkdp (approved):
Thank you for the update!

tavianator:
The remaining two cases are 32-bit arm and x86.  I got x86 working in debug builds but release builds still fail to link.  I think x86 may be fixed in a future Rust release by upgrading `compiler-builtins`.  I'm less sure about arm: jemalloc wants 1-byte atomics and I'm not sure arm has those before armv7.

Issue #1042, opened 2022-06-15
Add new flag for printing the executed command


jessestricker:
### Updaptes

1. As per https://github.com/sharkdp/fd/pull/1068#issuecomment-1195460922, 
   adding the short flag `-P` is no longer required. Only the long flag `--print-exec` may be implemented.

### Original

> This is essentially the same request as #943, which was closed by the original author.
> Now that #406 has been implemented, this feature seems feasible again.

I propose to add a new flag `-P, --print-exec`, which by default will be disabled.
When enabled, it causes **fd** to print the *full command line* just before executing it.
Also, when color output is enabled at the same time, all parts in the command line which originally were placeholders will be highlighted.

This new flag adds a shortcut and formatting as a bonus for the common case of using 
`--exec` twice to display and execute a command. 
A similar convenience flag `-l, --list-details` already exists.

This feature request is in contrast to #1037, which proposes for only the filename to be printed.

## Examples

**Existing default behavior does not change:**

```console
$ fd -e rs -x rustfmt
<no output>
```

**With `-P` flag enabled:**

```console
$ fd -P -e rs -x rustfmt
rustfmt ./src/lib.rs
rustfmt ./src/main.rs
```
Here, the `./src/###.rs` part would be highlighted.

**With `-P` flag enabled and multiple placeholders:**

```console
$ fd -P -e wav -x flac -o {.}.flac -T "TITLE=Song" {}
flac -o ./song.flac -T "TITLE=Song" ./song.wav
```
Here, both `./song.flac` and `./song.wav` would be highlighted.

jessestricker:
Also this issue does specifically **not** request a fully configurable output format, for the same reason as https://github.com/sharkdp/fd/issues/1037#issuecomment-1152725706.

ilmari-h:
Would this make for a good first issue? Having looked at the code and started part of the implementation, it seems doable for me. So fyi, I'd like to work on this.

ilmari-h:
Created a draft PR #1068 for this.
At least one thing to be resolved: Should the commands be printed in the same sequence they are ran.

So for example:
```
$ fd -P -x echo

echo file2
file2
echo file3
file3
echo file1
file1
```

Or alternatively
```
$ fd -P -x echo

echo file1
echo file2
echo file3
file2
file3
file1
```
In the latter case, due to commands being executed in multiple threads, the order will not match the order of execution as it's basically random.

tavianator:
[...]

Issue #963, opened 2022-02-23
Show currently processed file in -x


hyperknot:
I'm running a large batch processing job using -x. It'd be nice to display the name of the currently processed file or files (with parallel processing).

I don't know any other way to achieve this. If there is a workaround please let me not.

tmccombs:
Currently, the only way to do this is with something like `bash -c 'echo {}; my-command {}'`. However, once https://github.com/sharkdp/fd/pull/960 is merged you could do it with `-x echo {} \; -x my-command...`. 

I also think this is a duplicate of #943

hyperknot:
I understand. Just a small proposal for `--print-filename` or similar which would print the filename without any of the complexity described above.

tmccombs:
is `-x echo \;` that much more complicated than `--print-filename`?

hyperknot:
Yes, personally I'd find it more complicated. + `--print-filename` could use color output.

sharkdp:
This option seems to be requested a lot, given the amount of related discussions:
- https://github.com/sharkdp/fd/issues/963
- https://github.com/sharkdp/fd/issues/959
- https://github.com/sharkdp/fd/issues/943
- https://github.com/sharkdp/fd/pull/946
- https://github.com/sharkdp/fd/issues/406

@tmccombs Multi-exec is a fantastic feature which I'm very excited about, but maybe we should reconsider adding a simple option for this? Proper formatting would be one argument. Accessibility of the feature another.

tmccombs:
Yeah, I'm persuaded by the formatting/color option.

vladh:
I would like to second the request for this feature, with a slightly more detailed rationale.

When using the `-x` option, one is often dealing with a lot of files, and often performing destructive operations on those files.

In this scenario, I really do not want to add any other `echo` calls or any superfluous syntax to the command I am passing to `-x`, since I do not want to risk getting something wrong and losing data.

Having a simple flag to enable printing would make this common use-case much easier, without presenting the risk of the user accidentally getting the command wrong in the process.

I would even go as far as to say that I think this should be the default behaviour — it's important to know what commands are being run on your files.

sharkdp:
I think we can close this in favor of #1042. Let me know if anyone disagrees.

Kristinita:
**Type: Reply** 💬

### 1. Summary

> is `-x echo \;` that much more complicated than `--print-filename`?

[...]

PR #1068, opened 2022-07-22
Implement new flag for printing the executed commands

ilmari-h:
Implementation for #1042, flag `--print-exec` / `-P`
The flag causes the commands that are going to be executed to first be printed with colored highlighting. Highlighting uses the same logic as when running fd and just listing results. The highlighting is also applied to templates.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -442,6 +442,15 @@ pub fn build_app() -> Command<'static> {
                 ),
         )
         .arg(
+            Arg::new("print-exec")
+                .long("print-exec")
+                .short('P')
+                .help("Print each command ran with -x or -X.")
+                .long_help(
+                    "Print each command to be ran. If -x or -X is not used, this argument has no effect."
+                ),
+        )
+        .arg(
             Arg::new("batch-size")
             .long("batch-size")
             .takes_value(true)
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -5,11 +5,13 @@ use std::{
 
 use once_cell::unsync::OnceCell;
 
+#[derive(Clone)]
 enum DirEntryInner {
     Normal(ignore::DirEntry),
     BrokenSymlink(PathBuf),
 }
 
+#[derive(Clone)]
 pub struct DirEntry {
     inner: DirEntryInner,
     metadata: OnceCell<Option<Metadata>>,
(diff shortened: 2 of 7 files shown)


matu3ba:
Questions: 1. Why not using `-v` as `verbose` output ? 2. Is this use case very frequent to justify a 1 letter command abbreviation?

ilmari-h:
> Questions: 1. Why not using `-v` as `verbose` output ? 2. Is this use case very frequent to justify a 1 letter command abbreviation?

1. You mean, why not `verbose` instead of `print-exec`? I just personally liked the `print-exec` suggestion in the linked issue. I'd intuitively think verbose would refer to also verbose printing of other things aside from just executed commands. 
2. Yeah, I wasn't sure what the convention was in this repo. But looking at other options, you're right. I think a short option isn't required.

So the answer to both: kind of just went by what was proposed in the issue. At least for the short option, I think I'll remove that

jessestricker:
I agree with you and have thus updated the proposal.

ilmari-h:
Fixed the last style warnings, now ready for review. Hope the implementation is satisfactory, tried to keep changes to existing logic minimal and did my best to make sure the runtime performance is not impacted with unnecessary overhead, at least when not using the new flag.

ilmari-h:
Pushed a couple more commits, noticed printing was broken for multiple command instances. Now works with those as well.

sharkdp:

This looks good from a very high level, but I'll probably need some more time to review this in detail. If there are any things in here that could be split out into a separate PR (and are useful independent from this feature), that would be very much appreciated.

danielkrajnik:
wish I've had something more constructive to say other than I'm just hyped for this feature and looking forward to it being merged :+1:

Kristinita:
fd is actively maintained and accepted many pull requests. It’s strange for me that this pull request not reviewed already 2,5 years.

Currently, due to different delimiters between commands (`\;` works on UNIX, `;` on Windows) I have to write separate commands for UNIX and Windows like in [**my MCVE**](https://github.com/sharkdp/fd/issues/963#issuecomment-1974793847). It would be nice if would be possible not to do it only due to backslash.

Thanks.

nyurik:
@tmccombs a friendly ping - this feature would be absolutelly awesome to have. Are there any blockers for it, and/or can it be merged once conflicts are resolved?  Do you need any help with it? (I obviously can't push to this PR, but I could try to fix it up as a separate PR... maybe).  Thx!

Issue #1052, opened 2022-06-30
There is no link back to the Github repo in the man page or the `--help`


scottchiefbaker:
I installed `fd` on my machine a while back, and wanted to come back to the Github repo to make sure I had the most recent version. I checked both the `--help` and the man page for `fd` and neither link back to the project in any way. Googling `linux fd` isn't the most helpful either. 

Please consider adding a link back to Github page in either of the above mentioned places to help users.

PR #1086, opened 2022-08-17
Add a link back the Github page to address #1052

scottchiefbaker:
Add a simple link to the man page to report bugs on Github.

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -456,6 +456,6 @@ $ fd -e py
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
 .SH BUGS
-Bugs can be reported on the Github page: https://github.com/sharkdp/fd
+Bugs can be reported on Github: https://github.com/sharkdp/fd
 .SH SEE ALSO
 .BR find (1)
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -456,6 +456,6 @@ $ fd -e py
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
 .SH BUGS
-Bugs can be reported on Github: https://github.com/sharkdp/fd
+Bugs can be reported on GitHub: https://github.com/sharkdp/fd/issues
 .SH SEE ALSO
 .BR find (1)
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -455,5 +455,7 @@ $ fd -e py
 .TP
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
+.SH BUGS
+Bugs can be reported on the Github page: https://github.com/sharkdp/fd
 .SH SEE ALSO
 .BR find (1)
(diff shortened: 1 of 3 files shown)


scottchiefbaker:
It's hard to tell, but it looks like the `fd --help` page is generated from the man page? I think it would be a good idea to have something on the `--help` page also.

tmccombs:
> It's hard to tell, but it looks like the fd --help page is generated from the man page?

No, the `--help` output is generated by clap from the options defined in src/app.rs

tmccombs (approved):

scottchiefbaker:
Thanks for the clarification. I have updated the PR to include changes to `src/app.rs` to include the Github URL.

src/app.rs
```diff
@@ -12,9 +12,9 @@ pub fn build_app() -> Command<'static> {
         .color(clap_color_choice)
         .setting(AppSettings::DeriveDisplayOrder)
         .dont_collapse_args_in_usage(true)
-        .after_help(
-            "Note: `fd -h` prints a short and concise overview while `fd --help` gives all \
-                 details.",
+        .after_help("Bugs can be reported on Github: https://github.com/sharkdp/fd\n
```
tmccombs: I think you might need a `\` at the end of this line

scottchiefbaker: This line was updated (simplified) in a subsequent push.

scottchiefbaker:
Any updates on this? Anything else I can fix/update.

tmccombs (approved):

doc/fd.1
```diff
@@ -455,5 +455,7 @@ $ fd -e py
 .TP
 .RI "Open all search results with vim:"
 $ fd pattern -X vim
+.SH BUGS
+Bugs can be reported on Github: https://github.com/sharkdp/fd
```
tavianator, line 459: If we're talking about bugs, should we link directly to https://github.com/sharkdp/fd/issues?

scottchiefbaker:
What's next for this PR? I'd really love to see this get merged.

sharkdp:
Thank you! I added `/issues`, as @tavianator suggested and added a CHANGELOG entry. I also removed the comment from the help text because I would really like to keep the `fd -h` output small and concise. If there is a way to add it to `fd --help` only, I'm open to adding it there.

sharkdp (approved):

scottchiefbaker:
@sharkdp I'd really like to see a link in the `--help` output. It's entirely possible someone will have the binary and not the man page and have no way to get back to the project.

How do I put text in the `--help` output and not in the `-h` output?

tavianator:
@scottchiefbaker Clap seems to have an `after_long_help()` method: https://docs.rs/clap/latest/clap/builder/struct.App.html#method.after_long_help

scottchiefbaker:
@tavianator good find! I submitted a PR with some `after_long_help()`.

PR #1092, opened 2022-08-30
add md5sum example to README

### LINKED ISSUE #1090: Is `-x` stdout unified across threads?
`fd 8.4.0`

I am curious because I would like to calculate the `md5sum` of a bunch of files relative to a single directory. Do I need to worry about the different threads or can I naively just pipe the output of `fd` to a file like I would do with the single-threaded `find`?

e.g. What I would do with `find` is
```
find . -type f -exec md5sum {} ';' &> my-sums.md5
```

I _don't_ want to use `-X` because I am calculating the sum for a lot of multi-GB files so utilizing multiple threads could [^1] be helpful.

If `fd` already unifies the terminal output of its threads, then I would humbly request an addition to the documentation that reflect that much (e.g. another example in the "Command Execution" section?).

[^1] : Disk I/O seems to be the main limiting factor but maybe not.

tomeichlersmith:
this includes a statement that `fd` will not interlace or garble the terminal output from its multiple threads

this resolves #1090

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,13 @@ fd -e jpg -x convert {} {.}.png
 Here, `{}` is a placeholder for the search result. `{.}` is the same, without the file extension.
 See below for more details on the placeholder syntax.
 
+The terminal output of commands run from parallel threads using `-x` will not be interlaced or garbled,
+so `fd -x` can be used to rudimentarily parallelize a task run over many files. 
+An example of this is calculating the checksum of each individual file within a directory.
+```
+fd -x md5sum
+```
+
 #### Placeholder syntax
 
 The `-x` and `-X` options take a *command template* as a series of arguments (instead of a single string).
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

README.md
```diff
@@ -206,6 +206,13 @@ fd -e jpg -x convert {} {.}.png
 Here, `{}` is a placeholder for the search result. `{.}` is the same, without the file extension.
 See below for more details on the placeholder syntax.
 
+The terminal output of commands run from parallel threads using `-x` will not be interlaced or garbled,
+so `fd -x` can be used to rudimentarily parallelize a task run over many files. 
+An example of this is calculating the checksum of each individual file within a directory.
+```
+fd -x md5sum
```
sharkdp, line 213: Maybe
```suggestion
fd -tf -x md5sum > file_checksums.txt
```

Issue #1083, opened 2022-08-04
"Command not found" error message could be improved


sharkdp:
Since fd 8.4, we show this error message:

```
▶ fd foo -X non-existent
[fd error]: Command not found: Command { inner: "non-existent" "./foo", remaining_argument_length: 2079626 }
```
which is confusing for the average user.

The reason for that is the usage of [`argmax::Command`](https://docs.rs/argmax/latest/argmax/struct.Command.html) instead of `std::process::Command`. This should probably be fixed by a custom `Debug` `impl` for  `argmax::Command`, instead of using `#[derive(…)]`. This could then delegate to the `Debug` implementation of the inner `std::process::Command`, to maintain maximum compatibility, even in debug prints.

An even better option might be to not rely on debug formats at all and implement a proper formatting function.

fprasx:
I can give this a crack! Looking at `argmax::Command`, it's API doesn't expose the underlying `std::proces::command` at all, so I'm not sure how to even print something like the name without using the complete `Debug` impl `argmax` provides.

PR #1109, opened 2022-09-11
Simplistic "Command not found" error message

themkat:
Don't know if this is too simplistic from what you had in mind. This at least makes sense to me, and I've been using this fix locally for a little while. Feel free to close it if you think the solution should be more advanced or different 🙂 

Attempt at #1083

Looks like this when I use it:
```
▶ fd foo .. -X non-existent
[fd error]: Command not found: "non-existent"
```


themkat:
Seems like the pipeline gives a warning on `use of unstable library feature 'command_access'`. Is this something we can allow? If not, I will close the PR. (Wonder why this didn't show up with cargo clippy locally...)

sharkdp:
`get_program` is only available from Rust 1.57: https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.get_program

We can bump the min. supported Rust version to 1.57. Just search for the current MSRV (1.56) to find all places where it needs updating (`.github/workflows/CICD.yml`, `clippy.toml`, and possible some READMEs)

themkat:
@sharkdp Ah! I forgot that it was not included in 1.56. Updated MSRV for the places I could find now 🙂

sharkdp (approved):

I believe we were previously (when we still used a plain `std::process::Command`) showing the full command line including arguments... But I guess it's not really needed for this (error) use case, correct? Or are there any cases where it would be useful to see the full command-line?

It might even be better not to see the full argument list, e.g. in case of `-X` when we potentially have thousands of them.

themkat:
@sharkdp , I don't really think all the arguments are necessary for this use case. The most important thing to know is just that the command was not found in my view.

src/exec/command.rs
```diff
@@ -96,7 +96,7 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
 pub fn handle_cmd_error(cmd: Option<&Command>, err: io::Error) -> ExitCode {
     match (cmd, err) {
         (Some(cmd), err) if err.kind() == io::ErrorKind::NotFound => {
-            print_error(format!("Command not found: {:?}", cmd));
+            print_error(format!("Command not found: {:?}", cmd.get_program()));
```
sharkdp, line 99: Minor: by using `{:?}`, we further rely on Debug outputs which are typically not guaranteed to be stable. Furthermore, `.get_program()` returns an `OsStr` and I'm not sure what it would do in case of invalid UTF-8 program names (which, agreed, is hopefully a non-existent corner case). But still, taking these two things together, I think I would prefer something more explicit like (untested):
```suggestion
            print_error(format!("Command not found: '{}'", cmd.get_program().to_string_lossy()));
```

themkat: Sure! That is a good idea 🙂  Did in 9c9097f

sharkdp:
Thank you!

PR #1118, opened 2022-09-26
RFC: --exec with placeholder corresponding to regex groups/ocurrences

juarezr:
### What

Allow the groups/occurrences matched by the regex pattern over the filename/path to be used as placeholders for commands executed with the flag `--exec`.

Note however that it's just a proof-of-concept yet. I just have written for checking if such a feature would become accepted. The code is far from done but at least it's possible to run and play with any supported regex and the placeholder support.

So, please share your opinion.

### Features

- Allow the use of regex group constructs for extracting specific pieces from the files/directories name.
- Allow capturing as many occurrences as matched by the regex pattern and use them as placeholders.
- Allow specifying a default text when the group/occurrence is not matched.

##### Renaming or moving files and folders:

``` bash
fd '(.+)( copy.*)(\.\w+)$' -x mv "{}" "{1}.{3}"
```

##### Adding prefixes, text, and extensions to file names:
  
``` bash
fd '([\w-]+)(\.[\w-]+)?$' -t f -x echo "|| {} |groups| {//}/{1}{2:-.ext}"
```

### Syntax

From the help:

```
-x, --exec <cmd>...
    Execute a command for each search result in parallel (use --threads=1 for sequential command execution). All positional
    arguments following --exec are considered to be arguments to the command - not to fd. It is therefore recommended to
    place the '-x'/'--exec' option last.
    The following placeholders are substituted before the command is executed:
      '{}':   path (of the current search result)
      '{/}':  basename
      '{//}': parent directory
      '{.}':  path without file extension
      '{/.}': basename without file extension
+     '{N}':  text matched by the N-th group in the first pattern occurrence. Text outside groups is discarded.
+     '{M.N}': text matched in the M-th pattern occurrence by the N-th group over the path or filename.
+     '{N:-D}': text matched by the optional occurrence/group or defaults to 'D' when not matched.
+   Obs:
+   - Using 0 for M/N substitutes by the text from all groups or all occurrences respectively.
    - If no placeholder is present, an implicit "{}" at the end is assumed.
 ```

#### Examples

``` bash
cargo run -- '(\w+)\.(\w+)$' -x echo "|| {} |groups| {//}/{2}.{1}" | column -t

cargo run -- '([a-zA-Z0-9]+)([\._-]\w+)\.(\w+)$' -x echo "|| {} |groups| {//}/{1}.{3}" | column -t

cargo run -- '([\w-]+)(\.[\w-]+)?$' -t f -x echo "|| {} |groups| {//}/{1}{2:-.ext}" | column -t

cargo run -- '([\./])([\w\.]+)' -p  -x echo "|| {} |matches| {1.0} {2.0} {3.0} {4.0} {5.0} {6.0} {7.0} {8.0} {9.0} {10.0}" | column -t

cargo run -- '(\w+)([_-]+\w+)(\.\w+)?' -x echo "|/| {/} |0| {0} {0.1} {0.2} {0.3} |1| {1.0} {1.1} {1.2} {1.3}" | column -t

cargo run -- '([-_\.]+)?([a-zA-Z0-9])([a-zA-Z0-9]+)' -x echo "|/| {/} |0| {0} {0.1} {0.2} {0.3} |1| {1.0} {1.1} {1.2} {1.3} |2| {2.0} {2.1} {2.2} {2.3}" | column -t

cargo run -- '(/)(\w+)(\.\w+)?' -p -a -x echo "|| {} |groups| {0.1} {0.2} {0.3}" | column -t
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -402,8 +402,11 @@ pub fn build_app() -> Command<'static> {
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
                        '{/.}': basename without file extension\n \
-                       '{<num>}': <num> is the regex group substitution index for the pattern matched \n\n\
-                     If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
+                       '{N}':  text matched by the N-th group in the first pattern ocurrence. Text outside groups are discarted.\n\
+                       '{M.N}':text matched in the M-th pattern ocurrence by the N-th group over the the path or filename.\n\n\
+                     Obs:\n
+                       - Using 0 for M/N substitutes by the text from all groups or all ocurrences respectively.\n\
+                       - If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
                      Examples:\n\n  \
                        - find all *.zip files and unzip them:\n\n      \
                            fd -e zip -x unzip\n\n  \
@@ -435,7 +438,6 @@ pub fn build_app() -> Command<'static> {
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
                        '{/.}': basename without file extension\n \
-                       '{<num>}': <num> is the regex group substitution index for the pattern matched \n\n\
                      If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
                      Examples:\n\n  \
                        - Find all test_*.py files and open them in your favorite editor:\n\n      \
--- a/src/app.rs
+++ b/src/app.rs
@@ -401,7 +401,8 @@ pub fn build_app() -> Command<'static> {
                        '{/}':  basename\n  \
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
-                       '{/.}': basename without file extension\n\n\
+                       '{/.}': basename without file extension\n \
+                       '{<num>}': <num> is the regex group substitution index for the pattern matched \n\n\
                      If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
                      Examples:\n\n  \
                        - find all *.zip files and unzip them:\n\n      \
@@ -410,6 +411,8 @@ pub fn build_app() -> Command<'static> {
                            fd -e h -e cpp -x clang-format -i\n\n  \
                        - Convert all *.jpg files to *.png files:\n\n      \
                            fd -e jpg -x convert {} {.}.png\
+                       - Rename all *.jpeg files with extension *.jpg:\n\n      \
+                           fd '(.+)\\.(jpg)$' -x mv {} {1}.{2}\
                     ",
                 ),
         )
@@ -431,7 +434,8 @@ pub fn build_app() -> Command<'static> {
                        '{/}':  basename\n  \
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
-                       '{/.}': basename without file extension\n\n\
+                       '{/.}': basename without file extension\n \
+                       '{<num>}': <num> is the regex group substitution index for the pattern matched \n\n\
                      If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
                      Examples:\n\n  \
                        - Find all test_*.py files and open them in your favorite editor:\n\n      \
--- a/src/app.rs
+++ b/src/app.rs
@@ -404,6 +404,7 @@ pub fn build_app() -> Command<'static> {
                        '{/.}': basename without file extension\n \
                        '{N}':  text matched by the N-th group in the first pattern ocurrence. Text outside groups are discarted.\n\
                        '{M.N}':text matched in the M-th pattern ocurrence by the N-th group over the the path or filename.\n\n\
+                       '{N:-D}': text matched by the optional ocurrence/group or defaults to 'D' when not matched.\n\
                      Obs:\n
                        - Using 0 for M/N substitutes by the text from all groups or all ocurrences respectively.\n\
                        - If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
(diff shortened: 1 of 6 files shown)


sharkdp:
> So, please share your opinion.

Thank you for your contribution and especially for sharing not just an implementation but also use cases and examples. To be frank, I'm rather skeptical. But certainly open to discuss this.



> ##### Renaming or moving files and folders:
> 
> ```shell
> fd '(.+)( copy.*)(\.\w+)$' -x mv "{}" "{1}.{3}"
> ```

I typically use [`rename`](https://github.com/util-linux/util-linux) for a job like this. Could even be combined with `fd`. Haven't tested this, but something similar to
```bash
fd ' copy' -X replace ' copy.*\.' '.'
```
should do the same job... in an (arguably) more descriptive way?

> fd '([\w-]+)(\.[\w-]+)?$' -t f -x echo "|| {} |groups| {//}/{1}{2:-.ext}"

> ##### Adding prefixes, text, and extensions to file names:
> 
> ```shell
> fd '([\w-]+)(\.[\w-]+)?$' -t f -x echo "|| {} |groups| {//}/{1}{2:-.ext}"
> ```

Not exactly sure what that does, but adding prefixes, suffixes, and different extensions should already be possible with the existing placeholders, no?

juarezr:
Thanks for considering to discuss it and sorry for my examples not being good enough to show the use cases.

> I typically use rename for a job like this. Could even be combined with fd. Haven't tested this, but something similar to
> fd ' copy' -X replace ' copy.*\.' '.'
> should do the same job... in an (arguably) more descriptive way?

You are right that there is alternatives for handling these use cases.
The feature I'm proposing is to have additional flexibility in `fd` for convenience.
What I'm looking for is to reduce the need of including another tool for solving the problem.

Also, the current tricks are still useful. :)

> Not exactly sure what that does, but adding prefixes, suffixes, and different extensions should already be possible with the existing placeholders, no?

Partially:
- Only before and after the whole filename placeholders: `{/}` (basename) and `{/.}` (basename).
- Only using fixed strings as `-x` parameters

But not being able to:
- Put anything directly in the middle of the filename.
- Replacing variable text in the middle of `base` and `name` parts of the filename
- Not exchanging positions of variable text
- Not choosing specific classes/slices of characters for deleting/transposing

See:

```bash
❱ fd 'fixed.*.txt' -x echo "rename fixed adequated {}"
rename fixed adequated ./files/a-fixed-name.txt
❱ cargo run -- '(.*)-(.*)-(.*).txt' -x echo "mv {} {1}-adequated-{3}.txt"
...
mv ./files/a-dynamic-title.txt a-adequated-title.txt
mv ./files/a-fixed-name.txt a-adequated-name.txt
```

For sure that would be possible to use additional tools to solve more complex and more dynamic cases like in this example.

For example, one can pipe `fd` with `awk` or `sed` and with `xargs` and with `mv` too for renaming very dynamic parts on a large list of files.

But having placeholders for substitution of regex matches in `fd` reduces the complexity of dealing with many more tools and their arguments. Also, fd gains much more power and flexibility for modifying any parts of filename and path as allowed by the possibilities of regex groups and matches.

I missed this when renaming some music files that had very messy file names. At that time I needed to invert the position of the artist with the song name, like in this example:

```bash
❱ cargo run -- '([^-\n]+)(?:(\s+-\s+)(.+))?(.mp3)' -t f -x echo "{}_{//}/{3:-Unknown} - {1}.mp3" | column -t -s '_'
...
./music/I Still Haven't Found What I'm Looking For - U2.mp3  ./music/U2 - I Still Haven't Found What I'm Looking For.mp3
./music/We Will Rock You - Queen.mp3                         ./music/Queen - We Will Rock You.mp3
./music/Angie - Rolling Stones.mp3                           ./music/Rolling Stones - Angie.mp3
./music/Time to Love.mp3                                     ./music/Unknown - Time to Love.mp3
./music/I Will Follow - U2.mp3                               ./music/U2 - I Will Follow.mp3
./music/Sunday Bloody Sunday - U2.mp3                        ./music/U2 - Sunday Bloody Sunday.mp3
./music/We Are the Champions - Queen.mp3                     ./music/Queen - We Are the Champions.mp3
./music/Brown Sugar - Rolling Stones.mp3                     ./music/Rolling Stones - Brown Sugar.mp3
./music/Desire - U2.mp3                                      ./music/U2 - Desire.mp3
./music/One - U2.mp3                                         ./music/U2 - One.mp3
./music/I Want to Break Free - Queen.mp3                     ./music/Queen - I Want to Break Free.mp3
./music/Satisfaction - Rolling Stones.mp3                    ./music/Rolling Stones - Satisfaction.mp3
./music/With or Without You - U2.mp3                         ./music/U2 - With or Without You.mp3
```

I hope that these examples let the motivations become clearer. :)

cclifford:
Hi, I just wanted to express my interest in a feature like this being included in fd. I use fd pretty heavily to select data files from a collection of files for processing. Often we store metadata as path components within our data repository (for example, what sensor it originates from). various scripts we use require this information, so we pass it in with commandline flags. Right now I use short inline bash scripts to parse the information out of the file path and pass it to the scripts, but this is pretty fragile and gross. 

Anyway, I ran into this older PR researching to see if I could implement a similar feature. I though I would comment because I have a use case that is not file renaming.

dsummersl:
This would be an awesome feature - every now and then I need to make a rename or move based on some pattern within files or directories. This would make such changes kinda trivial/magic!

juarezr:
Old implementation [here](https://github.com/juarezr/fd/tree/exec_with_regex_placeholders_old)

juarezr:
> i just found this PR because this feature is precisely what i'm missing right now and having it would be a life-saver 👍🏼

I've found the [Rnr](https://github.com/ismaelgv/rnr) tool, which filled the gap for my use case of renaming files.
However, if you have a use case that isn't batch renaming files, I believe it won't help.

PR #1130, opened 2022-10-07
Fix `--min-depth` for broken symlinks

### LINKED ISSUE #1017: --min-depth is broken for broken symlinks
```console
foo/bar/baz/link
tavianator@tachyon $
```

Because:

theirix:
Fixes #1017 .
Instead of computing depth for broken symlink we can just ensure the directory entry passes the minimal depth check. The `DirEntry::depth` function is not needed anymore so it has been removed.

### DIFF
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -83,11 +83,22 @@ impl DirEntry {
     }
 
     /// Returns true if the entry has the depth greater or equal compared to `depth`
-    pub fn deeper(&self, depth: usize) -> bool {
+    pub fn deeper(&self, depth: usize, root_paths: &[PathBuf]) -> bool {
         match &self.inner {
             DirEntryInner::Normal(e) => e.depth() >= depth,
-            // broken symlink passes match if it is just longer thatn the `depth`
-            DirEntryInner::BrokenSymlink(p) => p.components().count() - 1 >= depth,
+            DirEntryInner::BrokenSymlink(path) => {
+                for root_path in root_paths {
+                    // Detect which root path originates to current `path`
+                    if path.starts_with(root_path) {
+                        let root_depth = root_path.components().count();
+                        let entry_depth = path.components().count();
+                        if entry_depth >= root_depth + depth {
+                            return true;
+                        }
+                    }
+                }
+                false
+            }
         }
     }
 }
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -82,10 +82,12 @@ impl DirEntry {
             .as_ref()
     }
 
-    pub fn depth(&self) -> Option<usize> {
+    /// Returns true if the entry has the depth greater or equal compared to `depth`
+    pub fn deeper(&self, depth: usize) -> bool {
         match &self.inner {
-            DirEntryInner::Normal(e) => Some(e.depth()),
-            DirEntryInner::BrokenSymlink(_) => None,
+            DirEntryInner::Normal(e) => e.depth() >= depth,
+            // broken symlink passes match if it is just longer thatn the `depth`
+            DirEntryInner::BrokenSymlink(p) => p.components().count() - 1 >= depth,
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fd returns an error when current working directory does not exist while a search path is
   specified, see #1072 (@vijfhoek)
 - Improved "command not found" error message, see #1083 and #1109 (@themkat)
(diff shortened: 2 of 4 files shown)


src/dir_entry.rs
```diff
@@ -82,10 +82,12 @@ impl DirEntry {
...
 
-    pub fn depth(&self) -> Option<usize> {
+    /// Returns true if the entry has the depth greater or equal compared to `depth`
+    pub fn deeper(&self, depth: usize) -> bool {
         match &self.inner {
-            DirEntryInner::Normal(e) => Some(e.depth()),
-            DirEntryInner::BrokenSymlink(_) => None,
+            DirEntryInner::Normal(e) => e.depth() >= depth,
+            // broken symlink passes match if it is just longer thatn the `depth`
+            DirEntryInner::BrokenSymlink(p) => p.components().count() - 1 >= depth,
```
tavianator, line 90: What if the starting directory has multiple complements?

theirix: Yes, the result depends on the size of the root component too so my approach wouldn't work.

It will erroneously match the `link`
    
    fd -L link foo/bar/baz --min-depth 3

while the symmetric call for non-link would exclude match:

    mkdir -p foo/bar/baz/fourth
    fd -L fourth foo/bar/baz --min-depth 3

`ignore` crate hides too much information so it's problematic to extract root.

tavianator (changes requested):
Needs a different approach for starting paths with multiple components

InfinityGlitch:
You could take the depth from the `ignore::Error::depth()` function and store it the `DirEntryInner::BrokenSymlink` variant when constructing a broken symlink `DirEntry`.

theirix:
> 



> You could take the depth from the `ignore::Error::depth()` function and store it the `DirEntryInner::BrokenSymlink` variant when constructing a broken symlink `DirEntry`.

Thanks for the idea, I'll try it. I had tried something similar before resorting to simpler (and erroneous!) variant in this PR.

theirix:
Unfortunately, using `ignore::Error::depth()` does not help because ignore creates a variant `Error::Io` for broken symlinks which does not have depth (`None` returned). Check this example for a broken link `link`:
```
box: Err(WithPath { path: "foo/bar/baz/link", err: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" }) })
```

Another approach may be to detect a root depth [here](https://github.com/sharkdp/fd/blob/master/src/walk.rs#L407) but it breaks for multi-threaded walkers.

So the simplest approach is to detect from which root path (could be multiple ones) is a parent for entry with symlink and to calculate a difference.

src/dir_entry.rs
```diff
@@ -83,11 +83,22 @@ impl DirEntry {
...
 
     /// Returns true if the entry has the depth greater or equal compared to `depth`
-    pub fn deeper(&self, depth: usize) -> bool {
+    pub fn deeper(&self, depth: usize, root_paths: &[PathBuf]) -> bool {
         match &self.inner {
             DirEntryInner::Normal(e) => e.depth() >= depth,
-            // broken symlink passes match if it is just longer thatn the `depth`
-            DirEntryInner::BrokenSymlink(p) => p.components().count() - 1 >= depth,
+            DirEntryInner::BrokenSymlink(path) => {
+                for root_path in root_paths {
```
tavianator, line 90: I think this is wrong for things like `fd . foo foo/bar`

tavianator (changes requested):

src/dir_entry.rs
theirix: What could go wrong, roots nesting in each other? Searching for a dir (`fd -L fourth foo foo/bar` ) and a broken link (`fd -L broken foo foo/bar`) works the same way for layout like
```foo/
foo/bar/
foo/bar/baz/
foo/bar/baz/fourth/
foo/bar/baz/broken
```

theirix: Thanks for the tip, I forgot to specify depth.

That's right, current implementation will output more than one entry for this configuration. At least, it is not empty output.

theirix: I have rethought the PR a little, according to the latest `ignore` and `fd` changes. I think that correct depth tracking with `WalkParallel` invoked with multiple paths is impossible. If only one thread in the walker exists, one can track depth with a root `DirEntry`. The best way to solve this is to have a "root" entry passed by the 'ignore' crate to each `FnVisitor` call alongside a `DirEntry`. It's unlikely to change `ignore` crate in this way since it's a breaking API change.

So far, PR solves the symlink problem with the best effort, and I cannot come up with a better idea.

sharkdp:

PR #1137, opened 2022-10-13
Respect exit codes with `--exec-batch`

### LINKED ISSUE #1136: Regression: `--exec-batch` no longer preserves status code
Suppose you have a `fd` command targeting at least one file, e.g. a directory `/dir` containing a file `foo`. Then running
```bash
fd foo /dir -X bash -c 'exit 1'
```
succeeds (based on exit code) in 8.4.0, but fails in e.g. 8.3.2. I would expect the latter to be the correct behavior (since #477).

Bisecting shows that #960 (specifically 9fb0c5d372062d6193798519526cb7b14ea24fcc) caused this regression; happy to open a PR if desired.

amesgen:
Closes #1136

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -16,7 +16,7 @@ use argmax::Command;
 use once_cell::sync::Lazy;
 use regex::Regex;
 
-use crate::exit_codes::ExitCode;
+use crate::exit_codes::{merge_exitcodes, ExitCode};
 
 use self::command::{execute_commands, handle_cmd_error};
 use self::input::{basename, dirname, remove_extension};
@@ -120,7 +120,7 @@ impl CommandSet {
                     }
                 }
 
-                ExitCode::Success
+                merge_exitcodes(builders.iter().map(|b| b.exit_code()))
             }
             Err(e) => handle_cmd_error(None, e),
         }
@@ -136,6 +136,7 @@ struct CommandBuilder {
     cmd: Command,
     count: usize,
     limit: usize,
+    exit_code: ExitCode,
 }
 
 impl CommandBuilder {
@@ -163,6 +164,7 @@ impl CommandBuilder {
             cmd,
             count: 0,
             limit,
+            exit_code: ExitCode::Success,
         })
     }
 
@@ -196,13 +198,19 @@ impl CommandBuilder {
     fn finish(&mut self) -> io::Result<()> {
         if self.count > 0 {
             self.cmd.try_args(&self.post_args)?;
-            self.cmd.status()?;
+            if !self.cmd.status()?.success() {
+                self.exit_code = ExitCode::GeneralError;
+            }
 
             self.cmd = Self::new_command(&self.pre_args)?;
             self.count = 0;
         }
 
         Ok(())
+    }
+
+    fn exit_code(&self) -> ExitCode {
+        self.exit_code
     }
 }
 
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1496,6 +1496,8 @@ fn test_exec_batch() {
             &["foo", "--exec-batch", "echo {}"],
             "[fd error]: First argument of exec-batch is expected to be a fixed executable",
         );
+
+        te.assert_failure_with_error(&["a.foo", "--exec-batch", "bash", "-c", "exit 1"], "");
     }
 }
 
@@ -1550,6 +1552,20 @@ fn test_exec_batch_multi() {
                 "directory_foo"
             ],
         ]
+    );
+
+    te.assert_failure_with_error(
+        &[
+            "a.foo",
+            "--exec-batch",
+            "echo",
+            ";",
+            "--exec-batch",
+            "bash",
+            "-c",
+            "exit 1",
+        ],
+        "",
     );
 }
 


src/exec/mod.rs
```diff
@@ -196,14 +198,20 @@ impl CommandBuilder {
     fn finish(&mut self) -> io::Result<()> {
         if self.count > 0 {
             self.cmd.try_args(&self.post_args)?;
-            self.cmd.status()?;
+            if !self.cmd.status()?.success() {
+                self.exit_code = ExitCode::GeneralError;
```
tmccombs, line 202: Rather than storing an exit code here, I think we should return an `Err`.

tmccombs (changes requested):

src/exec/mod.rs
amesgen: Hmm, my idea went like this: We *do not* want to short-circuit on non-success exit codes (compare `-x` semantics), so we were to return an `Err`, we would have to catch it (but just this one, not other `Err`s) in `execute_batch` and still maintain sth like this `exit_code` there. But this seemed just more convoluted, so I took this approach.

Happy to change though if you have a different idea :+1:

tmccombs: The way `-x` works is if a command fails, we short circuit for that particular path.

So for example if you have `fd foo -x c1  \; -x c2`   and c1 fails for a specific path, then c2 won't be run for that path. But it will continue trying to run c1 for other paths 

I think we would probably want similar behavior for exec-batch. If one command fails, we probably don't want to run the next one in the series, because it might depend on the first one succeeding.

However, maybe we do want to run the commandset with the next batch of paths if we have more paths than can fit in a single command invocation?

I created #1138  which will short circuit.

src/exec/mod.rs
```diff
@@ -196,14 +198,20 @@ impl CommandBuilder {
...
+            }
 
             self.cmd = Self::new_command(&self.pre_args)?;
             self.count = 0;
```
tmccombs, line 206: If we do go with this approach, then I think we might need to reset the exit code here

src/exec/mod.rs
amesgen: I was actually only thinking about batching in my comment above, thanks for mentioning multiple commands.

> However, maybe we do want to run the commandset with the next batch of paths if we have more paths than can fit in a single command invocation?

Yes, I think that we should run either all or no batches of any given command, to be consistent with `-x` (where one can be sure that a command is either executed for all files or for none).

---

So in short, it seems reasonable to have short circuiting for mulitple commands, but no short circuiting for multiple batches of the same command.

I can update this PR if you agree, but if short circuiting batches is prefered, then #1138 seems nicer :+1:

tmccombs: I don't have a very strong preference on  whether or not we short circuit for batches for a single command. @sharkdp do you have any thoughts on it?

tavianator: `find` doesn't short circuit in this case, and I think that's the right behaviour.  Short circuiting is confusing since the commands don't necessarily get the same arguments.

sharkdp: Okay, to summarize (mostly for me to make sure I understand it correctly):

Let's say we execute
```
fd --batch-size=2 -X cmdA \; -X cmdB
```
and there are five search results `1`, `2`, …, `5`. Then if everything goes well (both commands always succeed), we run the following commands (modulo ordering):
```
cmdA ./1 ./2
cmdB ./1 ./2
cmdA ./3 ./4
cmdB ./3 ./4
cmdA ./5
cmdB ./5
```

Now the question is what happens if `cmdA` fails for the first batch (`cmdA ./1 ./2`). There are three reasonable options:
1. Quit immediately.
2. Only continue to run `cmdA` on the remaining batches (`cmdA ./3 ./4` and `cmdA ./5`). Do not run `cmdB` on any batch.
3. Still run everything as if nothing happened.

> I think we would probably want similar behavior for exec-batch. If one command fails, we probably don't want to run the next one in the series, because it might depend on the first one succeeding.
> 
> However, maybe we do want to run the commandset with the next batch of paths if we have more paths than can fit in a single command invocation?

@tmccombs I understand you would like to advocate for option 2?

> `find` doesn't short circuit in this case, and I think that's the right behaviour. Short circuiting is confusing since the commands don't necessarily get the same arguments.

@tavianator It wasn't clear to me what you meant by "in this case". But if my experiments are correct, `find` does not short-circuit at all, neither with respect to batches, nor with respects to follow-up commands. This means `find` implements option 3.


If my interpretation is correct, I'm also voting for `find`-compatibility with option 3. However, I'm not sure I understand your reasoning here:

> Short circuiting is confusing since the commands don't necessarily get the same arguments.

tavianator: > @tavianator It wasn't clear to me what you meant by "in this case"

I just meant that `find -exec ... {} +` does no short circuiting, as opposed to `find -exec ... {} \;`

> If my interpretation is correct, I'm also voting for find-compatibility with option 3. However, I'm not sure I understand your reasoning here:
>
> > Short circuiting is confusing since the commands don't necessarily get the same arguments.

What I mean is that in general, you cannot assume `cmdA` has run for a particular file before `cmdB`, as the batches may end up different due to command line length limits.  E.g. you may get

```
cmdB ./1 ./2 ./3
cmdA ./1 ./2 ./3 ./4
cmdA ./5
cmdB ./4 ./5
```

or something, especially if `cmdB` is longer than `cmdA`.

Since you can't rely on it in general, I don't think we should offer it on a best-effort basis either, so that people don't rely on it accidentally.

tmccombs: Ah.... I didn't think about that. Now I think 3 might be the right way to go, at least for batches. 

Or possibly we change the way batches work, and we run all batches for each command before moving on to the next one, but that could hurt performance.

tavianator: > Or possibly we change the way batches work, and we run all batches for each command before moving on to the next one, but that could hurt performance.

That's not a bad idea, but it's probably material for another PR :).  I don't think the perf impact would be too bad

sharkdp: Okay. So how do we proceed with this PR? I think it would be great to add some tests for the batch-behavior as well. We can use `--batch-size=N` for that.

amesgen: So I understand it correctly that the semantics of this PR (no short circuiting at all, i.e. @sharkdp option 3.) are good for now? In that case, I will add a few tests ensuring that all commands are executed even if only some commands/batches fail :+1:

sharkdp: @amesgen That would be great

sharkdp:
I'm going to merge this as-is, because it's definitely better than what we have on master. If you feel like adding those tests later @amesgen, that would be very much appreciated. But please don't feel obliged.

PR #1138, opened 2022-10-15
Propagate failed status code from subprocess with exec-batch

tmccombs:
If a child process returns a failure status code for --exec-batch, then have fd also return a failed status code.

Closes: #1136


sharkdp:
Thank you. If I understand correctly, this can be closed after the discussion in #1137, right?

Issue #315, opened 2018-07-31
Finding multiple patterns


tjkirch:
I'd like to be able to search for multiple patterns, like with grep's -e argument.  It seems (with fd 7.0.0) the only way is to use alternation in the regex pattern, but this can be less clear than multiple arguments, and is harder to build up programmatically.

sharkdp:
Thank you for the feedback!

I certainly see the need for this, but I'm not sure we should introduce a new command-line argument, given that there is a reasonable solution via `fd '(pattern1|pattern2)'`. On the other hand, the analogy to `grep` would be nice. Unfortunately, `-e` is already taken for `--extension`.

Another option to achieve something like this could be the `--path-before-pattern` flag that we were discussing over in #312.

sharkdp:
Actually, the `--path-before-pattern` doesn't feel very natural for me. I'd be okay with adding a new `--regexp <PATTERN>` option in analogy to `grep`/`rg`, if someone wants to work on this.

sharkdp:
I'm currently not planning to implement this. Going to close this for now, but happy to reconsider if there is a significant interest in this.

aschmolck:
How about 

```
fd Makefile --or GNuMakefile --or make
```
?

That reads naturally, and it would make it possible to add a git grep or find style boolean query language at some point.

sharkdp:
Let's reopen this for further discussion.

aschmolck:
Here is a concrete example I just did with `find`, I think it would be nice to be able to do the same thing with fd as well:

```
time find . -type d -and \( -name node_modules -or -name build \) -exec rm -rf '{}' '+'
```

sharkdp:
I'm definitely against including a full-blown query language with `--and`/`--or`. `fd` was never designed to be this powerful. It's focused on easier use-cases.

Your use-case can be solved by running
```
fd -td '^(node_modules|build)$' -X rm -rf
```
or
```
fd -td node_modules -X rm -rf
fd -td build -X rm -rf
```

Both of which are shorter than the find equivalent (which is not the main issue here though).

zw963:
> I'm definitely against including a full-blown query language with `--and`/`--or`. `fd` was never designed to be this powerful. It's focused on easier use-cases.
> 
> Your use-case can be solved by running
> 
> ```
> fd -td '^(node_modules|build)$' -X rm -rf
> ```
> 
> or
> 
> ```
> fd -td node_modules -X rm -rf
[...]

PR #1139, opened 2022-10-15
add support for matching multiple patterns

Uthar:
Added the ~`--and`~ ~`--expr`~ `--and` flags that search for matches containing all the patterns, similar to the `-and` flag of GNU find.

https://github.com/minad/consult/issues/6
#315

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -144,6 +144,17 @@ pub struct Opts {
     )]
     pub fixed_strings: bool,
 
+    /// Additional search patterns that need to be matched
+    #[arg(
+        long = "and",
+        value_name = "pattern",
+        long_help = "Add additional required search patterns, all of which must be matched. Multiple \
+                     additional patterns can be specified. The patterns are regular expressions, \
+                     unless '--glob' or '--fixed-strings' is used.",
+        hide_short_help = true
+    )]
+    pub exprs: Option<Vec<String>>,
+
     /// Show absolute instead of relative paths
     #[arg(
         long,
@@ -499,17 +510,6 @@ pub struct Opts {
                  pass '--' first, or it will be considered as a flag (fd -- '-foo')."
     )]
     pub pattern: String,
-
-    /// Additional search patterns that need to be matched
-    #[arg(
-        long = "and",
-        value_name = "pattern",
-        long_help = "Add additional required search patterns, all of which must be matched. Multiple \
-                     additional patterns can be specified. The patterns are regular expressions, \
-                     unless '--glob' or '--fixed-strings' is used.",
-        hide_short_help = true
-    )]
-    pub exprs: Option<Vec<String>>,
 
     /// Set path separator when printing file paths
     #[arg(
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -500,6 +500,17 @@ pub struct Opts {
     )]
     pub pattern: String,
 
+    /// Additional search patterns that need to be matched
+    #[arg(
+        long = "and",
+        value_name = "pattern",
+        long_help = "Add additional required search patterns, all of which must be fulfilled. Multiple \
+                     additional patterns can be specified. The patterns are regular expressions,
+                     unless '--glob' or '--fixed-strings' is used.",
+        hide_short_help = true
+    )]
+    pub exprs: Option<Vec<String>>,
+
     /// Set path separator when printing file paths
     #[arg(
         long,
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -504,8 +504,8 @@ pub struct Opts {
     #[arg(
         long = "and",
         value_name = "pattern",
-        long_help = "Add additional required search patterns, all of which must be fulfilled. Multiple \
-                     additional patterns can be specified. The patterns are regular expressions,
+        long_help = "Add additional required search patterns, all of which must be matched. Multiple \
+                     additional patterns can be specified. The patterns are regular expressions, \
                      unless '--glob' or '--fixed-strings' is used.",
         hide_short_help = true
     )]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -151,7 +151,8 @@ pub struct Opts {
         long_help = "Add additional required search patterns, all of which must be matched. Multiple \
                      additional patterns can be specified. The patterns are regular expressions, \
                      unless '--glob' or '--fixed-strings' is used.",
-        hide_short_help = true
+        hide_short_help = true,
+        allow_hyphen_values = true
     )]
     pub exprs: Option<Vec<String>>,
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,7 @@ use atty::Stream;
 use clap::{CommandFactory, Parser};
 use globset::GlobBuilder;
 use lscolors::LsColors;
-use regex::bytes::{RegexBuilder, RegexSetBuilder};
+use regex::bytes::{Regex, RegexBuilder, RegexSetBuilder};
 
 use crate::cli::{ColorWhen, Opts};
 use crate::config::Config;
@@ -81,12 +81,28 @@ fn run() -> Result<ExitCode> {
     }
 
     ensure_search_pattern_is_not_a_path(&opts)?;
-    let pattern_regex = build_pattern_regex(&opts)?;
+    let pattern = &opts.pattern;
+    let exprs = &opts.exprs;
+    let empty = Vec::new();
 
-    let config = construct_config(opts, &pattern_regex)?;
-    ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regex)?;
-    let re = build_regex(pattern_regex, &config)?;
-    walk::scan(&search_paths, Arc::new(re), Arc::new(config))
+    let pattern_regexps = exprs
+        .as_ref()
+        .unwrap_or(&empty)
+        .iter()
+        .chain([pattern])
+        .map(|pat| build_pattern_regex(pat, &opts))
+        .collect::<Result<Vec<String>>>()?;
+
+    let config = construct_config(opts, &pattern_regexps)?;
+
+    ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regexps)?;
+
+    let regexps = pattern_regexps
+        .into_iter()
+        .map(|pat| build_regex(pat, &config))
+        .collect::<Result<Vec<Regex>>>()?;
+
+    walk::scan(&search_paths, Arc::new(regexps), Arc::new(config))
 }
 
 #[cfg(feature = "completions")]
@@ -145,8 +161,7 @@ fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
     }
 }
 
-fn build_pattern_regex(opts: &Opts) -> Result<String> {
-    let pattern = &opts.pattern;
+fn build_pattern_regex(pattern: &str, opts: &Opts) -> Result<String> {
     Ok(if opts.glob && !pattern.is_empty() {
         let glob = GlobBuilder::new(pattern).literal_separator(true).build()?;
         glob.regex().to_owned()
@@ -172,11 +187,14 @@ fn check_path_separator_length(path_separator: Option<&str>) -> Result<()> {
     }
 }
 
-fn construct_config(mut opts: Opts, pattern_regex: &str) -> Result<Config> {
+fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config> {
     // The search will be case-sensitive if the command line flag is set or
-    // if the pattern has an uppercase character (smart case).
-    let case_sensitive =
-        !opts.ignore_case && (opts.case_sensitive || pattern_has_uppercase_char(pattern_regex));
+    // if any of the patterns has an uppercase character (smart case).
+    let case_sensitive = !opts.ignore_case
+        && (opts.case_sensitive
+            || pattern_regexps
+                .iter()
+                .any(|pat| pattern_has_uppercase_char(pat)));
 
     let path_separator = opts
         .path_separator
@@ -415,14 +433,18 @@ fn extract_time_constraints(opts: &Opts) -> Result<Vec<TimeFilter>> {
 
 fn ensure_use_hidden_option_for_leading_dot_pattern(
     config: &Config,
-    pattern_regex: &str,
+    pattern_regexps: &[String],
 ) -> Result<()> {
-    if cfg!(unix) && config.ignore_hidden && pattern_matches_strings_with_leading_dot(pattern_regex)
+    if cfg!(unix)
+        && config.ignore_hidden
+        && pattern_regexps
+            .iter()
+            .any(|pat| pattern_matches_strings_with_leading_dot(pat))
     {
         Err(anyhow!(
-            "The pattern seems to only match files with a leading dot, but hidden files are \
+            "The pattern(s) seems to only match files with a leading dot, but hidden files are \
             filtered by default. Consider adding -H/--hidden to search hidden files as well \
-            or adjust your search pattern."
+            or adjust your search pattern(s)."
         ))
     } else {
         Ok(())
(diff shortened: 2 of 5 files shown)


sharkdp:

Before we consider the integration of this new feature, please see my comment in https://github.com/sharkdp/fd/issues/315#issuecomment-968323590:

> Ok, I'm inclined to accept a feature request to support `--and <pattern>`. Before we implement this, we need:
> 
> - A short discussion about the command-line option name. What do other tools use?
> - A detailed analysis if this could clash with any of the other command-line options or features of `fd`. There are some immediate questions like: what does `fd patternA --and patternB --type f` mean? (we are not going to support the meaning *patternA AND (patternB AND type==file)*).

Ideally, the second point should be discussed in detail and represented in corresponding unit tests.

Uthar:
> A short discussion about the command-line option name. What do other tools use?

- ripgrep uses `-e`

```
-e, --regexp <PATTERN>...                    
  A pattern to search for. This option can be provided multiple times, where
  all patterns given are searched. Lines matching at least one of the provided
  patterns are printed. This flag can also be used when searching for patterns
  that start with a dash.
  
  For example, to search for the literal '-foo', you can use this flag:
  
      rg -e -foo
  
  You can also use the special '--' delimiter to indicate that no more flags
  will be provided. Namely, the following is equivalent to the above:
  
      rg -- -foo

```
- GNU find uses the following syntax:

```
expr1 expr2
       Two expressions in a row are taken to be joined with an implied -a; expr2 is not evaluated if expr1 is false.

expr1 -a expr2
       Same as expr1 expr2.

expr1 -and expr2
       Same as expr1 expr2, but not POSIX compliant.
```

Uthar:
> what does fd patternA --and patternB --type f mean?

I think the `--and` flag would simply mean "Find all files whose names match all patterns specified by the initial [pattern] and all the --and flags"

No such nesting like you described would be possible

I don't know all of fd's flags, but it looks to me like it will compose well, being just an additional filter for names.

> unit tests

I created some basic unit tests. What other cases should be covered?

sharkdp:
> I think the `--and` flag would simply mean "Find all files whose names match all patterns specified by the initial [pattern] and all the --and flags"

Sounds reasonable. I would still like to see some thought going into potential conflicts (or chances for combination) with other command line options - like `--type` in the example quoted above. Other potentially interesting options to look at:
* `--glob`/`--fixed-strings`
* `--ignore-case`/`--case-sensitive`
* `--full-path`
* `--extension`/`--type`/`--size`/etc.


> a broken pattern in any of the --and flags should signal an error

That is definitely worth checking.

Another thing to potentially test would be patterns starting with a dash. Usually, we solve this using the `--` separator: `fd -- '-foo'`. Not sure if we could make it work for `--and` as well. Might not even require the `--` separator, but maybe a special clap setting to allow the value to start with a dash.

Uthar:
> patterns starting with a dash

One could also use `^[-]`

Uthar:
Hmm you're right, maybe an `--expr`, stolen from ripgrep, would sound better?

Uthar:
Wrote tests for all these in new commit:

> --glob

`--and` patterns contribute to the search as globs, not regexps


> --fixed-strings

`--and` patterns are matched as whole strings, not regexp


> --ignore-case

case is ignored in `--and` patterns as well as the primary pattern


> --case-sensitive

case is sensitive in `--and` patterns as well as the primary pattern


> --full-path

`--and` patterns contribute to the search in the whole path, not just file or dir name


> --extension

`--extension` and `--and` patterns are composed in a logical way, i.e. patterns match file names via AND and extension matches extension via OR


> --type

`--type` additionally restricts the search done by all the `--and` patterns 


> --size

didn't check, I didn't know how

also added these from my post:

> test with empty patterns
> test with unreasonable pattern should match nothing
> a broken pattern in any of the --and flags should signal an error


tests fails for non-ubuntu CI... let me check...

Uthar:
> tests fails for non-ubuntu CI... let me check...

Can't make sense out of it. Is there some Mac and Windows experts on board?

tmccombs:
We probably want to clearly document how `--and` interacts with other options. The behaviors described above makes sense to me and are probably what I would expect in the absence of documentation, but I could easily see someone else expecting different behavior. 

Regarding hyphens. We probably want to use the [`allow_hyphen_values`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.allow_hyphen_values) option on the argument.

sharkdp:
@Uthar Thank you very much for the updates! Unfortunately, there are some larger merge conflicts now that I merged a few other PRs, especially #1067. I hope you can sort this out, otherwise please let us know.

Uthar:
Yes, I will attempt to solve this, stay tuned

Uthar:
Damn, looks like I will have to write it from scratch
It will take some time, I haven't written any rust code before

tmccombs:
I'm willing to help if you would like. After all, I made the PR that caused most of the conflicts.

Uthar:
Very nice of you tmccombs. My main problem is to write the code again, it takes a lot of time especially debugging the type errors with borrow and so on, because I lack experience.  I just don't have the time right now

Ultimately we just have to change `pattern` to `patterns` in this line:

https://github.com/sharkdp/fd/blob/master/src/walk.rs#L153

Everything else is just argument parsing/changing parameters to functions

Uthar:
I started some work here, the easy stuff :-) https://github.com/Uthar/fd/tree/conflicts

Uthar:
Ok, I've got this working now... but there are some issues that Rust experienced programmer will probably easily solve:

I'm trying to push the "initial" pattern to the optional vector of "and" patterns, but can't do that without borrowing, do you know how to handle it?

Also its panicking with no --expr flag currently, maybe unwrap() is not the best there?

I also don't know how to exit the program from the match .. Ok .. Err clause, when there is an error, it still requires a regex type.

Could you help me with these issues? See this commit: https://github.com/Uthar/fd/commit/d66470a50b05a1eb176d5ce37a2e6f89d32171a1

Uthar:
Full diff here https://github.com/sharkdp/fd/compare/master...Uthar:fd:conflicts

Uthar:
Done after all, we're back in the game :-)

But you should review the changes

sharkdp:
Thank you. The tests are currently failing on Windows and macOS

Uthar:
yeah... the failuress are really weird, dont make sense...

My friend has a macbook, I will ask them to try on it, also try to find a windows machine

Uthar:
fixed the tests

src/main.rs
```diff
@@ -166,11 +192,14 @@ fn check_path_separator_length(path_separator: Option<&str>) -> Result<()> {
     }
 }
 
-fn construct_config(mut opts: Opts, pattern_regex: &str) -> Result<Config> {
+fn construct_config(mut opts: Opts, pattern_regexps: &Vec<String>) -> Result<Config> {
```
tmccombs, line 195: This should probably be:

```suggestion
fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config> {
```
(note that a `Vec<String>` will implicitly dereference to a `&[String]`, so `&pattern_regexps` would still fit the type)

We could almost get away with just accepting an iterator, except that we would need to be able to iterate over it it twice in order to determine if we should be case sensitive or not if the options aren't expllicit.

src/main.rs
```diff
@@ -409,9 +438,13 @@ fn extract_time_constraints(opts: &Opts) -> Result<Vec<TimeFilter>> {
 
 fn ensure_use_hidden_option_for_leading_dot_pattern(
     config: &Config,
-    pattern_regex: &str,
+    pattern_regexps: &Vec<String>,
```
tmccombs, line 441: Similaraly this should be `&[String]` or maybe `&[&str]`

tmccombs (changes requested):

Uthar:
Thank you, tmccombs, for the thorough review.

See the last commit, where I added your solution to avoid `match` and `panic!` during generation of the regexp and changed type hints as you requested.

Let me know if you have some other improvement ideas.

src/cli.rs
```diff
@@ -523,6 +523,15 @@ pub struct Opts {
     )]
     pub pattern: String,
 
+    /// Additional search patterns to match a'la AND (a regular expression, unless '--glob' is used; optional)
+    #[arg(
+        long = "expr",
+        value_name = "pattern",
+        long_help = "(Additionally) filter search results by all patterns via AND. Multiple \
+                     additionalpatterns can be specified."
+    )]
```
sharkdp, line 532: I would appreciate if we keep this out of the short help text for now (`hide_short_help = true`).

src/cli.rs
```diff
@@ -523,6 +523,15 @@ pub struct Opts {
     )]
     pub pattern: String,
 
+    /// Additional search patterns to match a'la AND (a regular expression, unless '--glob' is used; optional)
+    #[arg(
+        long = "expr",
+        value_name = "pattern",
+        long_help = "(Additionally) filter search results by all patterns via AND. Multiple \
+                     additionalpatterns can be specified."
```
sharkdp, line 531: Maybe
```suggestion
        long_help = "Add additional required search patterns, all of which must be fulfilled. Multiple \
                     additional patterns can be specified. The pattern is a regular expression, unless '--glob' \
                     or '--fixed-strings' is used."
```

src/cli.rs
```diff
@@ -523,6 +523,15 @@ pub struct Opts {
     )]
     pub pattern: String,
 
+    /// Additional search patterns to match a'la AND (a regular expression, unless '--glob' is used; optional)
```
sharkdp, line 526:
```suggestion
    /// Additional search patterns that need to be matched
```

src/main.rs
```diff
@@ -81,12 +81,28 @@ fn run() -> Result<ExitCode> {
...
 
     ensure_search_pattern_is_not_a_path(&opts)?;
-    let pattern_regex = build_pattern_regex(&opts)?;
+    let pattern = &opts.pattern;
+    let exprs = &opts.exprs;
+    let empty = Vec::new();
+
+    let pattern_regexps = exprs
+        .as_ref()
+        .unwrap_or(&empty)
```
sharkdp, line 90:
```suggestion
        .unwrap_or_else(|| vec![])
```

src/main.rs
```diff
@@ -409,9 +427,13 @@ fn extract_time_constraints(opts: &Opts) -> Result<Vec<TimeFilter>> {
...
     config: &Config,
-    pattern_regex: &str,
+    pattern_regexps: &[String],
 ) -> Result<()> {
-    if cfg!(unix) && config.ignore_hidden && pattern_matches_strings_with_leading_dot(pattern_regex)
+    if cfg!(unix)
+        && config.ignore_hidden
+        && pattern_regexps
+            .iter()
+            .any(|pat| pattern_matches_strings_with_leading_dot(pat))
```
sharkdp, line 436: does this work?

```suggestion
            .any(pattern_matches_strings_with_leading_dot)
```

src/main.rs
```diff
@@ -166,11 +181,14 @@ fn check_path_separator_length(path_separator: Option<&str>) -> Result<()> {
...
+fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config> {
     // The search will be case-sensitive if the command line flag is set or
     // if the pattern has an uppercase character (smart case).
-    let case_sensitive =
-        !opts.ignore_case && (opts.case_sensitive || pattern_has_uppercase_char(pattern_regex));
+    let case_sensitive = !opts.ignore_case
+        && (opts.case_sensitive
+            || pattern_regexps
+                .iter()
+                .any(|pat| pattern_has_uppercase_char(pat)));
```
sharkdp, line 191: please update the doc comment above (or if any of the patterns has ...)

src/main.rs
```diff
@@ -81,12 +81,28 @@ fn run() -> Result<ExitCode> {
...
 
-    let config = construct_config(opts, &pattern_regex)?;
-    ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regex)?;
-    let re = build_regex(pattern_regex, &config)?;
-    walk::scan(&search_paths, Arc::new(re), Arc::new(config))
+    ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regexps)?;
+
+    let regexps = pattern_regexps
+        .iter()
+        .map(|pat| build_regex(pat.to_string(), &config))
```
sharkdp, line 102: isn't `pat` a string already?

src/main.rs
Uthar: I'm getting: `rustc [E0631]: expected signature of fn(&std::string::String) -> _`

src/main.rs
Uthar: There is an error message when I remove the `to_string()`:

```
error[E0308]: mismatched types: expected struct `std::string::String`, found `&std::string::String
```

src/walk.rs
```diff
@@ -49,7 +49,7 @@ pub const DEFAULT_MAX_BUFFER_TIME: Duration = Duration::from_millis(100);
 /// If the `--exec` argument was supplied, this will create a thread pool for executing
```
Uthar, line 49: pattern -> patterns

Uthar:
https://github.com/sharkdp/fd/blob/882e13a77276255e21c65af7eef6382a1fcb45a8/src/walk.rs#L47

pattern -> patterns

Uthar:
So the `--expr` flag instead of `--and` sounds better to you?

Uthar:
https://github.com/sharkdp/fd/blob/d498724edd8be4680ea74ad235f4d2ee8a304911/src/main.rs#L438-L442

The pattern -> A pattern / Some patterns ?

sharkdp:
> So the `--expr` flag instead of `--and` sounds better to you?

No, I like `--and`.

> The pattern -> A pattern / Some patterns ?

Maybe: "The pattern(s) seems to only match… or adjust your search pattern(s)".

src/main.rs
sharkdp: ok, then let's leave this.

src/main.rs
sharkdp: Then I guess you could use `.into_iter()` instead of `.iter()` (and remove `.to_string()`)?

src/cli.rs
```diff
@@ -523,6 +523,17 @@ pub struct Opts {
     )]
     pub pattern: String,
 
+    /// Additional search patterns that need to be matched
+    #[arg(
+        long = "and",
+        value_name = "pattern",
+        long_help = "Add additional required search patterns, all of which must be fulfilled. Multiple \
+                     additional patterns can be specified. The patterns are regular expressions,
+                     unless '--glob'or '--fixed-strings' is used.",
```
sharkdp, line 532:
```suggestion
                     unless '--glob' or '--fixed-strings' is used.",
```

src/main.rs
```diff
@@ -166,11 +181,14 @@ fn check_path_separator_length(path_separator: Option<&str>) -> Result<()> {
     }
 }
 
-fn construct_config(mut opts: Opts, pattern_regex: &str) -> Result<Config> {
+fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config> {
     // The search will be case-sensitive if the command line flag is set or
-    // if the pattern has an uppercase character (smart case).
-    let case_sensitive =
-        !opts.ignore_case && (opts.case_sensitive || pattern_has_uppercase_char(pattern_regex));
+    // if any of the patterns have an uppercase character (smart case).
```
sharkdp, line 186: minor
```suggestion
    // if any of the patterns has an uppercase character (smart case).
```

src/main.rs
Uthar: https://github.com/sharkdp/fd/compare/d8196a109e81e812ac9e558363a210ca2d660e11..1409ad6d306b400c07d75538a804810d8a569bc0

src/main.rs
Uthar: https://github.com/sharkdp/fd/compare/5794d62b45d4db9d02a19b967ec3d3ad3171c36f..d8196a109e81e812ac9e558363a210ca2d660e11

sharkdp (approved):
Thank you. I made a couple of minor improvements.

PR #1144, opened 2022-10-23
chore: Update outdated LICENSE year

bluzir:
Updated the outdated copyright year to the present. 
So there will not be a need for any further license year updates


sharkdp:
Thanks

PR #1147, opened 2022-10-24
walk: Simplify the code in scan() a bit

tavianator:


src/walk.rs
```diff
@@ -49,14 +49,11 @@ pub const DEFAULT_MAX_BUFFER_TIME: Duration = Duration::from_millis(100);
 /// If the `--exec` argument was supplied, this will create a thread pool for executing
 /// jobs in parallel from a given command line and the discovered paths. Otherwise, each
 /// path will simply be written to standard output.
-pub fn scan(path_vec: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> Result<ExitCode> {
-    let mut path_iter = path_vec.iter();
-    let first_path_buf = path_iter
-        .next()
-        .expect("Error: Path vector can not be empty");
+pub fn scan(paths: &[PathBuf], pattern: Arc<Regex>, config: Arc<Config>) -> Result<ExitCode> {
+    let first_path = &paths[0];
```
tmccombs, line 53: Another option, not sure if it is necessarily better could be to do:

```rust
let (first_path, paths) = paths.split_first().expect("Error: Path vector can not be empty");
```
then keep the rest of your changes, except line 121 becomes
```rust
for path in paths {
```

tmccombs (approved):

src/walk.rs
sharkdp: Both versions seem fine to me. If we go with @tmccombs suggestion, maybe rename `paths` to `remaining_paths`?

tavianator: I don't have a strong preference either, but my weak preference is for not having to write the `expect()` message :)

PR #1148, opened 2022-10-24
dir_entry: Implement lscolors::Colorable

tavianator:
Now we can finally avoid calling `stat()` unless actually necessary to color a path.  For example, with `LS_COLORS="ex=0:sg=0:su=0:st=0:ow=0:tw=0:"`

|:---|---:|---:|---:|---:|
| `./fd-master . ~/code/linux -u -j1 --color=always` | 225.8 ± 6.4 | 209.1 | 231.6 | 1.60 ± 0.10 |
| `./fd-colorable . ~/code/linux -u -j1 --color=always` | 140.9 ± 8.3 | 122.6 | 150.0 | 1.00 |

And still a (slight) improvement with my default `LS_COLORS`:

|:---|---:|---:|---:|---:|
| `./fd-master . ~/code/linux -u -j1 --color=always` | 259.2 ± 5.6 | 243.5 | 265.4 | 1.02 ± 0.04 |
| `./fd-colorable . ~/code/linux -u -j1 --color=always` | 255.2 ± 7.5 | 238.0 | 262.4 | 1.00 |

### DIFF
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -1,8 +1,8 @@
-use std::{
-    fs::{FileType, Metadata},
-    path::{Path, PathBuf},
-};
+use std::ffi::OsString;
+use std::fs::{FileType, Metadata};
+use std::path::{Path, PathBuf};
 
+use lscolors::Colorable;
 use once_cell::unsync::OnceCell;
 
 use crate::config::Config;
@@ -96,6 +96,7 @@ impl PartialEq for DirEntry {
         self.path() == other.path()
     }
 }
+
 impl Eq for DirEntry {}
 
 impl PartialOrd for DirEntry {
@@ -109,5 +110,27 @@ impl Ord for DirEntry {
     #[inline]
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         self.path().cmp(other.path())
+    }
+}
+
+impl Colorable for DirEntry {
+    fn path(&self) -> PathBuf {
+        self.path().to_owned()
+    }
+
+    fn file_name(&self) -> OsString {
+        let name = match &self.inner {
+            DirEntryInner::Normal(e) => e.file_name(),
+            DirEntryInner::BrokenSymlink(_) => todo!(),
+        };
+        name.to_owned()
+    }
+
+    fn file_type(&self) -> Option<FileType> {
+        self.file_type()
+    }
+
+    fn metadata(&self) -> Option<Metadata> {
+        self.metadata().cloned()
     }
 }
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -2,7 +2,8 @@ use std::ffi::OsString;
 use std::fs::{FileType, Metadata};
 use std::path::{Path, PathBuf};
 
-use lscolors::Colorable;
+use lscolors::{Colorable, LsColors, Style};
+
 use once_cell::unsync::OnceCell;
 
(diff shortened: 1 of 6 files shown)


sharkdp:
I think this would be worth a ChangeLog entry.

src/dir_entry.rs
```diff
@@ -111,3 +112,25 @@ impl Ord for DirEntry {
...
+
+impl Colorable for DirEntry {
+    fn path(&self) -> PathBuf {
+        self.path().to_owned()
+    }
+
+    fn file_name(&self) -> OsString {
+        let name = match &self.inner {
+            DirEntryInner::Normal(e) => e.file_name(),
+            DirEntryInner::BrokenSymlink(_) => todo!(),
```
tmccombs, line 124: This todo should probably be done before merging. 

I imagine it could probably be something like:


```suggestion
            DirEntryInner::BrokenSymlink(p) => p.file_name().map_or_else(OsString::new, OsStr::to_os_string),
```

well, I guess then you might need to move the `to_owned` up to line 123 as well.

tmccombs (changes requested):

src/dir_entry.rs
sharkdp: oh, I totally missed that

tavianator: Oh whoops.  I wrote this commit a while ago and it was marked WIP.  Couldn't figure out why...

tavianator: This is the right implementation: https://github.com/sharkdp/lscolors/blob/master/src/lib.rs#L433-L443

tavianator: Fixed!  Technically this code is unreachable because lscolors doesn't care about the file name of symlinks.  Still ought to be correct though.

tavianator:
Doing colors in parallel is a gigantic win:

|:---|---:|---:|---:|---:|
| `./fd-master . ~/code/linux --color=always` | 272.8 ± 14.6 | 244.2 | 287.5 | 2.31 ± 0.15 |
| `./fd-colorable . ~/code/linux --color=always` | 118.0 ± 4.3 | 108.8 | 125.5 | 1.00 |

No significant difference without colors:

|:---|---:|---:|---:|---:|
| `./fd-master . ~/code/linux` | 144.5 ± 4.4 | 136.7 | 151.5 | 1.00 |
| `./fd-colorable . ~/code/linux` | 145.3 ± 3.7 | 138.2 | 150.9 | 1.01 ± 0.04 |

sharkdp:
> Doing colors in parallel is a gigantic win:

Oh wow, very nice!

src/walk.rs
```diff
@@ -535,6 +535,13 @@ fn spawn_senders(
                 }
             }
 
+            if config.is_printing() {
+                if let Some(ls_colors) = &config.ls_colors {
+                    // Try to compute colors in parallel
```
sharkdp, line 540: I'm not sure I understand this comment ("Try to…"). The code looks to me like we *do* compute the style in parallel (since it happens here, not in the receiver). Or is there a scenario where we still need to compute it in the receiver?

tavianator: I said "try to" because I thought if it returned `None`, it would retry from `print_entry_colored()`.  However that's not how `OnceCell` works, it will cache the `None`.  I'll change the comment.

sharkdp (approved):

PR #1159, opened 2022-11-02
Silence warning about large enum variant

tmccombs:
Because boxing the large variant is probably worse.


sharkdp (approved):
Thank you. Needs a small `cargo fmt` update.

Issue #1163, opened 2022-11-02
Panic when using owner


mbhall88:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

```
$ fd -d 2 -o $USER -t d . /hps/scratch/lsf_tmpdir
thread 'main' panicked at 'Mismatch between definition and access of `owner`. Could not downcast to TypeId { t: 14109425959541630025 }, need to downcast to TypeId { t: 563790022623738129 }
', /nfs/research/zi/mbhall//Software/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-4.0.18/src/parser/error.rs:30:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I guess this is akin to #1160? But as it is not a `u32` I wasn't sure how to put in a fix (I was planning to copy the fix for that issue)

### Describe what you expected to happen:

Before upgrading to this version, the command ran fine. I expect it to list all directories owned by me (`$USER`)

### What version of `fd` are you using?

0.5.1

### Which operating system / distribution are you on?

```shell
Linux 4.18.0-240.22.1.el8_3.x86_64 x86_64


LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 8.3.2011
Release:        8.3.2011
Codename:       n/a
```

tmccombs:
hmm, I wonder if these are caused by a bug in clap

tmccombs:
I'll look into this later tonight.

sharkdp:
Fix released: https://github.com/sharkdp/fd/releases/tag/v8.5.2

tmccombs:
Thanks @sharkdp

PR #1164, opened 2022-11-03
Fix panic when using --owner

tmccombs:
Unfortunately, clap_derive can't combine a value_parser of Option<T> with an optional argument to get a merged Option<T> so we need to do the check for the nop outside of the value parser.

Also adds some tests for --owner

Fixes: #1163


sharkdp (approved):
Thank you very much!

Issue #1166, opened 2022-11-07
[BUG] building `fd` with `--no-default-features` fails


hadfl:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

```
error: cannot find macro `anyhow` in this scope
   --> src/cli.rs:845:13
    |
845 |         Err(anyhow!(
    |             ^^^^^^
    |
    = note: consider importing one of these items:
            anyhow::anyhow
            crate::anyhow
    = note: `anyhow` is in scope, but it is a crate, not a macro
```

### Describe what you expected to happen:

the `fd` build to succeed.

### What version of `fd` are you using?

8.5.2

### Which operating system / distribution are you on?

```shell
SunOS nemesis 5.11 omnios-master-74f47d7f9b2 i86pc i386 i86pc
```

PR #1167, opened 2022-11-07
anyhow is not only used for the completions feature

hadfl:
fixes: https://github.com/sharkdp/fd/issues/1166

currently building `fd` with `--no-default-features` fails:

```
error: cannot find macro `anyhow` in this scope
   --> src/cli.rs:845:13
    |
845 |         Err(anyhow!(
    |             ^^^^^^
    |
    = note: consider importing one of these items:
            anyhow::anyhow
            crate::anyhow
    = note: `anyhow` is in scope, but it is a crate, not a macro
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,7 +1,6 @@
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
-#[cfg(feature = "completions")]
 use anyhow::anyhow;
 use clap::{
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

sharkdp:
I'm wondering if we should add CI tests for thing like this. I do this for other projects as well. If we only run `cargo check`, it's also reasonably quick. We could test various (sensible) combination of feature flags.

tmccombs:
sounds good to me.

PR #1169, opened 2022-11-11
Using faccess lib to detect executable files

### LINKED ISSUE #1106: help: mention that fd detects executables by file permissions
Detecting an executable efficiently is not simple: https://stackoverflow.com/questions/4458120/search-for-executable-files-using-find-command

fd uses the executable bit of the file permissions, which should be mentioned in the help page.

Ptipiak:
* The detection of executable files was not exactly the same as the original find
* resolve #1106 and close #1106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@
 - Fix completion generation to not include full path of fd command
 - Fix build error if completions feature is disabled
 
+## Changes
+
+- On Unix-like system change the detection of executable,
+  `--type executable` now checks if file is executable by the current user,
+  see #1106 (@ptipiak)
+
 # v8.5.2
 
 ## Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,27 @@
 # Upcoming release
 
+## Features
+
+
+## Changes
+
+- Breaking: On Unix-like systems, `--type executable` now additionally checks if
+  the file is executable by the current user, see #1106 and #1169 (@ptipiak)
+
+
+## Bugfixes
+
+
+## Other
+
+
+
 # v8.5.3
 
 ## Bugfixes
 
 - Fix completion generation to not include full path of fd command
 - Fix build error if completions feature is disabled
-
-## Changes
-
-- On Unix-like system change the detection of executable,
-  `--type executable` now checks if file is executable by the current user,
-  see #1106 (@ptipiak)
 
 # v8.5.2
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ chrono = "0.4"
 once_cell = "1.15.0"
 crossbeam-channel = "0.5.6"
 clap_complete = {version = "4.0.5", optional = true}
+faccess = "0.2.4"
 
 [dependencies.clap]
 version = "4.0.22"
@@ -62,9 +63,6 @@ nix = { version = "0.24.2", default-features = false, features = ["signal"] }
 
 [target.'cfg(all(unix, not(target_os = "redox")))'.dependencies]
 libc = "0.2"
-
-[target.'cfg(windows)'.dependencies]
-faccess = "0.2.4"
 
 # FIXME: Re-enable jemalloc on macOS
 # jemalloc is currently disabled on macOS due to a bug in jemalloc in combination with macOS
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -643,6 +643,7 @@ pub enum FileType {
     Directory,
     #[value(alias = "l")]
     Symlink,
+    /// A file which is executable by the current effective user
     #[value(alias = "x")]
     Executable,
     #[value(alias = "e")]
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -4,11 +4,8 @@ use std::ffi::OsStr;
 use std::fs;
 use std::io;
 #[cfg(any(unix, target_os = "redox"))]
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# unreleased
+
+## Bugfixes
+
+- Using faccess library to detect executable files, see #1106 (@ptipiak)
```
tmccombs, line 5: This is interesting. I'm not entirely sure if I would classify this as a bug. The previous behavior was technically still in alignment with the documentation. The problem is that the definition of what it means for a file to be "executable" was not well defined. 

Speaking of which, it would probably still be good to better document what is meant by "executable" here. But now clarify that executable means "executable by the current effective user".

tmccombs (approved):

CHANGELOG.md
Ptipiak: I wasn't exatcly sure either how to classify it, since the goal is to implement every feature of the original find, I assume correcting deviation would be considered bugfix.

tavianator: I think "Using faccess library" is just an implementation detail that isn't necessary for the changelog.  It would be better to focus on how the change affects users, something "`--type executable` on Unix-like systems now checks whether a file is executable by the current user"

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# unreleased
+
+## Changes
+
+- On Unix-like system change the detection of executable,
+  `--type executable` now check if a file is executable and if it's executable by the current user,
```
tmccombs, line 6:
```suggestion
  `--type executable` now checks if a file is executable by the current user,
```

tmccombs (approved):

Cargo.toml
```diff
@@ -50,7 +50,8 @@ normpath = "0.3.2"
 chrono = "0.4"
 once_cell = "1.15.0"
 crossbeam-channel = "0.5.6"
-clap_complete = {version = "4.0.5", optional = true}
+clap_complete = {version = "4.0", optional = true}
```
sharkdp, line 53: This looks like a wrongly resolved merge conflict?

CHANGELOG.md
sharkdp: > since the goal is to implement every feature of the original find, I assume correcting deviation would be considered bugfix.

Just to clarify: It is *not* `fd`s goal to implement every feature of find. But in terms of behavior, we certainly only want to deviate from `find` if there are really good reasons to do so.

For this particular feature (`--type executable`), it was really more of an oversight on my part. So both bugfix or change seem fine to me. Let's keep it under changes now.

src/cli.rs
```diff
@@ -643,6 +643,7 @@ pub enum FileType {
     Directory,
     #[value(alias = "l")]
     Symlink,
+    /// Filter file which are executable by the current effective user
```
sharkdp, line 646: This is not the filter itself. Maybe just
```suggestion
    /// A files which is executable by the current effective user
```

Also, what is the current *effective* user? Is that a known term? Or could it simply be
```suggestion
    /// A file which is executable by the current user
```

src/filesystem.rs
```diff
@@ -44,13 +43,8 @@ pub fn is_existing_directory(path: &Path) -> bool {
...
 
-#[cfg(any(unix, target_os = "redox"))]
-pub fn is_executable(_: &Path, md: &fs::Metadata) -> bool {
-    md.permissions().mode() & 0o111 != 0
-}
-
-#[cfg(windows)]
-pub fn is_executable(path: &Path, _: &fs::Metadata) -> bool {
+/// Return true if the path point to a file executable by the current effective user
+pub fn is_executable(path: &Path) -> bool {
```
sharkdp, line 47: This whole helper function is now completely unnecessary. I think we should simply use `.executable()` at the usage site?

src/cli.rs
Ptipiak: I agree, the filter part is not relevant, as for the term "effective user" it came from the doc at [faccess](https://docs.rs/faccess/0.2.4/faccess/trait.PathExt.html#tymethod.access)
From a quick search it appear that:
* effective user: is the user which call the executable, which can be root in case of non-privileged user using `sudo`
* RealId user: user which is the original user who have called the executable

Cargo.toml
Ptipiak: Indeed, my bad I didn't notice the minor version number

src/filesystem.rs
Ptipiak: Good question, I'd be in favor of less lines is better, but it can also be more elegant and coherent to keep the check itself in the filesystem part.

src/cli.rs
tavianator: The effective and real users are only different for setuid executables themselves.  Presumably no one makes `fd` a setuid executable, so they will be the same.

tmccombs: I think the effective and real user could also be different if a program running as root set the effective user to something different, then execed fd with the right flags or environment. But that also seems contrived. I'm fine with just saying current user.

src/filesystem.rs
sharkdp: The `filesystem` module  is only here to add *additional* functionality that is not provided by the standard library (+ `faccess`, now). It was never intended to be a wrapper around all possible filesystem functionality. So I'd be in favor of removing `is_executable`.

src/cli.rs
sharkdp: Thank you. In that case, I'm fine with either option.

PR #1171, opened 2022-11-12
Use docstrings for long help in clap derive

tmccombs:
I'm not entirely sure this is better than passing strings to long_help, but I did figure out a way to use doc comments for the long_help with clap_derive without changing the help text. it's a little weird to get the formatting right in some cases though.


sharkdp:
Cool. Thank you for looking into this. What is the "a little weird" part? The `{n}` annotations (what are those exactly)? Do we expect contributors to have problems with this in the future?

tmccombs:
Yes the {n} annotations. They mean insert a literal newline. And the have to be at the beginning of the line so that the leading whitespace doesn't get stripped.

sharkdp:
It's actually better than the previous `\n    \` dance to get indented text, right? I think I'm in favor of this change.

tmccombs:
Yeah, I  think I prefer it as well, the biggest downside to the `{n}` is probably that won't look good in the generated rustdocs... but that isn't really even exposed anyway, so I'm not worried about that for fd.

sharkdp (approved):
Yes, good point. But I agree with you. This looks good to merge for me, but I would really appreciate if we could make sure there are no (or only irrelevant) diffs with `fd --help`/`fd -h` on `master`.

tmccombs:
There is no change to the short help output. The long help output has the following changes:

```
❯ diff -uZ long-help.old long-help.new                                                                                                                                                                   
--- long-help.old	2022-12-06 01:27:48.158572102 -0700
+++ long-help.new	2022-12-06 01:28:41.865242942 -0700
@@ -144,6 +144,7 @@
   -e, --extension <ext>
           (Additionally) filter search results by their file extension. Multiple allowable file
           extensions can be specified.
+          
           If you want to search for files without extension, you can use the regex '^[^.]+$' as a
           normal search pattern.
 
@@ -151,6 +152,7 @@
           Limit results based on the size of files using the format <+-><NUM><UNIT>.
              '+': file size must be greater than or equal to this
              '-': file size must be less than or equal to this
+          
           If neither '+' nor '-' is specified, file size must be exactly equal to this.
              'NUM':  The numeric size (e.g. 500)
              'UNIT': The units for NUM. They are not case-sensitive.
@@ -169,18 +171,21 @@
           Filter results based on the file modification time. Files with modification times
           greater than the argument are returned. The argument can be provided as a specific point
           in time (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). If the time is not
-          specified, it defaults to 00:00:00. '--change-newer-than' or '--newer' can be used as
-          aliases.
+          specified, it defaults to 00:00:00. '--change-newer-than', '--newer', or
+          '--changed-after' can be used as aliases.
+          
           Examples:
               --changed-within 2weeks
               --change-newer-than '2018-10-27 10:00:00'
               --newer 2018-10-27
+              --changed-after 1day
 
       --changed-before <date|dur>
           Filter results based on the file modification time. Files with modification times less
           than the argument are returned. The argument can be provided as a specific point in time
           (YYYY-MM-DD HH:MM:SS) or as a duration (10h, 1d, 35min). '--change-older-than' or
           '--older' can be used as aliases.
+          
           Examples:
               --changed-before '2018-10-27 10:00:00'
               --change-older-than 2weeks
@@ -189,6 +194,7 @@
   -o, --owner <user:group>
           Filter files by their user and/or group. Format: [(user|uid)][:(group|gid)]. Either side
           is optional. Precede either side with a '!' to exclude files instead.
+          
           Examples:
               --owner john
               --owner :students
               ```
               
  I did add a reference to the new --changed-after alias. The other changes are additional newlines. I think those are probably an improvement, but would be fine removing them if you want.

sharkdp:
Thank you!

PR #1173, opened 2022-11-13
Use just the file name of the program name for generating completions

### LINKED ISSUE #1172: [BUG] Generated completions aren't picked up
### Checks

Automatic generation of completions use path to executable and thus won't work. On current master (85e3adaf1800fd8cb192f7a4a2bb9000f993f5bc) after issuing `make completions` the created completion for e.g. `fish` looks like (just the first 3 lines are shown as the pattern should be clear):
```
complete -c target/release/fd -s d -l max-depth -d 'Set maximum search depth (default: none)' -r
complete -c target/release/fd -l min-depth -d 'Only show search results starting at the given depth' -r
complete -c target/release/fd -l exact-depth -d 'Only show search results at the exact given depth' -r
...
```
as the (later) installed binary doesn't match the command.

The generated completions should not use path names.

This is caused by `src/main.rs:96`, which takes the full first argument without removing paths.

fd 8.5.2

```shell
$ uname -srm
Linux 6.0.8-arch1-1 x86_64
```

```
$ lsb_release -a
LSB Version:    n/a
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a
```

zappolowski:
If passed a full path to the executable (like done in the Makefile) that value is used in the generated completions. At least for fish this just doesn't work.

This fixes #1172

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -93,10 +93,16 @@ fn run() -> Result<ExitCode> {
 #[cold]
 fn print_completions(shell: clap_complete::Shell) -> Result<ExitCode> {
     // The program name is the first argument.
-    let program_name = env::args().next().unwrap_or_else(|| "fd".to_string());
+    let first_arg = env::args().next();
+    let program_name = first_arg
+        .as_ref()
+        .map(Path::new)
+        .and_then(|path| path.file_name())
+        .and_then(|file| file.to_str())
+        .unwrap_or("fd");
     let mut cmd = Opts::command();
     cmd.build();
-    clap_complete::generate(shell, &mut cmd, &program_name, &mut std::io::stdout());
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -93,10 +93,16 @@ fn run() -> Result<ExitCode> {
 #[cold]
 fn print_completions(shell: clap_complete::Shell) -> Result<ExitCode> {
     // The program name is the first argument.
-    let program_name = env::args().next().unwrap_or_else(|| "fd".to_string());
+    let first_arg = env::args().next();
+    let program_name = first_arg
+        .as_ref()
+        .map(|cmd| Path::new(cmd))
```
tmccombs, line 99:
```suggestion
        .map(Path::new)
```

tmccombs (approved):

PR #1183, opened 2022-11-24
Adding a simple test for executable files

### LINKED ISSUE #1106: help: mention that fd detects executables by file permissions

Ptipiak:
resolve #1106 close #1106

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1197,6 +1197,13 @@ fn test_type_executable() {
         .open(te.test_root().join("executable-file.sh"))
         .unwrap();
 
+    fs::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .mode(0o645)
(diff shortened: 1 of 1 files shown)


tests/tests.rs
```diff
@@ -1197,6 +1197,13 @@ fn test_type_executable() {
         .open(te.test_root().join("executable-file.sh"))
         .unwrap();
 
+    fs::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .mode(0o005)
```
tmccombs, line 1203: 0o645 ?

Ptipiak: `0o645` would work too, originaly I've tested on 005

tmccombs (approved):

PR #1190, opened 2022-11-30
Replace chrono with just time.

tmccombs:
And ugprade to time 0.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,15 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


azzamsa:
@tmccombs. Not being able to get the local offset in parallel is why you switch back to Chrono?

I

tmccombs:
The master branch never switched away from chrono.  But yes, the problem with getting the offset in parallel was one reason we stuck with chrono. Specifically because it rust runs testa in parallel by default, so we couldn't get the offset in unit tests without disabling parallel test runs

azzamsa:
Thanks for taking the time to answer. 

I port some of my crates to `time` months ago. I can live with `--test-threads 1` (https://github.com/nextest-rs/nextest/issues/715). But lately, there have been cases where my user wants to use Tokio `spawn` but it fails because of the local offset issue. So, I plan to port it back to Chrono sometime this week.

Issue #1188, opened 2022-11-29
cargo audit finding on default branch


baffling-knee:
didnt see sec contact steps in readme so opening issue instead

looks like a couple unmaintained crates, semi ok, but `time` has issue. `chrono` may not have right update and ignored issue for awhile. other project switch to `time` specific like https://github.com/Drakulix/simplelog.rs/pull/95

```
Crate:     time
Version:   0.1.44
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.44
└── chrono 0.4.22
    └── fd-find 8.5.3

Crate:     ansi_term
Version:   0.12.1
Warning:   unmaintained
Title:     ansi_term is Unmaintained
Date:      2021-08-18
ID:        RUSTSEC-2021-0139
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0139
Dependency tree:
ansi_term 0.12.1
├── lscolors 0.12.0
│   └── fd-find 8.5.3
└── fd-find 8.5.3

Crate:     tempdir
Version:   0.3.7
Warning:   unmaintained
Title:     `tempdir` crate has been deprecated; use `tempfile` instead
Date:      2018-02-13
ID:        RUSTSEC-2018-0017
URL:       https://rustsec.org/advisories/RUSTSEC-2018-0017
Dependency tree:
tempdir 0.3.7
└── fd-find 8.5.3

error: 1 vulnerability found!
warning: 2 allowed warnings found
```

Ptipiak:
The library tempdir is not maintained and the github repository as been archive: https://github.com/rust-lang-deprecated/tempdir
The library should be replaced with a new maintained one with support to non-clashing directory names ?
Or the random tempory directory generation could be taken internaly by `fd` to make sure it fullfills it's need ?

tavianator:
There is nu_ansi_term that lscolors also supports

PR #1191, opened 2022-11-30
Remove time from transitive dependencies of chrono

tmccombs:
Because chrono was depending on an older version of time, and we don't actually need it with recent versions of chrono.

Relates-To: #1188
Alernative-To: #1190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,6 @@ globset = "0.4"
 anyhow = "1.0"
 dirs-next = "2.0"
 normpath = "0.3.2"
-chrono = "0.4"
 once_cell = "1.15.0"
 crossbeam-channel = "0.5.6"
 clap_complete = {version = "4.0.6", optional = true}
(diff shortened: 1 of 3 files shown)


tavianator:
I think I prefer this to #1190

Cargo.toml
```diff
@@ -57,6 +56,12 @@ faccess = "0.2.4"
 version = "4.0.22"
 features = ["suggestions", "color", "wrap_help", "cargo", "unstable-grouped", "derive"]
 
+[dependencies.chrono]
+version = "0.4.23"
+default-features = false
+features = ["std", "clock"]
+
+
```
tavianator, line 64: Nit: extra newline

src/filter/time.rs
```diff
@@ -20,7 +20,7 @@ impl TimeFilter {
                     .ok()
                     .or_else(|| {
                         NaiveDate::parse_from_str(s, "%F")
-                            .map(|nd| nd.and_hms(0, 0, 0))
+                            .map(|nd| nd.and_hms_opt(0, 0, 0).unwrap())
                             .ok()
```
tavianator, line 24:
```suggestion
                            .and_then(|nd| nd.and_hms_opt(0, 0, 0).ok())
```

tmccombs: It shouldn't ever fail with those values, but sure.

tavianator (approved):
LGTM, just needs the conflict fixed

src/filter/time.rs
tavianator: Oh true, I was looking at `LocalResult` but for `NaiveDate` that should always be valid.

tmccombs:
Conflicts resolved.

PR #1194, opened 2022-12-01
Fixing clashing naming of temp directories

### LINKED ISSUE #1181: Fix flaky `--full-path` unit tests

Some of our `--full-path` unit tests are non-deterministic and can fail.

The reason is that `--full-path` searches the full *absolute* path. This can lead to additional `fd` matches if (by chance) the *temporary directory name contains one of the search patterns*. For example, in [this CI run](https://github.com/sharkdp/fd/actions/runs/3517827315/jobs/5896045947), the `test_and_plus_full_path` unit test failed with an additional unexpected match (`one/two/three/d.foo`), even though it only contains two of the three search patterns of the `fd three --full-path --and foo --and dir` call.

I was able to reproduce this locally after running the test thousands of times (the odds are 1 : 3000). The test fails if the temporary directory name contains a *dir* substring (case insensitive):

image

All unit tests should be deterministic. When we fix this, we should go through all `--full-path` tests and find a good strategy to avoid this kind of "vulnerability".

c9d39684758d0d898082370d5c5fe7e8aef3a385

Linux. I'm sure this can similarly fail on other operating systems as well.

@Uthar FYI

Ptipiak:
## Objective
* The random names of temporay directories used for test are clashing witht the actual test
* Fixes #1181 

## Solution
* using 2 bytes for naming the directories

### DIFF
--- a/tests/testenv/mod.rs
+++ b/tests/testenv/mod.rs
@@ -27,7 +27,7 @@ fn create_working_directory(
     directories: &[&'static str],
     files: &[&'static str],
 ) -> Result<TempDir, io::Error> {
-    let temp_dir = tempfile::Builder::new().prefix("fd-tests").tempdir()?;
+    let temp_dir = tempfile::Builder::new().prefix("fd-tests").rand_bytes(2).tempdir()?;
 
     {
         let root = temp_dir.path();


tmccombs (changes requested):
cargo fmt needs to be run, but other than that, this looks ok to me.

tavianator:
I think two bytes is not enough.  I can reliably make individual test cases take more than 1 minute just by running a couple copies of the test suite at once, as my `/tmp` fills up with `fd-tests??` directories.

This also doesn't fix the problem that the tests fail 100% of the time when run from e.g. `/home/dirk/fd`.

Instead I think we need to change our test cases to use names that can't be generated by `tempfile` (e.g. by using symbols), and that also aren't likely to show up in normal paths.

Ptipiak:
I think it should be the temp file generation which should provide a way to avoid certain sub-string in the genreated names.
Would it be possible to use a pre-defined set of directory names as random name for `tempfile`? This way we could make sure non of the randomly choosen names would clash with the tests

tavianator:
I believe #1200 supersedes this.  Thank you for the contribution!  It's always great to be able to try multiple approaches.

PR #1200, opened 2022-12-02
Change full path tests to be more resilient

tmccombs:
Use patterns that won't conflict with randomly generated portion of temporary directory

Fixes: #1181

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -263,12 +263,26 @@ fn test_and_plus_full_path() {
     let te = TestEnv::new(DEFAULT_DIRS, AND_EXTRA_FILES);
 
     te.assert_output(
-        &["three", "--full-path", "--and", "_foo", "--and", "/dir"],
+        &[
+            "three",
+            "--full-path",
+            "--and",
+            "_foo",
+            "--and",
+            r"[/\\]dir",
+        ],
         "one/two/three/directory_foo/",
     );
 
     te.assert_output(
-        &["three", "--full-path", "--and", "/two", "--and", "/dir"],
+        &[
+            "three",
+            "--full-path",
+            "--and",
+            r"[/\\]two",
+            "--and",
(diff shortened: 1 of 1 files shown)


tavianator:
Maybe the failing Windows tests are due to `/` vs `\`?

tmccombs:
Yeah, they are. I didn't think about that :(

sharkdp:
Thank you! Does this completely solve #1181 though? Don't we have other full-path tests that could suffer from this issue?

tmccombs:
I looked through the other full-path tests and didn't see any others that would run into this problem. 

Specifically:

- test_full_path_glob_searches: the pattern includes "/one/", so it will only match if it is in a directory named exactly "one", which the tempdir should not be.
- test_full_path would have to match "three.*foo", and since "threefoo" is 7 characters long, it is impossible for the current implementation to generate a temporary directory that includes that in the random sequence (it only uses 6 ascii characters right now). Although I suppose that is an implementation detail and might change in the future if we don't explicitly specify how many random chars to use. We could probably do something similar here. 

However, I did realize, that it might be possible for the base temporary directory (as returned by `std::env::temp_dir()`) to be a subdirectory of the user's home directory (or really any directory, depending on environment variables).  That would be more difficult to deal with, because we have no control over that (when users run tests themselves), the best we can do is make the patterns unlikely to match.  That shouldn't impact CI though.

sharkdp:
Thank you for the explanation!

> However, I did realize, that it might be possible for the base temporary directory (as returned by `std::env::temp_dir()`) to be a subdirectory of the user's home directory (or really any directory, depending on environment variables). That would be more difficult to deal with, because we have no control over that (when users run tests themselves), the best we can do is make the patterns unlikely to match. That shouldn't impact CI though.

Ok. Let's postpone this for now :smile:

PR #1201, opened 2022-12-02
Completion gen faster

tmccombs:
This is an attempt to speed up CI builds, by pulling completion generation out into a separate job. I don't know if it will work though.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -54,9 +54,39 @@ jobs:
     - name: Run tests
       run: cargo test --locked
 
+  completions:
+    name: Generate shell completions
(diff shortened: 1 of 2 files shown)


tmccombs:
This might actually make it slower though, by having additional jobs. 

Some other things we might be able to do:

1.  Use my first commit on this branch so that we can re-use the output from the build step to generate the completions. That was my original plan, but then I realized it wont' work if for the jobs where we cross-compile to another architecture.
2. Set up caching so that if one job generates completions it can be re-used by other jobs. Would probably want to do 1 in addition to this.
3. Do this PR, but also use more caching, so that building on the completion jobs does work that benefits some of the build/test jobs.

PR #1212, opened 2022-12-18
Add tests for global ignore file

tmccombs:


tmccombs:
hmm, looks like `XDG_CONFIG_HOME` doesn't work on windows. I'm not sure if there is a good way to accomplish the same thing on windows, or if I should just disable these tests on windows.

vegerot (approved):
I'd like to use these to give me more confidence in #1218

PR #1216, opened 2022-12-29
Add flag --no-require-git to always respect gitignore files

vegerot:
Add flag --no-require-git to always respect gitignore files

Summary: Currently, `--ignore-vcs` only serves to unset `--no-ignore-vcs`.
There is currently no way to tell fd to respect gitignore files when not in a
git repository.  This commit adds the flag `--no-require-git` to make fd always
respect all gitignore files.

This behaves the same as the `--no-require-git` option in [ripgrep](https://github.com/BurntSushi/ripgrep/blob/3bb71b0cb8727ac43237af78ba5c707298de0128/crates/core/app.rs#L2214-L2226)

This commit also contains an unrelated wording fix to CONTRIBUTING.md

Test Plan: `tests/tests.rs`

Background: I am using [Sapling](https://sapling-scm.com/docs/introduction/)
for working with git repositories (including this commit ☺️).  Since Sapling
uses `.sl` instead of `.git`, tools using the `ignore` crate (rg and fd) would show gitignored files.
I made a patch (https://github.com/vegerot/ripgrep/commit/ebf17eef22a04450ca86e9dc790db7cb00e11e52)
to `ignore` to respect gitignores with _either_ `.git` or `.sl`.  However,
@BurntSushi said he did not want to merge that patch and instead suggested I
use `--no-require-git` (https://github.com/BurntSushi/ripgrep/issues/2374).
This works fine, but I couldn't use this workaround for my other favorite tool!
That's what this patch is for 😁

(a follow-up patch will add a similar `FD_CONFIG_PATH` environment variable
like `RG_CONFIG_PATH`)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/sharkdp/fd/pull/1216).
* #1218
* __->__ #1216

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -74,6 +74,28 @@ pub struct Opts {
     #[arg(long, overrides_with = "no_ignore_vcs", hide = true, action = ArgAction::SetTrue)]
     ignore_vcs: (),
 
+    /// Do not require a git repository to respect gitignores.
+    /// By default, fd will only respect global gitignore rules, .gitignore rules,
+    /// and local exclude rules if fd detects that you are searching inside a
+    /// git repository. This flag allows you to relax this restriction such that
+    /// fd will respect all git related ignore rules regardless of whether you're
+    /// searching in a git repository or not.
+    ///
+    ///
+    /// This flag can be disabled with --require-git.
+    #[arg(
+        long,
+        overrides_with = "require_git",
+        hide_short_help = true,
+        // same description as ripgrep's flag: ripgrep/crates/core/app.rs
+        long_help
+    )]
+    pub no_require_git: bool,
+
+    /// Overrides --no-require-git
+    #[arg(long, overrides_with = "no_require_git", hide = true, action = ArgAction::SetTrue)]
+    require_git: (),
+
     /// Show search results from files and directories that would otherwise be
     /// ignored by '.gitignore', '.ignore', or '.fdignore' files in parent directories.
     #[arg(
--- a/src/config.rs
+++ b/src/config.rs
@@ -30,6 +30,9 @@ pub struct Config {
     /// Whether to respect VCS ignore files (`.gitignore`, ..) or not.
     pub read_vcsignore: bool,
 
+    /// Whether to require a `.git` directory to respect gitignore files.
+    pub require_git_to_read_vcsignore: bool,
+
     /// Whether to respect the global ignore file or not.
     pub read_global_ignore: bool,
 
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -808,6 +808,62 @@ fn test_custom_ignore_precedence() {
     te.assert_output(&["--no-ignore", "foo"], "inner/foo");
 }
 
+/// Don't require git to respect gitignore (--no-require-git)
+#[test]
+fn test_respect_ignore_files() {
+    let te = TestEnv::new(DEFAULT_DIRS, DEFAULT_FILES);
+
+    // Not in a git repo anymore
+    fs::remove_dir(te.test_root().join(".git")).unwrap();
+
+    // don't respect gitignore because we're not in a git repo
+    te.assert_output(
+        &["foo"],
+        "a.foo
+        gitignored.foo
+        one/b.foo
+        one/two/c.foo
+        one/two/C.Foo2
+        one/two/three/d.foo
+        one/two/three/directory_foo/",
+    );
+
+    // respect gitignore because we set `--no-require-git`
+    te.assert_output(
+        &["--no-require-git", "foo"],
+        "a.foo
+        one/b.foo
+        one/two/c.foo
+        one/two/C.Foo2
+        one/two/three/d.foo
+        one/two/three/directory_foo/",
+    );
+
+    // make sure overriding works
+    te.assert_output(
+        &["--no-require-git", "--require-git", "foo"],
+        "a.foo
+        gitignored.foo
+        one/b.foo
+        one/two/c.foo
+        one/two/C.Foo2
+        one/two/three/d.foo
+        one/two/three/directory_foo/",
+    );
+
+    te.assert_output(
+        &["--no-require-git", "--no-ignore", "foo"],
+        "a.foo
+        gitignored.foo
+        fdignored.foo
+        one/b.foo
+        one/two/c.foo
+        one/two/C.Foo2
+        one/two/three/d.foo
+        one/two/three/directory_foo/",
+    );
+}
+
 /// VCS ignored files (--no-ignore-vcs)
 #[test]
 fn test_no_ignore_vcs() {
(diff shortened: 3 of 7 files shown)


src/config.rs
```diff
@@ -27,9 +27,12 @@ pub struct Config {
     /// Whether to respect ignore files in parent directories or not.
     pub read_parent_ignore: bool,
 
-    /// Whether to respect VCS ignore files (`.gitignore`, ..) or not.
+    /// Whether to respect VCS ignore files (`.gitignore`, ..) if in a git repo (or not).
     pub read_vcsignore: bool,
 
+    /// Whether to always respect VCS ignore files (`.gitignore`, ..) whether or not in a git repo.
+    pub force_read_vcsignore: bool,
```
vegerot, line 34: minor: maybe
```suggestion
    pub always_read_vcsignore: bool,
```

tmccombs:
I think I'd rather have a separate option for this. In particular, because changing the semantics of --ignore-vcs would mean there is no longer a way to exactly negate --no-ignore-vcs

vegerot:
> I think I'd rather have a separate option for this. In particular, because changing the semantics of --ignore-vcs would mean there is no longer a way to exactly negate --no-ignore-vcs

Good point.  @tmccombs what do you think about using the same name as ripgrep's option: `--{no-,}require-git`?

tmccombs:
That seems ok to me

vegerot:
@tmccombs done ✅!

vegerot:
I should add `--{no-,}require-git` to the `test_opposing` test

tmccombs (approved):

src/cli.rs
```diff
@@ -74,6 +74,27 @@ pub struct Opts {
     #[arg(long, overrides_with = "no_ignore_vcs", hide = true, action = ArgAction::SetTrue)]
     ignore_vcs: (),
 
+    /// Do not require a git repository to respect gitignores.
+    #[arg(
+        long,
+        overrides_with = "require_git",
+        hide_short_help = true,
+        help = "Do not require a git repository to use gitignores.",
```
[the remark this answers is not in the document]
vegerot: thank you

src/cli.rs
```diff
@@ -74,6 +74,26 @@ pub struct Opts {
...
+        overrides_with = "require_git",
+        hide_short_help = true,
+        // same description as ripgrep's flag
+        long_help = "By default, fd will only respect global gitignore rules, .gitignore rules,\
+                     and local exclude rules if fd detects that you are searching inside a\
+                     git repository. This flag allows you to relax this restriction such that\
+                     fd will respect all git related ignore rules regardless of whether you're\
+                     searching in a git repository or not.\n\n\
+                     This flag behaves the same as --no-require-git flag from ripgrep.\n\n\
+                     This flag can be disabled with --require-git."
```
vegerot, line 89: FYI this is the same description as `--no-require-git` for ripgrep https://github.com/BurntSushi/ripgrep/blob/3bb71b0cb8727ac43237af78ba5c707298de0128/crates/core/app.rs#L2216-L2225

In your opinion does this port require a more formal attribution?

BurntSushi: fd isn't my project, but I personally wouldn't describe behavior by referencing other tools like this.

> In your opinion does this port require a more formal attribution?

On my end, no. It's probably good sense to mention it in a commit message or perhaps a source code comment so that others looking at it in the future have proper context, but I don't think there's any need for anything more.

vegerot: I did mention it in a source-code comment

vegerot: Thanks!  I did already mention this in a source code comment 👍

src/cli.rs
```diff
@@ -74,6 +74,26 @@ pub struct Opts {
...
+        long,
+        overrides_with = "require_git",
+        hide_short_help = true,
+        // same description as ripgrep's flag
+        long_help = "By default, fd will only respect global gitignore rules, .gitignore rules,\
+                     and local exclude rules if fd detects that you are searching inside a\
+                     git repository. This flag allows you to relax this restriction such that\
+                     fd will respect all git related ignore rules regardless of whether you're\
+                     searching in a git repository or not.\n\n\
+                     This flag behaves the same as --no-require-git flag from ripgrep.\n\n\
```
vegerot, line 88: As per @BurntSushi 's recommendation
```suggestion
```

tests/tests.rs
```diff
@@ -808,6 +808,62 @@ fn test_custom_ignore_precedence() {
...
+        "a.foo
+        gitignored.foo
+        one/b.foo
+        one/two/c.foo
+        one/two/C.Foo2
+        one/two/three/d.foo
+        one/two/three/directory_foo/",
+    );
+
+    // respect gitignore because we set `--ignore-vcs`
```
sharkdp, line 831:
```suggestion
    // respect gitignore because we set `--no-require-git`
```

src/cli.rs
```diff
@@ -74,6 +74,25 @@ pub struct Opts {
...
+        overrides_with = "require_git",
+        hide_short_help = true,
+        // same description as ripgrep's flag: ripgrep/crates/core/app.rs
+        long_help = "By default, fd will only respect global gitignore rules, .gitignore rules,\
+                     and local exclude rules if fd detects that you are searching inside a\
+                     git repository. This flag allows you to relax this restriction such that\
+                     fd will respect all git related ignore rules regardless of whether you're\
+                     searching in a git repository or not.\n\n\
+                     This flag can be disabled with --require-git."
+    )]
```
sharkdp, line 89: This contains several missing whitespace errors due to how you placed those backslashes (there should be a space in front). Please check the rendered version using `cargo run -- --help` or `fd --help`:
```
      --no-require-git
          By default, fd will only respect global gitignore rules, .gitignore rules,and local
          exclude rules if fd detects that you are searching inside agit repository. This flag
          allows you to relax this restriction such thatfd will respect all git related ignore
          rules regardless of whether you'researching in a git repository or not.
          
          This flag can be disabled with --require-git.
```

Also, I would prefer if we could put that long help text in the doc comment, like for most other options.

sharkdp (changes requested):
Thank you very much. Added two comments. Also, please read this section: https://github.com/sharkdp/fd/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog

src/cli.rs
vegerot: How's this look?

vegerot:
> Thank you very much. Added two comments. Also, please read this section: https://github.com/sharkdp/fd/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog

done ✅!  I also noticed in the doc you called it "Unreleased" but in CHANGELOG.md you called it "Upcoming release" so I updated that in CONTRIBUTING

tmccombs (approved):

tmccombs (approved):

tmccombs:
Merging is blocked for me. I think @sharkdp needs to approve because he requested changes.

sharkdp (approved):
Thank you for the updates!

jpcirrus:
This commit neglected to update the man page `doc/fd.1`.

vegerot:
@jpcirrus thanks for pointing this out!  Would you mind contributing a patch?

PR #1218, opened 2022-12-31
Draft: Add FD_CONFIG_PATH environment variable

vegerot:
Summary: Add FD_CONFIG_PATH environment variable which points to a file with config options that are prepended to any `fd` command.
This has the same behavior as `RG_CONFIG_PATH` from ripgrep

Test Plan: `cargo test`

TODO:
- add flag `--no-config` to not respect FD_CONFIG_PATH, `.fdignore`, or global
- add documentation
fd ignore
- make parsing config file blazingly fast
- if FD_CONFIG_PATH not set, or file does not exist, just use `args_os()`
directly
- fix `test_invalid_cwd`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -239,6 +239,7 @@ fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config
         ignore_hidden: !(opts.hidden || opts.rg_alias_ignore()),
         read_fdignore: !(opts.no_ignore || opts.rg_alias_ignore()),
         read_vcsignore: !(opts.no_ignore || opts.rg_alias_ignore() || opts.no_ignore_vcs),
+        require_git_to_read_vcsignore: !opts.no_require_git,
         read_parent_ignore: !opts.no_ignore_parent,
         read_global_ignore: !(opts.no_ignore
             || opts.rg_alias_ignore()
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,7 @@ use std::time;
 
 use anyhow::{anyhow, bail, Context, Result};
 use atty::Stream;
-use clap::{CommandFactory, Parser};
+use clap::{CommandFactory, FromArgMatches};
 use globset::GlobBuilder;
 use lscolors::LsColors;
 use regex::bytes::{Regex, RegexBuilder, RegexSetBuilder};
@@ -67,7 +67,7 @@ fn main() {
 }
 
 fn run() -> Result<ExitCode> {
-    let opts = Opts::parse();
+    let opts = parse_opts();
 
     #[cfg(feature = "completions")]
     if let Some(shell) = opts.gen_completions()? {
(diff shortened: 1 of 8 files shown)


vegerot:
@sharkdp @tmccombs I figured I'd just open this now to get some feedback on my approach :)

Happy New Year!

src/main.rs
```diff
@@ -105,6 +105,40 @@ fn run() -> Result<ExitCode> {
...
+fn parse_opts() -> Opts {
+    let mut cli_args = std::env::args_os(); // [/usr/bin/fd, --hidden, --etc]
+    let mut args: Vec<std::ffi::OsString> = args_from_config_file(); // [--no-require-git]
+
+    // TODO(vegerot): If `args` is empty we can just skip the next two lines and not pay the cost
+    // of moving things around
+    args.insert(0, cli_args.next().unwrap()); // [/usr/bin/fd, --no-require-git]
+
+    // this way, CLI arguments override config-file arguments
+    args.extend(cli_args.skip(1)); // [/usr/bin/fd, --no-require-git, --hidden, --etc]
```
tmccombs, line 117: perhaps:
```rust
cli_args.splice(1..1, args.into_iter())?
```

src/main.rs
```diff
@@ -105,6 +105,40 @@ fn run() -> Result<ExitCode> {
...
+
+    let mut matches = Opts::command().get_matches_from(args);
```
tmccombs, line 119: I think you could just do `Opts::parse_from(args)` for this and onward.

src/main.rs
```diff
@@ -105,6 +105,40 @@ fn run() -> Result<ExitCode> {
...
+}
+
+fn args_from_config_file() -> Vec<std::ffi::OsString> {
+    let config_path = match std::env::var_os("FD_CONFIG_PATH") {
+        Some(path) => path,
+        None => return vec![],
+    };
+
+    // TODO(vegerot): 🚀
+    match std::fs::read_to_string(config_path) {
```
tmccombs, line 136: Do we want to support comments in this file? Or allow leaving of the beginning "--"? Or restrict which options are allowed in here?

tests/testenv/mod.rs
```diff
@@ -161,6 +161,14 @@ impl TestEnv {
         }
     }
 
+    pub fn set_environment_variable(&self, key: &str, value: &str) {
```
tmccombs, line 164: These functions might not work well when multiple tests are run in parallel. Right now, it won't be an issue, because only a single test is using these functions. But if there are multiple tests, then one test might set this environment variable in one thread, then another test will overwrite that in another thread before the first test has a chance to execute `fd`. It would probably be better to set the enviornment when we spawn the `fd` process directly.

tmccombs: Actually... I think it could be a problem currently, because the test that uses FD_CONFIG_FILE could run at the same time as another test, and result in that test starting with the config file set, and inheriting the options from the config file.

tmccombs (changes requested):

fe53788347:
Actually... I think it could be a problem currently, because the test that uses FD_CONFIG_FILE could run at the same time as another test, and result in that test starting with the config file set, and inheriting the options from the config file.

fe53788347:
good

sharkdp:

Issue #1228, opened 2023-01-11
--exec changes order of results


dvogel:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

```
$ for x in `seq 100`; do touch $(cat /dev/urandom | tr -d -C 'A-Za-z' | head -c1); done
$ ls
a  A  B  c  C  d  D  e  f  F  G  i  I  j  k  K  L  m  M  n  N  o  O  p  P  q  Q  r  R  s  S  t  T  u  U  v  V  w  W  x  X  y  Y  z
$ fd -j1 -t f
A
B
C
D
F
G
I
K
L
M
N
O
P
Q
R
S
T
U
V
W
X
Y
a
c
d
e
f
i
j
k
m
n
o
p
q
r
s
t
u
v
w
x
y
z
$ fd -j1 -t f -x ls -1
./B
./V
./p
./n
./z
./s
./D
./P
./u
./d
./R
./G
./X
./Y
./t
./i
./j
./F
./M
./U
./f
./Q
./e
./r
./O
./q
./T
./A
./K
./I
./c
./C
./N
./W
./w
./v
./y
./k
./L
./m
./x
./S
./o
./a
```

### Describe what you expected to happen:

I would expect these commands to output the same files in the same order:

```
fd -j1 -t f
```

```
fd -j1 -t f -x ls -1
```

The `./` prefix is fine. It is the order that seems like a bug to me.

### What version of `fd` are you using?

8.3.0 and 8.6.0

### Which operating system / distribution are you on?

```shell
$ uname -srm
Linux 6.0.0-1009-oem x86_64
```

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```
```

tmccombs:
Here is why I think you are seeing the results in the order you do for each command:

`ls`: is sorting the results alphabetically (and mostly case insensitve, unless the names differ only in case).
`fd -j 1`: Since there are less than 1000 results, and they are returned fairly quickly, the results are sorted before printing (but case sensitively)
`fd -j 1 -x ls`: Is not doing any sorting. Since you have a single thread, the commands are run in the order that they are returned by the file system.

dvogel:
>  fd -j 1: Since there are less than 1000 results, and they are returned fairly quickly, the results are sorted before printing (but case sensitively)

[...]

PR #1230, opened 2023-01-13
Explain -x and -X result order.

dvogel:
This is meant to smooth over a sharp edge I ran into and reported in issue #1228. I think it's important to explain the non-sorted order of the results specifically related to `-x` because the output can look very much like a `-j` bug where `-j1` isn't really forcing sequential order.

Screenshot from 2023-01-13 09-32-13

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -769,9 +769,10 @@ impl clap::Args for Exec {
                 .conflicts_with("list_details")
                 .help("Execute a command for each search result")
                 .long_help(
-                    "Execute a command for each search result in parallel (use --threads=1 for sequential command execution). \
-                     All positional arguments following --exec are considered to be arguments to the command - not to fd. \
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -801,7 +802,8 @@ impl clap::Args for Exec {
                 .conflicts_with_all(["exec", "list_details"])
                 .help("Execute a command with all search results at once")
                 .long_help(
-                    "Execute the given command once, with all search results as arguments.\n\
+                    "Execute the given command once, with all search results as arguments, in the order \
```
tmccombs, line 805: "in the order they were found" is only true if `-j 1` is specified. In general the order is non-deterministic. And even in the `-j 1` case, the actual order is an implementation detail.  I'd rather say something to the effect of "You sholdn't rely on the order the results are given in".

sharkdp:

Issue #880, opened 2021-11-03
Find by inode number


sucrecacao:
Hi!

Thanks for this amazing tool :) I use it everyday

legacy `find` has an option to search by inode number ( `find -inum XXXXX`)  but it is terribly slow.

Would it be possible to implement this feature in `fd`, and if so, would it be significantly faster than `find` ?

tavianator:
Can you describe your use case for this?  Is it something like #300/#751?

I don't think we're likely to implement something like `-inum`, and `--samefile` has been rejected before.

sucrecacao:
>  Is it something like #300/#751?

No

> Can you describe your use case for this?

Searching by inode can be extremely usefull in various usecase, and it is a general purpose rather than specific usecase tool. But where I find it most usefull, is to keep track of files in media library ( image, music, video, torrent). A user can freely move files around and  rename them; the software will still be able to find the file, even if its path or name has changed.

sucrecacao:
Would it be easy to print the inode instead of the filepath in `fd` ? 

Then it can be used with `grep` to quickly find the inode

tavianator:
> Searching by inode can be extremely usefull in various usecase, and it is a general purpose rather than specific usecase tool.

Sure, but features are only added to `fd` if they have compelling usecases.  I personally have never used `find -inum` except to test that it works.  So I was curious why others find it useful.

> But where I find it most usefull, is to keep track of files in media library ( image, music, video, torrent). A user can freely move files around and rename them; the software will still be able to find the file, even if its path or name has changed.

I guess my question is, how do you even know the inode number of the file you're looking for?  Do you memorize them?  Or do you have a database that stores them?

> Would it be easy to print the inode instead of the filepath in fd ?  Then it can be used with grep to quickly find the inode

It would be possible.  I don't think it's a feature we're likely to implement though.  You could try `fd -X ls -1di` or similar to have `ls` do it for you, but I'm not sure how performant that is.

sucrecacao:
>  I personally have never used find -inum except to test that it works. So I was curious why others find it useful.

yes, you would not definitly never use as you use `fd` for working workflow. You would most probably use this feature inside a software or a script only. 

> I guess my question is, how do you even know the inode number of the file you're looking for? Do you memorize them? Or do you have a database that stores them?

Yes, you track the inode of set of files (e.g. when adding a torrent, when import photo/music to you media library ). You store there inode and path. If there are not located at the usual path, you start inode search. 

[...]

PR #1248, opened 2023-01-30
Add option to filter files by inode number

cgzones:
Add the option on Unix systems to filter files by their inode number.

The Windows equivalent FileIndex is not yet stabilized, see https://github.com/rust-lang/rust/issues/63010.

This is especially useful to debug audit records, e.g.:

    Jan 30 17:48:55 laptop audit: PATH item=0 name="pulse" inode=7340042 dev=fe:03 mode=040700 ouid=1001 ogid=1001 rdev=00:00 obj=system_u:object_r:unlabeled_t:s0 nametype=NORMAL cap_fp=0 cap_fi=0 cap_fe=0 cap_fver=0 cap_frootid=0

Closes: https://github.com/sharkdp/fd/issues/880

Maybe this option should enable `--hidden` by default?

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,7 @@ Options:
   -t, --type <filetype>            Filter by type: file (f), directory (d), symlink (l),
                                    executable (x), empty (e), socket (s), pipe (p)
   -e, --extension <ext>            Filter by file extension
+      --inum <num>                 Filter by inode number
   -S, --size <size>                Limit results based on the size of files
       --changed-within <date|dur>  Filter by file modification time (newer than)
       --changed-before <date|dur>  Filter by file modification time (older than)
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -243,6 +243,9 @@ Always colorize output.
 .BI "\-j, \-\-threads " num
 Set number of threads to use for searching & executing (default: number of available CPU cores).
 .TP
+.BI "\-\-inum " num
+Filter files by their inode number.
+.TP
 .BI "\-S, \-\-size " size
 Limit results based on the size of files using the format
 .I <+-><NUM><UNIT>
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -365,6 +365,20 @@ pub struct Opts {
     )]
     pub extensions: Option<Vec<String>>,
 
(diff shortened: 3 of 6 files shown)


tavianator:
IMHO we should decline this feature request.  But thank you for writing the pull request!  Contributions are always appreciated even if we don't ultimately accept them.  More info here: https://github.com/sharkdp/fd/issues/880#issuecomment-1775789788

kr-nn:
I don't quite follow why this is being declined. There's IMHO no reason to not accept this.

according to this issue:https://github.com/sharkdp/fd/issues/880
> features are only added to fd if they have compelling usecases

The compelling usecase is finding files that share an inode.
Off the top of my head the reasons this would be useful:
1. Tracking disk usage being saved across a directory via hardlinking
2. Finding which files you want to delete without losing the data
3. Confirming if the correct files are hardlinked in a series of files
4. Managing media libraries/knowledgebase articles/etc. that utilize hardlinks
5. Being a `A simple, fast and user-friendly alternative to 'find'`

I can't think of a reason why this shouldn't be a feature of fd. What's the risk of adding it?

> I personally have never used find -inum except to test that it works. So I was curious why others find it useful.

There's plenty of features of many commands I haven't used. Just because it doesn't fit _your_ usecases doesn't mean it shouldn't belong.

> I guess my question is, how do you even know the inode number of the file you're looking for? Do you memorize them? Or do you have a database that stores them?

you can run `stat ./file.md` and it'll feed you the inode. You can use that plus `find -inum <inodenum>` to find other files that share the inode. I assumed because fd is `A simple, fast and user-friendly alternative to 'find'` that fd would have this implemented.

Apologies if I come off as aggressive or entitled, I'm just at a loss for why this isn't implemented, especially when someone already did the work to implement it.

tmccombs:
Adding any additional option to fd makes it less "simple".  Especially a fairly niche use case.

And I think this is less user friendly than say finding files that have the same inode/are hardlinked to some other file.

And fwiw, this implementation isn't as fast as it could be, because it unnecessarily makes a stat call to get the inode, instead of using the inode from the DirEntry directly.

PR #1272, opened 2023-03-02
Allow exclude/ignore by absolute path

### LINKED ISSUE #851: `--exclude` doesn't work with absolute paths

The [examples](https://github.com/sharkdp/fd#excluding-specific-files-or-directories) for the `--exclude` or `-E` option imply that it should work with absolute paths (`/mnt/external-drive` is given as an example). However, it only seems to work with relative paths. For example, if I'm trying to exclude the directory `/home/user/Library/`:
```
fd -E Library pattern /home
```
works, as does
```
fd -E "*/Library/*" pattern /home
```

However,
```
fd -E /home/user/Library/ pattern /home
```
doesn't work (i.e. `/home/user/Library/pattern.txt` would still show up in search results ). Adding other options, such as `-p` or `-a` doesn't seem to affect this behavior.

The only way I've found to exclude absolute paths is to add them to `~/.config/fd/ignore`, which is sowewhat inconvenient.

```
fd 8.2.1
```

```
Linux 5.14.2-arch1-2 x86_64
LSB Version:    1.4
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a
```

cyqsimon:
Closes #851.

This is a crude implementation of my idea in https://github.com/sharkdp/fd/issues/851#issuecomment-1446215990. I don't think it's ideal (far from it), but at least we have some kind of support for people who want it.

I'll keep it a draft for now. Please chip in your ideas on how this could be improved.

## Main sore spots
- Performance penalty due to path canonicalization (although this is unavoidable)
- Globs work somewhat differently due to implementation
  - Excluding a directory doesn't cause its children to be excluded too
    - To exclude `/foo/bar` and all its children, it's necessary to specify a glob of `/foo/bar/*`
- The introduction of a separate flag and global ignore file for this purpose
  - `--exclude-absolute` and `~/.config/fd/ignore-absolute`

## Todo list
- [x] Implementation
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contrib/completion/_fd
+++ b/contrib/completion/_fd
@@ -110,6 +110,7 @@ _fd() {
     '*'{-t+,--type=}"[filter search by type]:type:(($fd_types))"
     '*'{-e+,--extension=}'[filter search by file extension]:extension'
     '*'{-E+,--exclude=}'[exclude files/directories that match the given glob pattern]:glob pattern'
+    '*--exclude-absolute=[exclude files/directories whose absolute path match the given glob pattern]:glob pattern'
     '*'{-S+,--size=}'[limit search by file size]:size limit:->size'
     '(-o --owner)'{-o+,--owner=}'[filter by owning user and/or group]:owner and/or group:->owner'
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -46,7 +46,7 @@ pub struct Opts {
     no_hidden: (),
 
     /// Show search results from files and directories that would otherwise be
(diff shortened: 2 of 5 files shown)


cyqsimon:
This implementation is rubbish (as is usually the judgement when you look at what you wrote 3 years ago).

Anyone who wishes to use the code in this PR (not sure why you would) is free to do so, subject to the same license of this project.

PR #1284, opened 2023-03-20
Cargo update

tmccombs:


sharkdp:
Thank you. This seems to include the `ignore` update from #1250, which I would have liked to review in more detail.

Sure, we could always put all of our trust in the integration/regression tests that we have, but this dependency is so crucial for `fd` that I usually make an exception and at least scan the ChangeLog. @tmccombs What do you think?

tmccombs:
Oh, I missed that it included that. I can look through the change log for it tonight. I'd also be ok reverting this merge until that has been reviewed more.

sharkdp:
> I can look through the change log for it tonight.

No rush. I just commented because I suspected that it might have been an oversight. If you want to review the changelog, that would be very much appreciated. But I didn't want to suggest that it would have to be done by you.

tavianator:
I reviewed [the diff](https://github.com/BurntSushi/ripgrep/compare/ignore-0.4.18...ignore-0.4.20#diff-43083856729390983c2f2dc067ec1840f39b6b349a4d6872664bd5d9cafbb2e2).  Only things that stick out are

- Avoiding some `stat()` calls in `--no-require-git` mode: https://github.com/BurntSushi/ripgrep/pull/2052
- Switching to std scoped threads: https://github.com/BurntSushi/ripgrep/commit/e95254a86f484eec663be4714924d91d3cf39cb8

Looks fine to me.  But yeah in the future I will check more carefully if `ignore` is updated :)

sharkdp:
Thank you. I guess this means we can close the other PR.

tmccombs:
I looked through it to, and I didn't see anything else of note.

Issue #1282, opened 2023-03-19
[BUG] man page missing `--and` flag


jellopuddingstick:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

It looks like #1139 forgot to update the man page to include `--and`:

```
$ man fd | rg -- '--and'
$ fd --help | rg -- '--and'
      --and <pattern>
```

### Describe what you expected to happen:

`man fd` should include all available flags.

### What version of `fd` are you using?

fd 8.6.0

### Which operating system / distribution are you on?

```shell
$ uname -srm
Linux 6.2.7-2-MANJARO x86_64
$ lsb_release -a
LSB Version:	n/a
Distributor ID:	ManjaroLinux
Description:	Manjaro Linux
Release:	22.0.5
Codename:	Sikaris
```

PR #1285, opened 2023-03-21
Add documentation and zsh updates

tmccombs:
- Include --and in man page and zsh completions.
- Include --no-require-git and --changed-after to zsh completions

Fixes: #1282


sharkdp (approved):
Thank you!

PR #1302, opened 2023-04-14
Don't use jemalloc on OpenBSD (jemalloc fails to compile)

n1000:
jemalloc fails to compile on OpenBSD, so disable it there by default.

The current OpenBSD ports script [already does this](https://github.com/openbsd/ports/blob/55cab6ed35e0902edf2014de1941c7a4bf210061/sysutils/fd/Makefile#L18), but committing this here makes it easier to build from source or to install fd via cargo if desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ libc = "0.2"
 # FIXME: Re-enable jemalloc on macOS
 # jemalloc is currently disabled on macOS due to a bug in jemalloc in combination with macOS
 # Catalina. See https://github.com/sharkdp/fd/issues/498 for details.
-[target.'cfg(all(not(windows), not(target_os = "android"), not(target_os = "macos"), not(target_os = "freebsd"), not(all(target_env = "musl", target_pointer_width = "32")), not(target_arch = "riscv64")))'.dependencies]
+[target.'cfg(all(not(windows), not(target_os = "android"), not(target_os = "macos"), not(target_os = "freebsd"), not(target_os = "openbsd"), not(all(target_env = "musl", target_pointer_width = "32")), not(target_arch = "riscv64")))'.dependencies]
 jemallocator = {version = "0.5.0", optional = true}
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


tmccombs (approved):

sharkdp:
Thank you. At some point we should think about using a whitelist instead of a blacklist :-)

n1000:
Thanks. BTW, as a side note, it looks like the [following commit](https://github.com/jemalloc/jemalloc/commit/58478412be842e140cc03dbb0c6ce84b2b8d096e) in the jemalloc source fixes the compilation issue on OpenBSD:

From `git@github.com:jemalloc/jemalloc.git`:
```
commit 58478412be842e140cc03dbb0c6ce84b2b8d096e (HEAD)
Author: David Carlier <devnexen@gmail.com>
Date:   Tue Jun 7 21:43:08 2022 +0100

    OpenBSD build fix. still no cpu affinity.

    - enabling pthread_get/pthread_set_name_np api.
    - disabling per thread cpu affinity handling, unsupported on this platform.
```

So maybe someday in the future, if the jemalloc submodule inside the jemalloc-sys crate is updated, jemalloc support could be re-enabled for OpenBSD. I did not extensively test on OpenBSD with the above change included in the jemalloc build, but basic fd functionality seemed to work fine with jemalloc and the above change in place.

Obviously, updating the jemalloc-sys crate is outside the scope of fd, but maybe I can raise an issue over there to ask about updating to a newer version of jemalloc.

---

Perf delta with / without jemalloc enabled, on OpenBSD 7.3:

Without jemalloc enabled:
```console
Benchmark 1: /home/nathan/.cargo/bin/fd
  Time (mean ± σ):      4.040 s ±  0.121 s    [User: 2.311 s, System: 5.610 s]
  Range (min … max):    3.898 s …  4.258 s    10 runs
```

With jemalloc enabled:
```console
Benchmark 1: /home/nathan/src/fd/target/release/fd
  Time (mean ± σ):      3.654 s ±  0.094 s    [User: 1.921 s, System: 5.196 s]
  Range (min … max):    3.507 s …  3.808 s    10 runs
```

PR #1314, opened 2023-05-04
feat: Add a way to escape { in exec templates.

### LINKED ISSUE #1303: Exec: Allow literal placeholder {}
Sometimes a literal {} is needed that doesn't get parsed as placeholder. 

## Use-case

I wanted to refactor my go code where `interface{}` can be now written as `any`. I therefore wanted to search-replace all occurences using `fd` and `sed`. I couldn't find a way telling `fd` to not interpret the `{}` as placeholder.

### Reproducer

```go
fd -e go -X echo "interface{}"
```

Within my folder with only a `hello.go` file, this just outputs: `interface./hello.go`

My goal would be: `interface{} hello.go`

## Proposed Solution

Allowing `{{}}` which gets detected as something like `LITERAL` and gets replaced with `{}` should work.

tmccombs:
fixes #1303

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -376,9 +376,14 @@ parent directory
 path without file extension
 .IP {/.}
 basename without file extension
+.IP {{}
+literal '{'
 .RE
 
 If no placeholder is present, an implicit "{}" at the end is assumed.
+
+If you need to include the literal text of one of the placeholders, you can use "{{}" to
+escape the first "{". For example "{{}}" expands to "{}", and "{{}{{}}}" expands to "{{}".
 
 Examples:
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -799,7 +799,8 @@ impl clap::Args for Exec {
                        '{/}':  basename\n  \
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
-                       '{/.}': basename without file extension\n\n\
+                       '{/.}': basename without file extension\n  \
+                       '{{}':  literal '{' (for escaping)\n\n\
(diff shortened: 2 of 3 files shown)


tavianator:
I think doing it [like Rust](https://doc.rust-lang.org/std/fmt/index.html#escaping) (`{{` -> `{`, `}}` -> `}`) would be nice, but maybe that's too much of a compatibility hazard?

tmccombs:
I briefly considered that, but maybe it deserves more thought. "{{" would be pretty simple to do. "}}" is a little more difficult, I can't just add it to the regex because then something like  "{}}" would match the "{}" before the "}}". A lookahead rule would fix that, but the regex crate doesn't support lookaheads.

I can think of a few ways to implement it:
1. Support "}}" but only if it doesn't overlap with one of the other patterns
2. Roll our own state machine to find the patterns instead of using regex
3. Have the regex match `[^}]|$` at the end, and have some logic to exclude that somehow (including backtracking the search one character unless it is at the end).

Of those I think I prefer 2.

tavianator:
I think you can do it with a regex by taking advantage of preference order.  `{{|}}|{}` will match the escaped `{{` or `}}` preferentially before it will match a `{}`.

tmccombs:
No: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9a2ea3ddff0596894ff5439d899463b4

I think it matches "{}" at the current position, so it doesn't try matching at the next character.

tavianator:
Oh I see.  It would work for Rust because `{}}` is a syntax error, every brace needs to be either escaped or part of a `{}` substitution.

sharkdp:
Let's close this in favor of #1318?

PR #1318, opened 2023-05-20
feat: Add a way to escape { and }in exec templates.

### LINKED ISSUE #1303: Exec: Allow literal placeholder {}

tmccombs:
fixes #1303 

Alternative to #1314

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -380,9 +380,14 @@ parent directory
 path without file extension
 .IP {/.}
 basename without file extension
+.IP {{}
+literal '{'
 .RE
 
 If no placeholder is present, an implicit "{}" at the end is assumed.
+
+If you need to include the literal text of one of the placeholders, you can use "{{}" to
+escape the first "{". For example "{{}}" expands to "{}", and "{{}{{}}}" expands to "{{}".
 
 Examples:
 
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -380,14 +380,16 @@ parent directory
 path without file extension
 .IP {/.}
 basename without file extension
-.IP {{}
-literal '{'
+.IP {{
+literal '{' (an escape sequence)
+.IP }}
+literal '}' (an escape sequence)
 .RE
 
 If no placeholder is present, an implicit "{}" at the end is assumed.
 
-If you need to include the literal text of one of the placeholders, you can use "{{}" to
-escape the first "{". For example "{{}}" expands to "{}", and "{{}{{}}}" expands to "{{}".
+Notice that you can use "{{" and "}}" to escape "{" and "}" respectively, which is especially
+useful if you need to include the literal text of one of the above placeholders.
 
 Examples:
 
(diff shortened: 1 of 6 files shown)


doc/fd.1
```diff
@@ -376,10 +376,15 @@ parent directory
...
 .IP {/.}
 basename without file extension
+.IP {{}
+literal '{'
 .RE
 
 If no placeholder is present, an implicit "{}" at the end is assumed.
 
+If you need to include the literal text of one of the placeholders, you can use "{{}" to
+escape the first "{". For example "{{}}" expands to "{}", and "{{}{{}}}" expands to "{{}".
```
sharkdp, line 386: @tmccombs Thank you very much for working on this and for even providing two implementations.

If I'm being really honest, this paragraph sounds incredible confusing at first. It does make sense and is logically consistent, but it's.. certainly not easy to grasp. This is a quite complex escaping rule. And one that I haven't seen in use anywhere else.

Given that you now implemented this with a custom tokenizer, is there still no way to implement Rusts `{{` escaping or something like `\{`?

What does `parallel` do (we borrowed the `--exec` syntax from them)?

tmccombs: oops. This branch actually does implement using `{{` and `}}` for escapes. I just forgot to update the help and documentation  to match :facepalm:.

tavianator: > What does `parallel` do (we borrowed the `--exec` syntax from them)?

I don't think you can escape braces in GNU parallel, but you can change the `{}` to a different string: https://stackoverflow.com/questions/42532693/

src/cli.rs
```diff
@@ -829,7 +831,8 @@ impl clap::Args for Exec {
                        '{/}':  basename\n  \
                        '{//}': parent directory\n  \
                        '{.}':  path without file extension\n  \
-                       '{/.}': basename without file extension\n\n\
+                       '{/.}': basename without file extension\n  \
+                       '{{}':  literal '{' (for escaping)\n\n\
```
sharkdp, line 835: is this still the old syntax? (comparing it with the above)

sharkdp (approved):
Thank you very much.

Just one minor comment.

src/cli.rs
tmccombs: Yeah, oops. Thanks for fixing that.

PR #1333, opened 2023-06-08
Rust 1.70 upgrade

tmccombs:
Rust 1.70 allows us to remove a couple of dependencies. But maybe we should wait a little bit before increasing the MSRV


tavianator (approved):

sharkdp:
Oh, great - thank you!

PR #1350, opened 2023-07-18
Regex upgrade

tmccombs:


sharkdp (approved):
Thank you

tavianator:
This needs a rebase but then I think we should merge it.  Looks like the `regex` crate has stabilized since the big changes of 1.8/1.9

PR #1366, opened 2023-08-23
Add a `--quote` option to put quotation marks around output files/dirs

### LINKED ISSUE #1365: Add an option to quote files in output for easier copy/pasting
I have some files with spaces/quotes in the names and I would like `fd` to quote the output files appropriately to make copy/pasting/chaining easier. My use case is to search for a couple of files and then copy/paste a subset of the found files to another command. If the file contains special characters it makes copy/pasting more complicated.

If I use vanilla `fd nova` I get:

```
nova - permissions.odt
nova-status.ods
```

The first will require quoting to work with the spaces, and the second will not. 

If use `fd nova -l` I get:

```
-rw-rw-r--. 1 bakers bakers  13K Apr  5  2011 './nova - permissions.odt'
-rw-rw-r--. 1 bakers bakers 9.8K Apr  6  2006  ./nova-status.ods
```

Note that the filename is automatically quoted as needed, which is what is 90% of I want. Is there a way to get this functionality **without** the `ls` style details? I only need the filename.

scottchiefbaker:
Following the rules set by [coreutils - quoting](https://www.gnu.org/software/coreutils/quotes.html) I have added a `--quote` option to put appropriate quotation marks around files/directories that need them. 

**Note:** This fixes #1365

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -54,6 +54,43 @@ fn print_trailing_slash<W: Write>(
     Ok(())
 }
 
+// Trying to copy: https://www.gnu.org/software/coreutils/quotes.html
+fn path_needs_quoting(path: &str) -> i8 {
+    // If it contains any special chars we return single quote
+    if path.contains(" ") || path.contains("$") || path.contains("\"") {
+        return 1;
+    // If it ONLY contains a ' we return double quote
+    } else if path.contains("'") {
+        return 2;
+    }
+
+    return 0;
+}
+
+// Quote a path with coreutils style quoting to make copy/paste
+// more friendly for shells
+fn quote_path(path_str: &str) -> String {
+    let quote_type         = path_needs_quoting(path_str);
+    let mut tmp_str:String = path_str.into();
+
+    // Quote with single quotes
+    if quote_type == 1 {
+        // Escape single quotes in path
+        tmp_str = str::replace(&tmp_str, "'", "'\\''");
+
+        format!("'{}'", tmp_str)
+    // Quote with double quotes
+    } else if quote_type == 2 {
+        // Escape double quotes in path
+        tmp_str = str::replace(&tmp_str, "\"", "\\\"");
+
+        format!("\"{}\"", tmp_str)
+    // No quoting required
+    } else {
+        path_str.to_string()
+    }
+}
+
 // TODO: this function is performance critical and can probably be optimized
 fn print_entry_colorized<W: Write>(
     stdout: &mut W,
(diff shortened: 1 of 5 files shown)


scottchiefbaker:
Also FWIW I've never written **any** rust before yesterday, so I'm pretty confident my code could be improved. Be gentle

scottchiefbaker:
I would like to request some help with the unit tests. I have written the tests, but they're not working. Specifically it's not engaging the new `--quote` feature required for the test to complete successfully. Can someone with more experience take a look at my unit tests and tell me what I'm doing wrong.

src/output.rs
```diff
@@ -54,6 +54,43 @@ fn print_trailing_slash<W: Write>(
     Ok(())
 }
 
+// Trying to copy: https://www.gnu.org/software/coreutils/quotes.html
+fn path_needs_quoting(path: &str) -> i8 {
+    // If it contains any special chars we return single quote
+    if path.contains(" ") || path.contains("$") || path.contains("\"") {
+        return 1;
```
tmccombs, line 61: It would be better to use an enumeration instead of an `i8`, where it isn't obvious what the different numbers mean.

src/output.rs
```diff
@@ -54,6 +54,43 @@ fn print_trailing_slash<W: Write>(
...
+        format!("'{}'", tmp_str)
+    // Quote with double quotes
+    } else if quote_type == 2 {
+        // Escape double quotes in path
+        tmp_str = str::replace(&tmp_str, "\"", "\\\"");
+
+        format!("\"{}\"", tmp_str)
+    // No quoting required
+    } else {
+        path_str.to_string()
```
tmccombs, line 90: This unnecessarily copies the string again even if there is no change

src/output.rs
```diff
@@ -62,9 +99,22 @@ fn print_entry_colorized<W: Write>(
...
+    let mut path_str      = path.to_string_lossy();
+    let mut needs_quoting = false;
+
+    // Wrap the path in quotes
+    if config.use_quoting {
+        let tmp_str = quote_path(&path_str);
+
+        // If the quoted string is new, then we flag that to tweak the offset
+        // so the colors line up
+        if tmp_str != path_str {
```
tmccombs, line 113: This has to compare every character of the string. It would be more efficient if you did something like have return `quote_path` return an `Option` that is only Some if quoting happened, and then do:

```rust
if let Some(quoted) = quote_path(&path_str) {
    path_str = quoted.into();
    needs_quoting = true;
}
```

src/output.rs
```diff
@@ -78,6 +128,11 @@ fn print_entry_colorized<W: Write>(
     }
 
     if offset > 0 {
+
+        if needs_quoting {
+            offset += 2;
```
tmccombs, line 133: I don't think this is correct. If it is only quoted shouldn't the offset be just 1 more? 

And if it is double quoted, and has escaped characters, then the proper offset will depend on how many characters in the parent ended up getting escaped.

src/output.rs
```diff
@@ -54,6 +54,43 @@ fn print_trailing_slash<W: Write>(
     Ok(())
 }
 
+// Trying to copy: https://www.gnu.org/software/coreutils/quotes.html
+fn path_needs_quoting(path: &str) -> i8 {
+    // If it contains any special chars we return single quote
+    if path.contains(" ") || path.contains("$") || path.contains("\"") {
```
tavianator, line 60: This is probably not careful enough.  See https://github.com/tavianator/bfs/blob/b8ed989642b9f0f6c1301bcff6f1498935cbd81c/src/bfstd.c#L906-L918 for example

scottchiefbaker:
My code is definitely sub par. This was my proof-of-concept trial. Someone with real Rust experience should tackle this.

PR #1396, opened 2023-10-15
feat: ignore `.git/` by default

### LINKED ISSUE #1387: Include `.git/` folder in the "gitignored" filtering
I mostly use `fd` in repositories, and I love having the ability of gitignoring pollution from build folders.

However, I keep getting tons of stuff from `.git/`.

I consider `.git/` to be _"gitignored by default"_, but that might just be my POV, any thoughts on this?

I imagine the odds of someone intentionally looking for stuff in the `.git/` directory should be low, but in that case, they can resort to `-I/--no-ignore` (or `--no-ignore-vcs`).

skoriop:
Closes #1387 (**breaking change!**)

`.git/` is now ignored by default even with `--hidden` / `-H`.

Use `--no-ignore` / `-I` or `--no-ignore-vcs` to override.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -66,6 +66,13 @@ pub fn scan(paths: &[PathBuf], patterns: Arc<Vec<Regex>>, config: Arc<Config>) -
             .add(pattern)
             .map_err(|e| anyhow!("Malformed exclude pattern: {}", e))?;
     }
+
+    if config.read_vcsignore {
+        override_builder
+            .add("!.git")
+            .map_err(|e| anyhow!("Malformed default exclude pattern: {}", e))?;
+    }
+
     let overrides = override_builder
         .build()
         .map_err(|_| anyhow!("Mismatch in exclude patterns"))?;
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -70,7 +70,7 @@ pub fn scan(paths: &[PathBuf], patterns: Arc<Vec<Regex>>, config: Arc<Config>) -
     if config.read_vcsignore {
         override_builder
             .add("!.git")
-            .map_err(|e| anyhow!("Malformed default exclude pattern: {}", e))?;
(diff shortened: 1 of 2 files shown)


src/walk.rs
```diff
@@ -66,6 +66,13 @@ pub fn scan(paths: &[PathBuf], patterns: Arc<Vec<Regex>>, config: Arc<Config>) -
             .add(pattern)
             .map_err(|e| anyhow!("Malformed exclude pattern: {}", e))?;
     }
+
+    if config.read_vcsignore {
+        override_builder
+            .add("!.git")
+            .map_err(|e| anyhow!("Malformed default exclude pattern: {}", e))?;
```
tmccombs, line 73: Since we are directly providing the pattern here, and know that it is a valid pattern, I think it would be more appropriate to use expect here:

```suggestion
           .expect("Invalid exclude pattern"));
```

tmccombs (changes requested):
I am in favor of including this change. 

I left one small bit of feedback.

src/walk.rs
skoriop: Yup, that makes sense

tmccombs (approved):

sharkdp:
Thank you for making this change. I'm also in favor of it. But I do think that this needs a bit of documentation and testing. I would suggest we adapt the `--help` texts for `--hidden` to reflect these changes. Otherwise, I think this can be very surprising.

As for the tests, I think we should at least add three simple checks that would assert that:
- results in `.git` *are not* included when using `--hidden`
- - results in `.git` *are not* included when using `--no-ignore`
- results in `.git` *are* included when using `--hidden --no-ignore`

At least that is my understanding how this feature should work?

skoriop:
Yup, that is how it should work.

tmccombs:
I'll write some tests and docs.

PR #1402, opened 2023-10-22
maint: Remove use of unmaintained users crate

tmccombs:
The users crate is no longer maintained. There is a maintained fork called uzers, but we already have a dependency on nix, and that has the functionality we need.


sharkdp (approved):
Great, thank you!

PR #1403, opened 2023-10-22
Add docs and tests for ignoring .git

tmccombs:
See #1396.

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -29,11 +29,22 @@ By default
 .B fd
 uses regular expressions for the pattern. However, this can be changed to use simple glob patterns
 with the '\-\-glob' option.
+.P
+By default
+.B fd
+will exclude hidden files and directories, as well as any files that match gitignore rules
+or ignore rules in .ignore or .fdignore files. For convenenience, '.git' is treated as if it
+was always included in gitignore rules. These files can be included with options such as
+'\-\-hidden' and '\-\-no\-ignore'.
 .SH OPTIONS
 .TP
 .B \-H, \-\-hidden
 Include hidden files and directories in the search results
 (default: hidden files and directories are skipped). The flag can be overridden with '--no-hidden'.
+.IP
+Ignored files and .git/ are still excluded unless \-\-no\-ignore or \-\-no\-ignore\-vcs
+is also used. The .git directory is excluded, because that is usually what is wanted if excluding vcs
+ignored files, and is more consistent with the output of most git commands.
 .TP
 .B \-I, \-\-no\-ignore
 Show search results from files and directories that would otherwise be ignored by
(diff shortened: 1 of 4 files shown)


doc/fd.1
```diff
@@ -29,11 +29,22 @@ By default
...
+'\-\-hidden' and '\-\-no\-ignore'.
 .SH OPTIONS
 .TP
 .B \-H, \-\-hidden
 Include hidden files and directories in the search results
 (default: hidden files and directories are skipped). The flag can be overridden with '--no-hidden'.
+.IP
+Ignored files and .git/ are still excluded unless \-\-no\-ignore or \-\-no\-ignore\-vcs
+is also used. The .git directory is excluded, because that is usually what is wanted if excluding vcs
+ignored files, and is more consistent with the output of most git commands.
```
marcospb19, line 47: I usually expect the program behavior to be explained in man pages, but not the reasons behind the behavior/choices. I'd say this is not unnecessary.

tests/tests.rs
```diff
@@ -2554,3 +2554,27 @@ fn test_invalid_cwd() {
...
+         .git/.foo
+         .git/a.foo",
+    );
+    te.assert_output(
+        &["--hidden", "--no-ignore-vcs", "foo"],
+        ".git/one/foo.a
+         .git/.foo
+         .git/a.foo",
+    );
+}
```
marcospb19, line 2580: Could you edit this test (or add a new one) to cover this example (or an equivalent one)?

```
let te = TestEnv::new(
    &["path/to/another/repository/.git/file1"],
    &["path/to/another/repository/.git/folder/file2"],
    &["a_repository/.git/file1"],
    &["a_repository/.git/folder/file2"],
);
```

Because the current tests are only covering `.git` in the current folder.

src/cli.rs
```diff
@@ -61,7 +64,8 @@ pub struct Opts {
     ignore: (),
 
     ///Show search results from files and directories that would otherwise be
-    /// ignored by '.gitignore' files. The flag can be overridden with --ignore-vcs.
+    /// ignored by '.gitignore' files or the rule to exclude .git/.
+    /// The flag can be overridden with --ignore-vcs.
```
marcospb19, line 68: My suggestion:
```
    /// Show search results from '.git/' folders and results that would
    /// otherwise be ignored by '.gitignore' rules.
```

src/cli.rs
```diff
@@ -46,7 +48,8 @@ pub struct Opts {
     no_hidden: (),
 
     /// Show search results from files and directories that would otherwise be
-    /// ignored by '.gitignore', '.ignore', '.fdignore', or the global ignore file.
+    /// ignored by '.gitignore', '.ignore', '.fdignore', the global ignore file,
+    /// or the rule to exclude .git/.
```
marcospb19, line 52: My suggestion:

```suggestion
    /// ignored by '.gitignore', '.ignore', '.fdignore', the global ignore file,
    /// or the default rule that excludes '.git/'.
```

marcospb19 (approved):

sharkdp (approved):
Thank you very much

PR #1405, opened 2023-10-23
docs: Explicitly warn against relying on result order for exec

tmccombs:
Supersedes: #1230


tmccombs:
Should I soften the language to use "in general" since there are cases when it is sorted?

sharkdp:
> Should I soften the language to use "in general" since there are cases when it is sorted?

The only case I have in mind relies on timing? (if results come in faster than X milliseconds, we sort them before printing/using in exec)

And relying on timing is certainly not a good idea, so I think your advise is good.

PR #1410, opened 2023-10-25
Replace num_cpus with `std::thread::available_parallelism`

tmccombs:
I think this might help startup time a little bit, since it looks like on linux at least the std implementation uses syscalls to determine the parallelism, whereas num_cpus was processing the /proc/cpuinfo file.

It also removes another dependency.


src/cli.rs
```diff
@@ -719,6 +718,16 @@ impl Opts {
     }
 }
 
+/// Get the default number of threads to use, if not explicitly specified.
+fn default_num_threads() -> usize {
+    // If we can't get the amount of parallelism for some reason, then
+    // default to a single thread, because that is safe.
+    const FALLBACK_PARALLELISM: NonZeroUsize = NonZeroUsize::MIN;
```
tavianator, line 725: This is a long-winded way of writing 1, right?

tmccombs: Indeed. Maybe it would be more readable if I changed this to just 1, and used a pattern match? 

I debated having the threads method return a NonZeroUsize, but wasn't sure if there was any benefit.

sharkdp (approved):
Thank you.

Issue #1203, opened 2022-12-06
fd starts slowly on systems with very many cores


tavianator:
_Originally posted by @pmarreck in https://github.com/sharkdp/fd/issues/1131#issuecomment-1336197854_

> I use `fd` in some test suites and I noticed a slowdown. I investigated and I found that simply running `fd` with no arguments in an empty directory on /tmp (which is in-memory!) was taking 0.2 seconds. This is ridiculously slow on my machine (Ryzen Threadripper 3990x with 64 cores/128 threads running on NVMe). Multiplied by the number of times I was calling `fd` in my suite, this immediately explained why multiple seconds had been added at some point.
> 
> When I add the `-j1` argument, as someone else indicated, the boot runtime of running `fd` alone drops to 0.01 seconds, which is 20x faster and far more reasonable an expectation.
> 
> I'm on NixOS unstable running on ZFS root.
> 
> @matu3ba & @tavianator , this is NOT a filesystem problem (IMHO). This looks like a heuristics bug regarding non-specification of the number of concurrent jobs to run (per the mandocs, not specifying it "uses heuristics")
> 
> At least I now know a workaround that works!

pmarreck:
(Just a note, I have a script at #1131 that anyone can run which shows the problem.)

sharkdp:
Can someone please provide some benchmark results? See https://github.com/sharkdp/fd/issues/1131#issuecomment-1338129673 as a starting point. Or https://github.com/sharkdp/fd-benchmarks for even more benchmarking scripts.

paulirish:
some benchmarks:

#### enormous folder (chromium repo w/ 443K files)

<img width="1178" alt="image" src="https://user-images.githubusercontent.com/39191/228330261-d8c758a3-a43e-4f32-80cd-e43e5e1b1ae7.png">


#### empty folder

<img width="1242" alt="image" src="https://user-images.githubusercontent.com/39191/228333248-2b2301b0-81f1-4bac-81dd-f51949de7c32.png">


fyi `nproc` says 48, though i had seen `96` in `lscpu`, thus the above command. :)  oops

json results for both: https://gist.github.com/paulirish/b94888c2e10f65702d8e09210c1584ad

PR #1412, opened 2023-10-26
Limit default threads

tmccombs:
Set maximum default threads

 Set a limit of how many threads fd will use by default. On hosts that
    have a large number of cores, using additional threads has diminishing
    returns, and having large numbers of threads increases the setup cost.
    Thus we don't necessarily want to use the same number of threads as we
    have cores.

 Fixes: #1203


tmccombs:
this builds on #1410

sharkdp (approved):
Thank you. Can you please add a ChangeLog with a reference to the bug ticket?

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@
 - Breaking: `.git/` is now ignored by default when using `--hidden` / `-H`, use `--no-ignore` / `-I` or
   `--no-ignore-vcs` to override, see #1387 and #1396 (@skoriop)
 
+
 ## Bugfixes
 
 ## Changes
 
+- The default number of threads is now constrained to be at most 16. This should improve startup time on
+  systems with many CPU cores. (#1203)
```
cuviper, line 14: Maybe I'm missing something, but isn't `MAX_DEFAULT_THREADS` actually 20?

I think it would also be good to note that maximum in the CLI help.

tmccombs: Ugh. I had a local change to set it to 16 instead of 20, but that didn't get included in the commit.

tavianator: We may want to tweak the limit again post #1422 anyway.  I get better perf all the way up to 48 threads on my 24 core CPU and the startup overhead is minimal now

tavianator: Actually I just checked more carefully and the ideal thread count for me is 24 on my 24 core/48 thread CPU.  I'm curious how this scales on different machines.

```
  ./fd-batch -u -j24 --search-path ~/code/bfs/bench/corpus/chromium ran
    1.02 ± 0.04 times faster than ./fd-batch -u -j28 --search-path ~/code/bfs/bench/corpus/chromium
    1.02 ± 0.03 times faster than ./fd-batch -u -j32 --search-path ~/code/bfs/bench/corpus/chromium
    1.07 ± 0.03 times faster than ./fd-batch -u -j20 --search-path ~/code/bfs/bench/corpus/chromium
    1.07 ± 0.03 times faster than ./fd-batch -u -j36 --search-path ~/code/bfs/bench/corpus/chromium
    1.08 ± 0.03 times faster than ./fd-batch -u -j40 --search-path ~/code/bfs/bench/corpus/chromium
    1.09 ± 0.02 times faster than ./fd-batch -u -j48 --search-path ~/code/bfs/bench/corpus/chromium
    1.09 ± 0.02 times faster than ./fd-batch -u -j44 --search-path ~/code/bfs/bench/corpus/chromium
    1.21 ± 0.02 times faster than ./fd-batch -u -j16 --search-path ~/code/bfs/bench/corpus/chromium
    1.52 ± 0.03 times faster than ./fd-batch -u -j12 --search-path ~/code/bfs/bench/corpus/chromium
    2.15 ± 0.04 times faster than ./fd-batch -u -j8 --search-path ~/code/bfs/bench/corpus/chromium
```

PR #1413, opened 2023-10-29
Refactor scan()

tavianator:


src/walk.rs
```diff
@@ -329,224 +212,374 @@ impl<W: Write> ReceiverBuffer<W> {
...
-            if cmd.in_batch_mode() {
-                exec::batch(rx, cmd, &config)
-            } else {
-                let out_perm = Mutex::new(());
+/// State shared by the sender and receiver threads.
+struct WorkerState {
+    /// The paths to scan.
+    paths: Vec<PathBuf>,
+    /// The search patterns.
+    patterns: Arc<Vec<Regex>>,
```
tmccombs, line 220: I wonder if it would be worth combining these Arcs into a single Arc, and pass that around, so we don't need to clone each individually.

src/walk.rs
```diff
@@ -329,224 +212,374 @@ impl<W: Write> ReceiverBuffer<W> {
...
-        // This will be set to `Some` if the `--exec` argument was supplied.
-        if let Some(ref cmd) = config.command {
-            if cmd.in_batch_mode() {
-                exec::batch(rx, cmd, &config)
-            } else {
-                let out_perm = Mutex::new(());
+/// State shared by the sender and receiver threads.
+struct WorkerState {
+    /// The paths to scan.
+    paths: Vec<PathBuf>,
```
tmccombs, line 218: AFAICT, this is only used to build the Walker. I think that if we changed `build_walker` and `build_overrides` to take the paths as a parameter, we could avoid having to allocate this Vec.

tmccombs (approved):

src/walk.rs
```diff
@@ -329,18 +212,123 @@ impl<W: Write> ReceiverBuffer<W> {
...
+        }
+
+        for path in &self.paths[1..] {
+            builder.add(path);
+        }
+
+        Ok(builder.threads(config.threads).build_parallel())
+    }
+
+    /// Run the receiver thread.
```
sharkdp, line 326: Maybe we can clarify a bit here: with this change, the receiving side now either (buffers and) prints results from the main thread. Or spawns new threads for job execution.

sharkdp (approved):
Thank you very much. Also a clear improvement on code structure!

src/walk.rs
tavianator: Even better: https://github.com/sharkdp/fd/pull/1413/commits/ed6396d9be4e41f07167899863045d083c3c5201 just removes the Arcs

src/walk.rs
tavianator: Done in https://github.com/sharkdp/fd/pull/1413/commits/ed6396d9be4e41f07167899863045d083c3c5201

sharkdp (approved):
Thank you for the update.

src/walk.rs
```diff
@@ -365,188 +347,230 @@ fn spawn_receiver(
...
+        let (tx, rx) = bounded(0x4000 * config.threads);
+
+        let exit_code = thread::scope(|scope| {
+            // Spawn the receiver thread(s)
+            let receiver = scope.spawn(|| self.receive(rx));
+
+            // Spawn the sender threads.
+            self.spawn_senders(walker, tx);
+
+            receiver.join().unwrap()
```
tmccombs, line 558: instead of spawning a separate thread and joining on it here, could we just call `self.receive(rx)` here?

tavianator: I tried that originally but the problem is `spawn_senders()` is blocking.  So we need an extra background thread either for `receive()` or `spawn_senders()`.

Issue #918, opened 2021-12-29
excessive memory usage on huge trees


cehteh:
When traversing some huge directory (5TB data/ 200Million Entries) i noticed that 'fd' will use excessive amounts of memory (I killed it at 35GB RSS). Some investigation shown that 'fd | wc -l' or 'fd >/dev/null' does not have this problem. While 'fd | less' again shows the same problem.

So my guess is that fd uses some unbounded queues to send the output which pile up because the terminal emulator is too slow to print 200 Million entries in time.

tmccombs:
Which version of fd are you using?

cehteh:
$ fd --version
fd 8.3.0

installed by cargo install

tmccombs:
Ah, I thought that we had set a limit on the channel size in #885, but it looks like that part got removed. So, yes, there is an unbounded queue of files and directories to process, and that could cause high memory usage if the terminal can't keep up. Maybe we should revisit adding a bound to that. I believe with the std mpsc crate there is  a performance cost to doing that, I'm not sure about crossbeam.

sharkdp:
> but it looks like that part got removed

Right. I removed it because you wrote *"I switch back to using channel instead of sync_channel than it is, at least not any slower"*. And I believe I was able to confirm this in my own benchmarks.

> Maybe we should revisit adding a bound to that. I believe with the std mpsc crate there is a performance cost to doing that, I'm not sure about crossbeam.

Yes, let's do this in a dedicated PR with new benchmark results (and memory profiles). A quick fix could be to use a much higher limit than the one suggested in #885.

cehteh:
hint: in my own directory traversal code i found that about 4k per thread (for memory calculation, not one channel per thread) is a somewhat sweet spot. Actually the limit isn't terribly important as long it doesn't explode, even with 100MB limit it should be okish.

mwaitzman:
Is this the cause of fd outputting "Killed" and then quitting when I search for all files with a specific extension from the root (Arch Linux, ~9/~14TB used space to search through across 5 mounted partitions)?

tavianator:
@mwaitzman It might be, do you see anything about the OOM killer in `dmesg`?

mwaitzman:
Yes, I see when running a similar command
```console
[...]

Issue #1362, opened 2023-08-04
[BUG] performance regression in trivial searches between 8.4 and 8.5 persisting in 8.7


moshelooks:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

It takes nearly 10x longer to search an empty directory with fd 8.7 or 8.5 versus 8.4
```
> mkdir empty
> hyperfine -N 'fd-v8.4.0-x86_64-unknown-linux-gnu/fd . empty/' 'fd-v8.5.0-x86_64-unknown-linux-gnu/fd . empty/' 'fd-v8.7.0-x86_64-unknown-linux-gnu/fd . empty/'
Benchmark 1: fd-v8.4.0-x86_64-unknown-linux-gnu/fd . empty/
  Time (mean ± σ):       4.0 ms ±   0.4 ms    [User: 2.2 ms, System: 2.6 ms]
  Range (min … max):     3.3 ms …   5.9 ms    594 runs

Benchmark 2: fd-v8.5.0-x86_64-unknown-linux-gnu/fd . empty/
  Time (mean ± σ):      39.2 ms ±   2.3 ms    [User: 3.9 ms, System: 36.2 ms]
  Range (min … max):    36.8 ms …  56.7 ms    79 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 3: fd-v8.7.0-x86_64-unknown-linux-gnu/fd . empty/
  Time (mean ± σ):      38.4 ms ±   2.9 ms    [User: 3.5 ms, System: 35.9 ms]
  Range (min … max):    35.5 ms …  55.3 ms    81 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  'fd-v8.4.0-x86_64-unknown-linux-gnu/fd . empty/' ran
    9.68 ± 1.12 times faster than 'fd-v8.7.0-x86_64-unknown-linux-gnu/fd . empty/'
    9.88 ± 1.05 times faster than 'fd-v8.5.0-x86_64-unknown-linux-gnu/fd . empty/'
```

FWIW this is not an issue with startup time per se; running `fd --version` does not show any performance drop. 

Using `-F ""` instead of `.` for the pattern show the same performance regression, if that's helpful to know.

[...]

PR #1414, opened 2023-10-30
walk: Use unbounded channels

### LINKED ISSUE #1408: Improve startup time
`fd`s startup time is quite slow. On my 12 core system, it takes ~ 20 ms for "searching" an empty folder. This is fast enough not to be noticeable by humans, but it looks bad in benchmarks when comparing `fd` with other tools on small folders [^1]. And it's also an actual problem for use cases where `fd` is called repeatedly from a script.

Some of that overhead is caused by the spawning of threads, and that problem is already tracked in https://github.com/sharkdp/fd/issues/1203. But I think there is more that can be done. Instead of using my usual go-to performance tool (`perf`), let's look at the [magic-trace](https://github.com/janestreet/magic-trace) output of a `fd` call in an empty folder [^2]. If someone is interested, I've attached the [trace](https://github.com/sharkdp/fd/files/13079377/trace.zip) to this post. Go to https://magic-trace.org/ to load it in their viewer.

The full trace looks like this:

image

The first 2.2 ms are typical process startup things (before `main`). I don't think there is any room for optimization here (?)

image

The next ~2 ms are more interesting:

image

some notable steps (even if insignificant in time)
- parsing command-line arguments (**724 µs**)
- a "is_existing_directory" check on the search path (28 µs)
- the parsing of the (empty) search pattern regex (32 µs)
- `isatty` check (5 ·s)
- `LsColors::from_env` (**579 µs**)
- num_cpus::linux::get_num_cpus` (**441 µs**)
- RegexBuilder::build (171 µs)

Some things were surprising to me. I didn't expect the `get_num_cpus` call to take this long. There might be some room for improvement here by doing things in parallel (e.g. `LsColors::from_env`)? But only if the thread overhead is not too high.

Then we start the actual scan, which takes the majority of the time:

image

Here, I'm not so sure how to interpret the trace, as things are actually happening on multiple threads. But we can (presumably) see some of the thread spawning/joining time here (~ **5 ms**):

image

and some gitignore matcher logic going on here (370 µs total):

image

Most of the time is actually unaccounted for in the trace, because I can only see:

image

We can see a bit more when switching off LTO:

image

Apparently, **11 ms** are spent in `crossbeam_channel::channel::bounded`'s `from_iter` method? (probably the receive call?) — even though we don't have any work to do. On a `-j1` run, this part only takes 1 ms.

[^1]: Those "small" folders can be pretty large, actually. It takes hundreds of thousands of files before we can make up for the startup "penalty".
[^2]: I recently discovered this and used it successfully to benchmark (and then optimize) the *startup* time of other programs.

tavianator:
Includes #1413.  The relevant commit is 5200718420577a6a2b4b474735567bad33318482:

> We originally switched to bounded channels for backpressure to fix https://github.com/sharkdp/fd/issues/918.
> However, bounded channels have a significant initialization overhead as
> they pre-allocate a fixed-size buffer for the messages.
> 
> This implementation uses a different backpressure strategy: each thread
> gets a limited-size pool of WorkerResults.  When the size limit is hit,
> the sender thread has to wait for the receiver thread to handle a result
> from that pool and recycle it.
> 
> Inspired by [snmalloc], results are recycled by sending the boxed result
> over a channel back to the thread that allocated it.  By allocating and
> freeing each WorkerResult from the same thread, allocator contention is
> reduced dramatically.  And since we now pass results by pointer instead
> of by value, message passing overhead is reduced as well.
> 
> Fixes https://github.com/sharkdp/fd/issues/1408.
> 
> [snmalloc]: https://github.com/microsoft/snmalloc

I benchmarked this with [`bfs`'s benchmark suite](https://github.com/tavianator/bfs/tree/main/bench).  Both `fd`s were built against a version of `ignore` that included https://github.com/BurntSushi/ripgrep/commit/d938e955af7a39ac0245b1ce84924f7f4a2141ac, so we won't actually see performance quite this good until a new `ignore` release happens.  Still, the results are good enough IMO that this fixes #1408 and #1362.

<details>
<summary>Benchmark results (updated)</summary>

### linux v6.5 (86,380 files)

|:---|---:|---:|---:|---:|
| `bfs bench/corpus/linux -false` | 20.2 ± 0.6 | 19.0 | 21.5 | 1.14 ± 0.11 |
| `find bench/corpus/linux -false` | 98.5 ± 0.2 | 98.1 | 98.9 | 5.58 ± 0.49 |
| `fd -u '^$' bench/corpus/linux` | 190.4 ± 47.0 | 133.0 | 231.3 | 10.78 ± 2.83 |
| `fd-master -u '^$' bench/corpus/linux` | 60.6 ± 15.7 | 29.5 | 70.5 | 3.43 ± 0.94 |
| `fd-unbounded -u '^$' bench/corpus/linux` | 17.7 ± 1.5 | 15.3 | 20.0 | 1.00 |

### rust 1.72.1 (192,714 files)

|:---|---:|---:|---:|---:|
| `bfs bench/corpus/rust -false` | 52.2 ± 1.7 | 50.2 | 56.4 | 1.55 ± 0.08 |
| `find bench/corpus/rust -false` | 313.5 ± 1.2 | 311.6 | 315.6 | 9.29 ± 0.40 |
| `fd -u '^$' bench/corpus/rust` | 274.6 ± 37.6 | 256.6 | 352.3 | 8.14 ± 1.17 |
| `fd-master -u '^$' bench/corpus/rust` | 55.7 ± 16.7 | 45.2 | 86.0 | 1.65 ± 0.50 |
| `fd-unbounded -u '^$' bench/corpus/rust` | 33.8 ± 1.4 | 31.6 | 36.3 | 1.00 |

### chromium 119.0.6036.2 (2,119,292 files)

|:---|---:|---:|---:|---:|
| `bfs bench/corpus/chromium -false` | 513.5 ± 11.4 | 490.6 | 528.1 | 2.06 ± 0.05 |
| `find bench/corpus/chromium -false` | 3285.0 ± 6.8 | 3275.2 | 3295.2 | 13.15 ± 0.14 |
| `fd -u '^$' bench/corpus/chromium` | 2538.8 ± 46.8 | 2476.3 | 2582.0 | 10.16 ± 0.22 |
| `fd-master -u '^$' bench/corpus/chromium` | 295.3 ± 17.6 | 264.9 | 307.4 | 1.18 ± 0.07 |
| `fd-unbounded -u '^$' bench/corpus/chromium` | 249.9 ± 2.6 | 246.5 | 254.8 | 1.00 |

#### linux v6.5

|:---|---:|---:|---:|---:|
| `bfs bench/corpus/linux` | 32.3 ± 1.6 | 27.5 | 34.8 | 1.03 ± 0.09 |
| `find bench/corpus/linux` | 103.0 ± 0.4 | 102.4 | 103.7 | 3.27 ± 0.24 |
| `fd -u --search-path bench/corpus/linux` | 192.0 ± 47.9 | 133.2 | 230.4 | 6.09 ± 1.58 |
| `fd-master -u --search-path bench/corpus/linux` | 87.1 ± 12.8 | 48.3 | 95.2 | 2.76 ± 0.45 |
| `fd-unbounded -u --search-path bench/corpus/linux` | 31.5 ± 2.3 | 29.5 | 40.2 | 1.00 |

#### linux v6.5

|:---|---:|---:|---:|---:|
| `bfs bench/corpus/linux -color` | 208.7 ± 2.6 | 204.3 | 214.0 | 2.71 ± 0.09 |
| `fd -u --search-path bench/corpus/linux --color=always` | 185.3 ± 49.1 | 133.2 | 230.3 | 2.40 ± 0.64 |
| `fd-master -u --search-path bench/corpus/linux --color=always` | 95.9 ± 21.1 | 67.4 | 121.2 | 1.24 ± 0.28 |
| `fd-unbounded -u --search-path bench/corpus/linux --color=always` | 77.1 ± 2.4 | 74.0 | 81.6 | 1.00 |

#### `-j1`

|:---|---:|---:|---:|---:|
| `bfs -j1 bench/corpus/rust -false` | 213.6 ± 0.5 | 212.8 | 214.4 | 1.00 |
| `fd -j1 -u '^$' bench/corpus/rust` | 277.2 ± 0.5 | 276.5 | 278.3 | 1.30 ± 0.00 |
| `fd-master -j1 -u '^$' bench/corpus/rust` | 283.4 ± 0.6 | 282.3 | 284.2 | 1.33 ± 0.00 |
| `fd-unbounded -j1 -u '^$' bench/corpus/rust` | 281.3 ± 0.5 | 280.6 | 282.1 | 1.32 ± 0.00 |

#### `-j2`

|:---|---:|---:|---:|---:|
| `bfs -j2 bench/corpus/rust -false` | 193.3 ± 1.0 | 191.5 | 195.2 | 1.24 ± 0.01 |
| `fd -j2 -u '^$' bench/corpus/rust` | 222.1 ± 5.3 | 216.5 | 231.8 | 1.42 ± 0.04 |
| `fd-master -j2 -u '^$' bench/corpus/rust` | 160.2 ± 1.5 | 158.3 | 162.7 | 1.03 ± 0.01 |
| `fd-unbounded -j2 -u '^$' bench/corpus/rust` | 155.9 ± 0.9 | 154.6 | 157.5 | 1.00 |

#### `-j3`

|:---|---:|---:|---:|---:|
| `bfs -j3 bench/corpus/rust -false` | 117.2 ± 6.2 | 108.7 | 125.6 | 1.05 ± 0.06 |
| `fd -j3 -u '^$' bench/corpus/rust` | 221.2 ± 2.5 | 217.1 | 223.7 | 1.99 ± 0.03 |
| `fd-master -j3 -u '^$' bench/corpus/rust` | 118.1 ± 2.4 | 112.9 | 121.0 | 1.06 ± 0.02 |
| `fd-unbounded -j3 -u '^$' bench/corpus/rust` | 111.2 ± 0.8 | 109.7 | 112.6 | 1.00 |

#### `-j4`

|:---|---:|---:|---:|---:|
| `bfs -j4 bench/corpus/rust -false` | 83.9 ± 4.1 | 77.1 | 89.6 | 1.00 |
| `fd -j4 -u '^$' bench/corpus/rust` | 231.4 ± 5.2 | 219.9 | 235.0 | 2.76 ± 0.15 |
| `fd-master -j4 -u '^$' bench/corpus/rust` | 95.4 ± 4.1 | 89.2 | 100.3 | 1.14 ± 0.07 |
| `fd-unbounded -j4 -u '^$' bench/corpus/rust` | 87.8 ± 1.1 | 85.7 | 89.5 | 1.05 ± 0.05 |

#### `-j6`

|:---|---:|---:|---:|---:|
| `bfs -j6 bench/corpus/rust -false` | 61.1 ± 1.4 | 58.1 | 63.8 | 1.00 |
| `fd -j6 -u '^$' bench/corpus/rust` | 230.6 ± 15.9 | 200.4 | 252.7 | 3.77 ± 0.27 |
| `fd-master -j6 -u '^$' bench/corpus/rust` | 74.0 ± 5.7 | 66.7 | 80.6 | 1.21 ± 0.10 |
| `fd-unbounded -j6 -u '^$' bench/corpus/rust` | 63.5 ± 0.9 | 61.8 | 64.8 | 1.04 ± 0.03 |

#### `-j8`

|:---|---:|---:|---:|---:|
| `bfs -j8 bench/corpus/rust -false` | 53.5 ± 2.2 | 50.1 | 57.4 | 1.04 ± 0.05 |
| `fd -j8 -u '^$' bench/corpus/rust` | 236.8 ± 13.2 | 222.3 | 259.0 | 4.61 ± 0.27 |
| `fd-master -j8 -u '^$' bench/corpus/rust` | 65.0 ± 7.5 | 57.0 | 73.2 | 1.27 ± 0.15 |
| `fd-unbounded -j8 -u '^$' bench/corpus/rust` | 51.4 ± 0.8 | 50.2 | 52.7 | 1.00 |

#### `-j12`

|:---|---:|---:|---:|---:|
| `bfs -j12 bench/corpus/rust -false` | 52.4 ± 2.4 | 47.9 | 57.0 | 1.27 ± 0.07 |
| `fd -j12 -u '^$' bench/corpus/rust` | 247.3 ± 13.7 | 230.0 | 268.8 | 5.99 ± 0.38 |
| `fd-master -j12 -u '^$' bench/corpus/rust` | 59.0 ± 12.0 | 46.9 | 73.3 | 1.43 ± 0.29 |
| `fd-unbounded -j12 -u '^$' bench/corpus/rust` | 41.3 ± 1.3 | 38.8 | 43.2 | 1.00 |

#### `-j16`

|:---|---:|---:|---:|---:|
| `bfs -j16 bench/corpus/rust -false` | 73.1 ± 5.5 | 65.9 | 83.5 | 2.06 ± 0.17 |
| `fd -j16 -u '^$' bench/corpus/rust` | 273.2 ± 10.3 | 246.7 | 280.3 | 7.69 ± 0.41 |
| `fd-master -j16 -u '^$' bench/corpus/rust` | 77.9 ± 0.9 | 76.3 | 80.0 | 2.19 ± 0.09 |
| `fd-unbounded -j16 -u '^$' bench/corpus/rust` | 35.5 ± 1.3 | 33.2 | 38.1 | 1.00 |

### Versions

```console
$ bfs --version | head -n1
bfs 3.0.4
$ find --version | head -n1
find (GNU findutils) 4.9.0
$ fd --version
fd 8.7.1
$ fd-master --version
fd 8.7.1
$ fd-unbounded --version
fd 8.7.1
```

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,6 +340,7 @@ dependencies = [
  "regex-syntax",
  "tempfile",
  "test-case",
+ "thread_local",
  "version_check",
 ]
 
@@ -797,6 +798,16 @@ dependencies = [
  "quote",
  "syn 2.0.38",
  "test-case-core",
+]
+
+[[package]]
+name = "thread_local"
+version = "1.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3fdd6f064ccff2d6567adcb3873ca630700f00b5ad3f060c25b5dcfd9a4ce152"
+dependencies = [
+ "cfg-if",
+ "once_cell",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ normpath = "1.1.1"
 crossbeam-channel = "0.5.8"
 clap_complete = {version = "4.4.4", optional = true}
 faccess = "0.2.4"
+thread_local = "1.1.7"
 
 [patch.crates-io]
 ignore = { git = "https://github.com/tavianator/ripgrep", branch = "fd" }
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -3,10 +3,9 @@ use std::sync::Mutex;
 use crossbeam_channel::Receiver;
 
 use crate::config::Config;
-use crate::dir_entry::DirEntry;
 use crate::error::print_error;
 use crate::exit_codes::{merge_exitcodes, ExitCode};
-use crate::walk::WorkerResult;
+use crate::walk::{WorkerMsg, WorkerResult};
 
 use super::CommandSet;
 
@@ -14,7 +13,7 @@ use super::CommandSet;
 /// generate a command with the supplied command template. The generated command will then
 /// be executed, and this process will continue until the receiver's sender has closed.
 pub fn job(
-    rx: Receiver<WorkerResult>,
+    rx: Receiver<WorkerMsg>,
     cmd: &CommandSet,
     out_perm: &Mutex<()>,
     config: &Config,
@@ -26,7 +25,8 @@ pub fn job(
     loop {
         // Obtain the next result from the receiver, else if the channel
         // has closed, exit from the loop
-        let dir_entry: DirEntry = match rx.recv() {
+        let result = rx.recv().map(WorkerMsg::take);
+        let dir_entry = match result {
             Ok(WorkerResult::Entry(dir_entry)) => dir_entry,
             Ok(WorkerResult::Error(err)) => {
                 if config.show_filesystem_errors {
@@ -49,18 +49,19 @@ pub fn job(
     merge_exitcodes(results)
 }
 
-pub fn batch(rx: Receiver<WorkerResult>, cmd: &CommandSet, config: &Config) -> ExitCode {
-    let paths = rx
-        .into_iter()
-        .filter_map(|worker_result| match worker_result {
-            WorkerResult::Entry(dir_entry) => Some(dir_entry.into_stripped_path(config)),
-            WorkerResult::Error(err) => {
-                if config.show_filesystem_errors {
-                    print_error(err.to_string());
+pub fn batch(rx: Receiver<WorkerMsg>, cmd: &CommandSet, config: &Config) -> ExitCode {
+    let paths =
+        rx.into_iter()
+            .map(WorkerMsg::take)
+            .filter_map(|worker_result| match worker_result {
+                WorkerResult::Entry(dir_entry) => Some(dir_entry.into_stripped_path(config)),
+                WorkerResult::Error(err) => {
+                    if config.show_filesystem_errors {
+                        print_error(err.to_string());
+                    }
+                    None
                 }
-                None
-            }
-        });
+            });
 
     cmd.execute_batch(paths, config.batch_size, config.path_separator.as_deref())
 }
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -13,7 +13,7 @@ use super::CommandSet;
 /// generate a command with the supplied command template. The generated command will then
 /// be executed, and this process will continue until the receiver's sender has closed.
 pub fn job(
-    rx: Receiver<WorkerMsg>,
+    rx: Receiver<WorkerMsg<'_>>,
     cmd: &CommandSet,
     out_perm: &Mutex<()>,
     config: &Config,
@@ -49,7 +49,7 @@ pub fn job(
     merge_exitcodes(results)
 }
 
-pub fn batch(rx: Receiver<WorkerMsg>, cmd: &CommandSet, config: &Config) -> ExitCode {
+pub fn batch(rx: Receiver<WorkerMsg<'_>>, cmd: &CommandSet, config: &Config) -> ExitCode {
     let paths =
         rx.into_iter()
             .map(WorkerMsg::take)
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -9,7 +9,7 @@ use std::thread;
 use std::time::{Duration, Instant};
 
 use anyhow::{anyhow, Result};
-use crossbeam_channel::{bounded, Receiver, RecvTimeoutError, Sender};
+use crossbeam_channel::{unbounded, Receiver, RecvTimeoutError, Sender};
 use etcetera::BaseStrategy;
 use ignore::overrides::{Override, OverrideBuilder};
 use ignore::{self, WalkBuilder, WalkParallel, WalkState};
@@ -43,6 +43,77 @@ pub enum WorkerResult {
     Error(ignore::Error),
 }
 
+/// Storage for a  WorkerResult.
+type ResultBox = Box<Option<WorkerResult>>;
+
+/// A WorkerResult that recycles itself.
+pub struct WorkerMsg {
+    inner: Option<ResultBox>,
+    tx: Sender<ResultBox>,
+}
+
+impl WorkerMsg {
+    /// Create a new message.
+    fn new(inner: ResultBox, tx: Sender<ResultBox>) -> Self {
+        Self {
+            inner: Some(inner),
+            tx,
+        }
+    }
+
+    /// Extract the result from this message.
+    pub fn take(mut self) -> WorkerResult {
+        self.inner.as_mut().unwrap().take().unwrap()
+    }
+}
+
+impl Drop for WorkerMsg {
+    fn drop(&mut self) {
+        let _ = self.tx.send(self.inner.take().unwrap());
+    }
+}
+
+/// A pool of WorkerResults that can be recycled.
+struct ResultPool {
+    size: usize,
+    tx: Sender<ResultBox>,
+    rx: Receiver<ResultBox>,
+}
+
+/// Capacity was chosen empircally to perform similarly to an unbounded channel
+const RESULT_POOL_CAPACITY: usize = 0x4000;
+
+impl ResultPool {
+    /// Create an empty pool.
+    fn new() -> Self {
+        let (tx, rx) = unbounded();
+
+        Self { size: 0, tx, rx }
+    }
+
+    /// Allocate or recycle a WorkerResult from the pool.
+    fn get(&mut self, result: WorkerResult) -> WorkerMsg {
+        let inner = if self.size < RESULT_POOL_CAPACITY {
+            match self.rx.try_recv() {
+                Ok(mut inner) => {
+                    *inner = Some(result);
+                    inner
+                }
+                Err(_) => {
+                    self.size += 1;
+                    Box::new(Some(result))
+                }
+            }
+        } else {
+            let mut inner = self.rx.recv().unwrap();
+            *inner = Some(result);
+            inner
+        };
+
+        WorkerMsg::new(inner, self.tx.clone())
+    }
+}
+
 /// Maximum size of the output buffer before flushing results to the console
 const MAX_BUFFER_LENGTH: usize = 1000;
 /// Default duration until output buffering switches to streaming.
@@ -56,8 +127,8 @@ struct ReceiverBuffer<'a, W> {
     quit_flag: &'a AtomicBool,
     /// The ^C notifier.
     interrupt_flag: &'a AtomicBool,
-    /// Receiver for worker results.
-    rx: Receiver<WorkerResult>,
+    /// Receiver for worker messages.
+    rx: Receiver<WorkerMsg>,
     /// Standard output.
     stdout: W,
     /// The current buffer mode.
@@ -72,7 +143,7 @@ struct ReceiverBuffer<'a, W> {
 
 impl<'a, W: Write> ReceiverBuffer<'a, W> {
     /// Create a new receiver buffer.
-    fn new(state: &'a WorkerState, rx: Receiver<WorkerResult>, stdout: W) -> Self {
+    fn new(state: &'a WorkerState, rx: Receiver<WorkerMsg>, stdout: W) -> Self {
         let config = &state.config;
         let quit_flag = state.quit_flag.as_ref();
         let interrupt_flag = state.interrupt_flag.as_ref();
@@ -104,7 +175,7 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
 
     /// Receive the next worker result.
     fn recv(&self) -> Result<WorkerResult, RecvTimeoutError> {
-        match self.mode {
+        let result = match self.mode {
             ReceiverMode::Buffering => {
                 // Wait at most until we should switch to streaming
                 self.rx.recv_deadline(self.deadline)
@@ -113,7 +184,8 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
                 // Wait however long it takes for a result
                 Ok(self.rx.recv()?)
             }
-        }
+        };
+        result.map(WorkerMsg::take)
     }
 
     /// Wait for a result or state change.
@@ -319,7 +391,7 @@ impl WorkerState {
 
     /// Run the receiver work, either on this thread or a pool of background
     /// threads (for --exec).
-    fn receive(&self, rx: Receiver<WorkerResult>) -> ExitCode {
+    fn receive(&self, rx: Receiver<WorkerMsg>) -> ExitCode {
         let config = &self.config;
 
         // This will be set to `Some` if the `--exec` argument was supplied.
@@ -355,12 +427,13 @@ impl WorkerState {
     }
 
     /// Spawn the sender threads.
-    fn spawn_senders(&self, walker: WalkParallel, tx: Sender<WorkerResult>) {
+    fn spawn_senders(&self, walker: WalkParallel, tx: Sender<WorkerMsg>) {
         walker.run(|| {
             let patterns = &self.patterns;
             let config = &self.config;
             let quit_flag = self.quit_flag.as_ref();
             let tx = tx.clone();
+            let mut pool = ResultPool::new();
 
             Box::new(move |entry| {
                 if quit_flag.load(Ordering::Relaxed) {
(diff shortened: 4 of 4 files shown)


tmccombs (approved):

tavianator:
> Wow 🤩

:)

> I would like to understand this first before merging. The situation is the following: we have a MPSC scenario where the consumer is sometimes too slow to handle the incoming results (even though its only job is to print the paths to a console). In the past, we used an unbounded channel which would lead to high memory usage because the channel was buffering all those results.

Exactly.  It was actually fairly easy to cause this by stalling the receiver thread with something like `fd | less`.

> Then we switched to a bounded channel, but that came at a high initialization cost because the bounded channels pre-allocate a fixed size buffer for the messages. The messages are `WorkerResult`s (`size_of::<WorkerResult>() == 312`). This fixed-size buffer presumably has a size of `channel_size × message_size`, i.e. `0x4000 × 312 byte ≈ 4.9 MiB` (?).

Yeah that times the number of threads: https://github.com/sharkdp/fd/blob/15329f9cfad43a1d9bef96914e9e30af5f68eae8/src/walk.rs#L60

> And that was slowing us down by ~70 ms on your machine? Is it because each thread allocates those 5 MiB?

I'm not sure exactly why `bounded()` channels are that slow to initialize, but the allocation happens up-front, not in each thread, which is part of the problem.

> In this changeset, you switch back to an unbounded channel, but `WorkerResult`s are re-cycled. I understand that this is getting rid of the pre-allocation hit. But what about long searches (when initialization time can be neglected)? You seem to indicate that we are still faster in those situations?

Yeah in my benchmarks it's universally faster.  But I don't know the separate impacts of

- Switching to an unbounded channel
- Shrinking `WorkerResult` to `WorkerMsg`
- Re-using `Box<WorkerResult>` allocations

> Couldn't this optimization be applied to bounded crossbeam channels in general (for large message sizes)? Is the allocation/deallocation really that expensive that it is worth all of this overhead (creating a sender for each worker result, making an additional copy for each worker result when recycling, …)?

Actually `Sender::clone()` is fairly cheap, it just [bumps an atomic refcount](https://github.com/crossbeam-rs/crossbeam/blob/17fecd9a442ab1582f75c81ce945bc4a66dc5c1e/crossbeam-channel/src/counter.rs#L47-L61).  I think it would even be possible to do

```rust
pub struct WorkerMsg<'a> {
    inner: Option<ResultBox>,
    tx: &'a Sender<ResultBox>,
}
```

but I'd have to plumb the lifetime through more stuff.  `WorkerState` would have to own the `Sender`s I think.

> Don't get me wrong. I love that this works. I'm just a bit puzzled that this isn't a strategy that could be used to speed up bounded channels in general (or maybe it is?).

Maybe it's possible?  Keep in mind this is somewhat different semantics because each thread is strictly limited to its pool of 0x4000 `WorkerResult`.  In the previous implementation, the capacity was shared between threads.

> Would another allocator help? I'm not really knowledgeable here but it seems to me like this whole recycling-memory part should/could be the job of a (special-purpose) allocator?

It's possible that https://github.com/microsoft/snmalloc would have some of the same benefits.  But here's the thing: this approach needs the receiver thread to tell the sender thread that it has handled a `WorkerResult` somehow.  My first thought was to use a semaphore to limit the number of `WorkerResult`s allocated.  But a single semaphore doesn't scale very well, so then I figured I could have a separate semaphore for each sender.

But actually, an SPSC channel scales better than a semaphore!  That's because the head/tail index can be on separate cache lines.  For a semaphore, both threads are always contending on the same counter.  So for that reason I just went with the channel implementation (and because we already had it available; I'd have to go find/write a semaphore otherwise).

tavianator:
> I think it would even be possible to do
> 
> ```rust
> pub struct WorkerMsg<'a> {
>     inner: Option<ResultBox>,
>     tx: &'a Sender<ResultBox>,
> }
> ```
> 
> but I'd have to plumb the lifetime through more stuff. `WorkerState` would have to own the `Sender`s I think.

Update: I just tried a hacky version of this and it wasn't appreciably faster

tmccombs:
> Re-using Box<WorkerResult> allocations

I suspect this is a pretty big component of it.  Unless the worker generates results faster than the receiver can process them, and we get to the maximum amount of allocation. 

It probably also helps that allocations for the channels now happens on the individual threads instead of in the spawning thread. 

Looking at the code for crossbeam, it isn't just that we have to allocate the memory for the channel up front, we also have to[ _initialize_ it](https://github.com/crossbeam-rs/crossbeam/blob/9dd4c9594dd7c84abf04e89192f1bdfe790100f5/crossbeam-channel/src/flavors/array.rs#L108), and it can't be done with `memset` or equivalent. 

> I'm just a bit puzzled that this isn't a strategy that could be used to speed up bounded channels in general (or maybe it is?)

I think it could be used in general. Basically implement a bounded channel as two unbounded channels and an atomic counter for the number of items allocated. And have the receiver pass the slot back after reading the value out of it. I'm not sure if that would universally improve performance, but it does have a couple of advantages: memory is allocated lazily, and it would be possible to dynamically change the size of the bound.

tavianator:
> Looking at the code for crossbeam, it isn't just that we have to allocate the memory for the channel up front, we also have to[ _initialize_ it](https://github.com/crossbeam-rs/crossbeam/blob/9dd4c9594dd7c84abf04e89192f1bdfe790100f5/crossbeam-channel/src/flavors/array.rs#L108), and it can't be done with `memset` or equivalent.

Well it *could* be done with `memset()` if they changed their representation slightly, e.g.

```rust
/// A slot in a channel.
struct Slot<T> {
    /// The current stamp.
    stamp: AtomicUsize,
    /// The message in this slot.
    msg: UnsafeCell<MaybeUninit<T>>,
}

impl<T> Slot<T> {
    fn new() -> Self {
        stamp: AtomicUsize::new(),
        msg: UnsafeCell::new(MaybeUninit::zeroed()),
    }
}
...
let slot = unsafe { self.buffer.get_unchecked(index) };
let stamp = slot.stamp.load(Ordering::Acquire) + index;
```

And if they did that they could allocate the whole buffer with `mmap()` and get lazily-initialized zero pages.  I use a similar trick in bfs to initialize a whole linked list from zeroed memory: https://github.com/tavianator/bfs/blob/b2ab7a151fca517f4879e76e626ec85ad3de97c7/src/alloc.c#L63-L88

sharkdp:
I did some benchmarks comparing current master (8bbbd7679b7eaf8782559c0320fc6cc86314319f) with this branch (d5889712454979f1f9ac4174b9f725135401b560).

I can confirm the huge increase in startup speed (`hyperfine -w5 -N -L version master,1414 "./fd-{version} -u . /tmp/empty" --export-markdown -`)

|:---|---:|---:|---:|---:|
| `./fd-master -u . /tmp/empty` | 25.4 ± 10.0 | 18.1 | 54.7 | 7.24 ± 3.34 |
| `./fd-1414 -u . /tmp/empty` | 3.5 ± 0.9 | 2.6 | 9.5 | 1.00 |

Unfortunately, there seems to be a large regression for longer searches with many results (`hyperfine -w5 -N -L version master,1414 "./fd-{version} -u . /folder/with/3M/files" --export-markdown -`):

|:---|---:|---:|---:|---:|
| `./fd-master -u . /home/ped1st/workspace` | 841.0 ± 125.2 | 737.2 | 1069.5 | 1.00 |
| `./fd-1414 -u . /home/ped1st/workspace` | 1386.5 ± 18.0 | 1358.8 | 1411.5 | 1.65 ± 0.25 |

tavianator:
I just pushed a couple more commits with different strategies, including passing senders by reference and my own little semaphore implementation.  Unfortunately, it seems like most of the overhead is actually from the unbounded channel implementation itself, i.e. initialization time is faster but sending is slower.  Let me try a different approach.

tmccombs:
I wonder if it would be worthwhile to try to reduce the initialization overhead in crossbeam channel.

tmccombs:
A couple of other ideas:

We could reduce the size of the channel, which would speed up initialization, but might hurt performance if the bound is the bottleneck.

Instead of having a single channel, we could have each sender thread create its own channel, that it passes back to the main thread via another channel. 

Then in the receiver we either select from all the channels, or if we spawn multiple receiver threads, each receiver could process a single input channel.

tavianator:
> A couple of other ideas:
> 
> We could reduce the size of the channel, which would speed up initialization, but might hurt performance if the bound is the bottleneck.

I've tried this before, usually it's a perf loss

> Instead of having a single channel, we could have each sender thread create its own channel, that it passes back to the main thread via another channel.
> 
> Then in the receiver we either select from all the channels, or if we spawn multiple receiver threads, each receiver could process a single input channel.

This was the "different approach" I mentioned above.  It's much slower.

But I just tried another idea: instead of individual `WorkerResult`s, send something like `Arc<Mutex<Option<Vec<WorkerResult>>>>` over the channel.  The senders keep adding to the same batch whenever they can, and the receiver thread drains whole batches at once.  It's **tremendously** faster than everything else I've tried:

```
Benchmark 1: bfs bench/corpus/chromium                                                                                                                                                       
  Time (mean ± σ):     679.5 ms ±  23.1 ms    [User: 967.2 ms, System: 3165.0 ms]
  Range (min … max):   648.4 ms … 710.3 ms    10 runs                                  
                                                                                              
Benchmark 2: find bench/corpus/chromium                                                                                                                                                      
  Time (mean ± σ):      3.213 s ±  0.020 s    [User: 0.559 s, System: 2.602 s]                                                                                                               
  Range (min … max):    3.181 s …  3.245 s    10 runs                                                                                                                                        
                                                                                              
Benchmark 3: fd -u --search-path bench/corpus/chromium
  Time (mean ± σ):      2.637 s ±  0.011 s    [User: 10.695 s, System: 99.795 s]              
  Range (min … max):    2.621 s …  2.654 s    10 runs                    
                                                                                              
Benchmark 4: fd-master -u --search-path bench/corpus/chromium     
  Time (mean ± σ):     767.8 ms ±  36.4 ms    [User: 10885.6 ms, System: 3892.2 ms]
  Range (min … max):   725.8 ms … 817.7 ms    10 runs                       
                                                                                              
Benchmark 5: fd-batch -u --search-path bench/corpus/chromium                                  
  Time (mean ± σ):     293.2 ms ±   6.7 ms    [User: 2767.8 ms, System: 3095.5 ms]            
  Range (min … max):   282.3 ms … 302.3 ms    10 runs                      
                                                                                              
Summary                                                                                       
  fd-batch -u --search-path bench/corpus/chromium ran                                         
    2.32 ± 0.09 times faster than bfs bench/corpus/chromium                                   
    2.62 ± 0.14 times faster than fd-master -u --search-path bench/corpus/chromium            
    8.99 ± 0.21 times faster than fd -u --search-path bench/corpus/chromium               
   10.96 ± 0.26 times faster than find bench/corpus/chromium                                                                                                                                 ```

I'll put up a PR for it in a bit.

PR #1421, opened 2023-11-04
fix: display color when `NO_COLOR` is an empty string

acuteenvy:
Same as https://github.com/sharkdp/bat/pull/2767.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ## Bugfixes
 
+- Fix `NO_COLOR` support, see #1421 (@acuteenvy)
+
 ## Changes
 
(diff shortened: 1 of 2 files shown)


tavianator (approved):

tavianator:
Hmm I just realized after merging this that we should add a test.  Do you feel like doing that @acuteenvy?

acuteenvy:
I can't seem to get `fd` to output color in tests at all, unless I pass `--color=always` which makes the test pointless.

tmccombs:
It would be possible to write a test, at least on linux. But it would working with a pty, which would be fairly involved.

tavianator:
Right yeah, I don't think it's worth the effort

Issue #400, opened 2019-01-27
Proposal: add `--filter <cmd>` option


sharkdp:
This is a proposal to add a `--filter <cmd>`/`-f <cmd>` command-line option that would work very similar to the `filter` option in (functional) programming languages. It would be implemented in a similar way like `--exec <cmd>` and execute the given command for every path. Instead of showing the output of the command, it would filter the search results based on the exit status of the command (i.e. only include the ones which succeed with exit code 0).

In addition, we could also add a `--reject <cmd>`/`-r <cmd>` option that would do the opposite.

This would allow us to do things like:

``` bash
# Show all C++ source files that contain the word TODO
fd -e cpp -e h -f grep TODO

# Show all entries that are NOT directories (this can not be done with `--type`)
fd -r test -d

# Only show files which exist and the current user has write permission
fd … test -w

# Only show files that are newer than 'reference.txt'
fd … -f test {} -nt reference.txt

# Only show files which have a diff with respect to the Git index
fd … -f git diff --exit-code

# Show all *.jpg files with a mime type that does NOT match image/jpeg
fd -e jpg -f sh -c '[ $(file -b --mime-type "{}") != "image/jpeg" ]'
```

I'd be happy to get some feedback on this idea or hear about other ways on how this could be used.

tavianator:
Your last suggestion is vulnerable to shell injection, or just strange behaviour if you have filenames with strange characters.  You could do
```sh
fd -e jpg -f sh -c '[ $(file -b --mime-type "$1") != "image/jpeg" ]' sh {}
```
instead to fix it.

tavianator:
@sharkdp Yeah it's a little complex.  Maybe some syntax that does it for you would be nice, like
```
fd --filter-shell '[ $(file -b --mime-type "$1") != "image/jpeg" ]'
```
and you could do the `sh -c '<command>' sh <file>` dance for the user.

But having all of `--exec`, `--exec-shell`, `--exec-batch`, `--exec-batch-shell`, `--filter`, `--filter-shell`, etc. might be a bit much.

crides:
I found this today when I was checking for mime type support in `fd`. Using shell filtering is definitely an interesting idea, but I would rather use the original `find` for that, as it would probably require more complex logics when you need to use other programs to check the file, which kind of contradicts `fd`'s philosophy of "simplicity". 

Another way to do it is just to use `-x --exec` itself as a predicate, like what `find` does.

sharkdp:
[...]

PR #1426, opened 2023-11-12
Draft: Filter by EXIF GPS location

simon04:
I wanted to find a photo I've taken at a certain location some years ago.

This PR adds a filter to extract the GPS location and filter files within a certain distance to that location.

This feature might be out of scope, but I (a) wanted to use this functionality, and (b) do some Rust programming. Let me know what you think!

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -21,6 +21,7 @@ use crate::error::print_error;
 use crate::exec;
 use crate::exit_codes::{merge_exitcodes, ExitCode};
 use crate::filesystem;
+use crate::filter::{exif_geo_distance, GeoLocation};
 use crate::output;
 
 /// The receiver thread can either be buffering results or directly streaming to the console.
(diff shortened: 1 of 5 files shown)


src/walk.rs
```diff
@@ -502,6 +503,19 @@ impl WorkerState {
                     }
                 }
 
+                let reference = GeoLocation {
+                    latitude: 47.3,
+                    longitude: 11.3,
+                };
+                let max_distance = 500.;
```
simon04, line 510: The reference location and `max_distance` need to be parsed from the command line ...

tmccombs:
I think this is definitely out of scope. And I think a more general feature like https://github.com/sharkdp/fd/issues/400 would probably be a better fit for fd.

sharkdp:
Thank you very much for your work, but I agree with @tmccombs.

PR #1431, opened 2023-11-29
cli: Tweak default thread count logic

tavianator:
See the discussion starting here: https://github.com/sharkdp/fd/pull/1422#issuecomment-1832189765

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -715,24 +715,14 @@ impl Opts {
 fn default_num_threads() -> NonZeroUsize {
     // If we can't get the amount of parallelism for some reason, then
     // default to a single thread, because that is safe.
-    // Note that the minimum value for a NonZeroUsize is 1.
-    // Unfortunately, we can't do `NonZeroUsize::new(1).unwrap()`
-    // in a const context.
-    const FALLBACK_PARALLELISM: NonZeroUsize = NonZeroUsize::MIN;
-    // As the number of threads increases, the startup time suffers from
-    // initializing the threads, and we get diminishing returns from additional
-    // parallelism. So set a maximum number of threads to use by default.
-    //
-    // This value is based on some empirical observations, but the ideal value
-    // probably depends on the exact hardware in use.
-    //
-    // Safety: The literal "20" is known not to be zero.
-    const MAX_DEFAULT_THREADS: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(20) };
+    let fallback = NonZeroUsize::MIN;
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -715,24 +715,25 @@ impl Opts {
...
+
+    // Get the total number of CPUs available, including hyperthreads
+    let threads = std::thread::available_parallelism().unwrap_or(fallback);
+    if threads <= threshold {
+        return threads;
+    }
+
+    // Compute min(threads, max(cores, threshold), limit). Avoid calling
+    // num_cpus::get_physical() if we don't have to, since it costs a bit of
+    // startup time to parse /proc/cpuinfo.
```
sharkdp, line 734: For my 6 core / 12 threads laptop, this would select

    min(12, max(6, 8), 64) = min(12, 8, 64) = 8

threads by default, instead of 12 previously.


This leads to a small performance regression for warm cache benchmarks:

`hyperfine --warmup 50 -L j 8,12 "fd -j{j} -u --search-path …"`

|:---|---:|---:|---:|---:|
| `fd -j8 -u --search-path …` | 347.8 ± 7.3 | 341.8 | 367.2 | 1.06 ± 0.03 |
| `fd -j12 -u --search-path …` | 327.9 ± 5.1 | 322.9 | 341.5 | 1.00 |

And to a quite large performance regression for cold cache benchmarks:

`hyperfine --prepare "sync; echo 3 | sudo tee /proc/sys/vm/drop_caches" -L j 8,12 "fd -j{j} -u --search-path …"`

|:---|---:|---:|---:|---:|
| `fd -j8 -u --search-path /home/ped1st/workspace` | 2.791 ± 0.017 | 2.780 | 2.837 | 1.36 ± 0.01 |
| `fd -j12 -u --search-path /home/ped1st/workspace` | 2.059 ± 0.018 | 2.038 | 2.082 | 1.00 |


This is a general observation that I made consistently in the past as well: cold cache traversals have a higher optimal thread count than warm cache benchmarks. Since there is no way to know, we should probably try to optimize for both scenarios? Or maybe favor one of the two, but still keep the other in mind.

Here is a parameter scan showing both warm cache (red) and cold cache (blue) results:

image

tavianator: Ah, I really need to add a cold cache benchmark to my usual set.  I'll do some tests on my machines, but maybe the best strategy is just to raise the current limit to ~64 and not bother with distinguishing cores vs. hyperthreads

tavianator: Done!

sharkdp (approved):
Thank you

Issue #1331, opened 2023-06-02
force cwd-prefix to be on


danielb2:
Ref: https://github.com/sharkdp/fd/issues/1243, https://github.com/sharkdp/fd/issues/1182, https://github.com/sharkdp/fd/issues/1046

I'm using fd with fzf and I want to be able to search from a relative path starting with ./

I've read several issues in regards to `./` and I think they would all be solved if we had two separate options

```
--strip-cwd-prefix
--show-cwd-prefix
```

instead of predicting defaults for different use-cases, this would allow anyone to override the behavior in either situation

Issue #1243, opened 2023-01-15
[BUG] --strip-cwd-prefix docs out of date


PatrickF1:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

man fd's docs for --strip-cwd-prefix says
```
By default, relative paths are prefixed with './' when the output goes to a non interactive terminal (TTY). Use this flag to disable this behaviour.
```
Actually, 8.5.0 removed that behavior: `No leading ./ prefix for non-interactive results`.

### Describe what you expected to happen:

The docs for that option should say

```
Relative paths are prefixed with './' when --exec/-x, --exec-batch/-X, or --print0/-0 are used. Use this flag to disable this behaviour.
```

### What version of `fd` are you using?

8.6.0

### Which operating system / distribution are you on?

```shell
Darwin 22.2.0 arm64
```

sharkdp:
Thank you. This should be fixed.

PR #1445, opened 2023-12-06
feat: Add option to always include cwd prefix

tmccombs:
Fixes: #1243
Fixes: #1331

- [x] add test
- [x] update changelog

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -163,12 +163,12 @@ path starting with '-' being treated as a command line option. Use
 this flag to change this behavior. If this flag is used without a value,
 it is equivalent to passing "always". Possible values are:
 .RS
-.IP auto
-Use the default behavior.
 .IP never
 Never strip the ./ at the beginning of paths
 .IP always
 Always strip the ./ at the beginning of paths
+.IP auto
+Only strip if used with --exec, --exec-batch, or --print0. That is, it resets to the default behavior.
 .RE
 .TP
 .B \-\-one\-file\-system, \-\-mount, \-\-xdev
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -156,9 +156,20 @@ can be used as an alias.
 Enable the display of filesystem errors for situations such as insufficient
 permissions or dead symlinks.
 .TP
-.B \-\-strip-cwd-prefix
-By default, relative paths are prefixed with './' when the output goes to a non interactive terminal
-(TTY). Use this flag to disable this behaviour.
+.B \-\-strip-cwd-prefix [when]
+By default, relative paths are prefixed with './' when -x/--exec,
+-X/--exec-batch, or -0/--print0 are given, to reduce the risk of a
+path starting with '-' being treated as a command line option. Use
+this flag to change this behavior. If this flag is used without a value,
+it is equivalent to passing "always". Possible values are:
+.RS
+.IP auto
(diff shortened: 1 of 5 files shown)


src/cli.rs
```diff
@@ -622,6 +622,18 @@ pub struct Opts {
...
+    /// is not included unless using --exec, --exec-batch, or --print0. But with
+    /// this option, the prefix is never stripped.
+    #[arg(
+        long,
+        hide_short_help = true,
+        conflicts_with("absolute_path"),
+        overrides_with("strip_cwd_prefix"),
+        long_help
+    )]
+    pub cwd_prefix: bool,
```
tmccombs, line 635: Alternatively, I could change the `--strip-cwd-prefix` option to take an enum value that is one of "always", "never", and "auto" (current default behavior). Although in that case I would want to treat it being given without a value as `always`. I'm not sure if clap can handle that.

tavianator:
Not sure if this was mentioned in the linked issues, but it is already possible to get the leading `./` by specifying `.` as the search path explicitly, e.g.

```console
$ fd . .
./cli.rs
./config.rs
...
$ fd --search-path .
./cli.rs
./config.rs
...
```

doc/fd.1
```diff
@@ -159,9 +159,20 @@ can be used as an alias.
...
+path starting with '-' being treated as a command line option. Use
+this flag to change this behavior. If htis flag is used without a value,
+it is equivalent to passing "always". Possible values are:
+.RS
+.IP auto
+Use the default behavior.
+.IP never
+Never strip the ./ at the beginning of paths
+.IP always
+Always strip the ./ at the beinning of paths
```
sharkdp, line 174:
```suggestion
Always strip the ./ at the beginning of paths
```

doc/fd.1
```diff
@@ -159,9 +159,20 @@ can be used as an alias.
...
 .TP
-.B \-\-strip-cwd-prefix
-By default, relative paths are prefixed with './' when the output goes to a non interactive terminal
-(TTY). Use this flag to disable this behaviour.
+.B \-\-strip-cwd-prefix [when]
+By default, relative paths are prefixed with './' when -x/--exec,
+-X/--exec-batch, or -0/--print0 are given, to reduce the risk of a
+path starting with '-' being treated as a command line option. Use
+this flag to change this behavior. If htis flag is used without a value,
+it is equivalent to passing "always". Possible values are:
```
sharkdp, line 167: I would have expected "auto" here, since it is described as the default behavior below?

tmccombs: Making it equivalent to always makes it simpler to turn it on.  And I think that having `--strip-cwd-prefix` without a specific when value meaning "auto" instead of "always" could be confusing since it reals like it is always stripping the prefix.

tmccombs: Also, using `always` if  this is used without an argument is more backwards compatible with existing usage of the argument.

sharkdp: I think the thing that confused me was that "auto" is described as "Use the default behavior" below. And I misinterpreted that as "auto = do whatever you do when no argument is given". After re-reading everything, it still seems a bit strange to me, but I agree that `always` makes sense as a default.

sharkdp (approved):

sharkdp:
Thank you for the update

PR #1452, opened 2023-12-13
walk: Flush stdout in batches

### LINKED ISSUE #1313: Add a streaming mode when running under pipe
Currently when not running under an interactive shell, it will buffer its output and output everything in one go.
However, I am writing a GUI front end and need this "streaming" mode similar to when running under an interactive shell so users can see the result in a faster manner.

Expected behaviour is that the user can instantly see the initial result with more items being added along the way.

Am I missing something or is there any switch to customize this behavior?

tavianator:
The previous behaviour was designed to mimic the output buffering of
typical UNIX tools: line-buffered if stdout is a TTY, and fully-buffered
otherwise.  More precicely, when printing to a terminal, fd would flush
explicitly after printing any buffered results, then flush after every
single result once streaming mode started.  When not printing to a
terminal, fd never explicitly flushed, so writes would only happen as
the BufWriter filled up.

The new behaviour actually unifies the TTY and non-TTY cases: we flush
after printing the buffered results, then once we start streaming, we
flush after each batch, but *only when the channel is empty*.  This
provides a good balance: if the channel is empty, the receiver thread
might as well flush before it goes to sleep waiting for more results.
If the channel is non-empty, we might as well process those results
before deciding to flush.

For TTYs, this should improve performance by consolidating write() calls
without sacrificing interactivity.  For non-TTYs, we'll be flushing more
often, but only when the receiver would otherwise have nothing to do,
thus improving interactivity without sacrificing performance.  This is
particularly handy when fd is piped into another command (such as head
or grep): with the old behaviour, fd could wait for the whole traversal
to finish before printing anything.  With the new behaviour, fd will
print those results soon after they are received.

Fixes #1313.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -214,7 +214,6 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
                                 }
                                 ReceiverMode::Streaming => {
                                     self.print(&dir_entry)?;
-                                    self.flush()?;
                                 }
                             }
 
@@ -231,6 +230,11 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
                             }
                         }
                     }
+                }
+
+                // If we don't have another batch ready, flush before waiting
+                if self.mode == ReceiverMode::Streaming && self.rx.is_empty() {
+                    self.flush()?;
                 }
             }
             Err(RecvTimeoutError::Timeout) => {
@@ -285,7 +289,7 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
 
     /// Flush stdout if necessary.
     fn flush(&mut self) -> Result<(), ExitCode> {
-        if self.config.interactive_terminal && self.stdout.flush().is_err() {
+        if self.stdout.flush().is_err() {
             // Probably a broken pipe. Exit gracefully.
             return Err(ExitCode::GeneralError);
         }


tavianator:
Benchmarks:

First of all, here's the purpose of the PR:

```console
tavianator@tachyon $ fd-master title_t00.mkv ~ | time head -n1
.../title_t00.mkv
head -n1  0.00s user 0.00s system 0% cpu 1.435 total
tavianator@tachyon $ fd-flush title_t00.mkv ~ | time head -n1
.../title_t00.mkv
head -n1  0.00s user 0.00s system 1% cpu 0.104 total
```

### `--output=null`

|:---|---:|---:|---:|---:|
| `fd-master -u --search-path ~/code/bfs/bench/corpus/chromium` | 296.7 ± 5.2 | 288.2 | 305.5 | 1.00 |
| `fd-flush -u --search-path ~/code/bfs/bench/corpus/chromium` | 296.8 ± 6.8 | 287.1 | 307.9 | 1.00 ± 0.03 |

### `--output=pipe`

|:---|---:|---:|---:|---:|
| `fd-master -u --search-path ~/code/bfs/bench/corpus/chromium` | 353.3 ± 19.3 | 320.7 | 381.1 | 1.01 ± 0.07 |
| `fd-flush -u --search-path ~/code/bfs/bench/corpus/chromium` | 351.1 ± 14.4 | 329.2 | 368.6 | 1.00 |

### `--output=inherit` (aka a TTY)

I sent the output to an inactive tmux pane so it wouldn't depend on my actual terminal emulator or my ssh connection.

|:---|---:|---:|---:|---:|
| `fd-master -u --search-path ~/code/bfs/bench/corpus/rust` | 609.1 ± 12.6 | 592.5 | 635.8 | 1.04 ± 0.03 |
| `fd-flush -u --search-path ~/code/bfs/bench/corpus/rust` | 588.2 ± 14.9 | 559.3 | 604.3 | 1.00 |

sharkdp:
Thank you very much!

Issue #1455, opened 2023-12-19
[BUG] Colorized output in `-l` despite `NO_COLOR=1` and `LS_COLORS=`


nagy:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

Despite specifying `NO_COLOR=1` and `LS_COLORS=` , I still see colorized output when using the `-l` flag.

```
$ LS_COLORS= NO_COLOR=1 fd null /dev/ -l
crw-rw-rw- 1 root root 1, 3 Dec 11 15:40 /dev/null
```

In this example, the final `/dev/null` part is colorized. With `--color never` it works though.

### Describe what you expected to happen:

Monochrome output

### What version of `fd` are you using?

fd 9.0.0

### Which operating system / distribution are you on?

```shell
Linux
```

sharkdp:
Thank you for reporting this.

This needs to be fixed here:

https://github.com/sharkdp/fd/blob/266311ca33a78924dcad1fc13f99974098286ed0/src/main.rs#L328

PR #1458, opened 2023-12-23
fix: Respect NO_COLOR with --list-details option

tmccombs:
Fixes: #1455


sharkdp (approved):
Thank you very much.

I manually verified that it also works in a "auto" setting, i.e. piped `fd -l` output is still uncolored.

PR #1480, opened 2024-01-18
Add a flag to anchor patterns in the input

### LINKED ISSUE #1476: More ways to anchor patterns
I would like to use fixed-string patterns to match against a whole basename (something like find -name while disabling fnmatch or quoting metacharacters).

Right now, glob patterns are anchored to match the entire input (as determined by --full-path), and regex patterns aren't, and by adding * or ^$ it's easy to tweak them to get the desired anchoring.  But with fixed-string patterns, there's no easy way to turn them into regex or blobs from a shell pipeline (one needs to call regex::escape with a matching crate version) to have patterns match start/end/whole input.

g2p:
Closes #1476.


tavianator:
I'll let @sharkdp decide if he wants this additional command-line flag.  I will point out that `--anchor={input,-{start,end}}` could be made compatible with `--glob` by just adding `*` at the beginning or end as appropriate.

sharkdp:
> I'll let @sharkdp decide if he wants this additional command-line flag.

I'd be okay with adding this new command-line option.

> I will point out that `--anchor={input,-{start,end}}` could be made compatible with `--glob` by just adding `*` at the beginning or end as appropriate.

That would be great.

PR #1483, opened 2024-01-20
Makefile: use `--locked` for build

ognevny:
to be ensured that Cargo.lock is up to date. also it's needed for disributions


tmccombs (approved):

tmccombs:
Thank you!

PR #1492, opened 2024-02-11
Add installation instructions for more Linux distributions in README

AlbydST:
Added installation instructions for Arch AUR, ALT Linux and Solus, added a disclaimer to replace sudo with doas in commands depending on what you use and made a few minor changes.


tavianator:
I don't think adding `sudo` to the front of most commands is worth it.  Users typically know whether their package manager needs to run as root.  There is a convention to use the prompt (`$`/`#`) to indicate regular user/root, e.g.

```console
# apt install fd
```

README.md
```diff
@@ -668,6 +684,8 @@ Note that rust version *1.70.0* or later is required.
 
 `make` is also needed for the build.
 
+Make sure to replace **sudo** with **doas** in commands depending on what you use.
```
tmccombs, line 687: This is also unnecessary. If you use  linux and wen to the effort to install and set up doas, and use that instead of sudo, then you probably already know that you should change the commands to use doas instead of sudo.

tmccombs (approved):

PR #1505, opened 2024-03-04
set default path separator to '/' in UCRT64 of MSYS

leuven65:
Use path separator "/" also in UCRT64 of MSYS.

The environments of MSYS: [[https://www.msys2.org/docs/environments/]]


leuven65:
I have adapt the code to check if $MSYSTEM is non-empty.

Issue #1511, opened 2024-03-09
Add clippy check to github actions CI


garlic-hub:
Hello. I recently submitted a PR to fix all active clippy warnings (which was merged, thank you). With that done, I'm wondering if there's interest to automatically flag a warning on any PR that introduces a clippy warning?

garlic-hub:
I'm thinking something like this. Not too familiar with github actions
```yml
  check_cargo_clippy:
    name: Ensure 'cargo clippy' warnings have been addressed
    runs-on: ubuntu-20.04
    steps:
    - uses: dtolnay/rust-toolchain@stable
    - uses: actions/checkout@v4
    - run: cargo clippy --all-targets --all-features
```

tmccombs:
we do actually run clippy in our github actions. But right now only on the MSRV version.

PR #1520, opened 2024-03-14
Add clippy using latest rust version to CI

tmccombs:
Fixes: #1511

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -44,6 +44,16 @@ jobs:
     - uses: actions/checkout@v4
     - run: cargo fmt -- --check
 
+  lint_check:
+    name: Ensure 'cargo clippy' has no warnings
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        components: clippy
+    - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


sharkdp (approved):
Thank you

RossSmyth:
One thing to note is that this only checks lints for the native toolchain target triple. For example, this would not check the lints of things that are `#[cfg(windows)]` or `#[cfg(not(unix))]`. Currently if you build fd from source on Windows this warning is emitted:

```powershell
warning: field `interactive_terminal` is never read
  --> src\config.rs:77:9
   |
13 | pub struct Config {
   |            ------ field in this struct
...
77 |     pub interactive_terminal: bool,
   |         ^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```
Due to this *nix cfg
https://github.com/sharkdp/fd/blob/8c7a84ea30fefd1258e37d5f20a8c75fd77d48f9/src/output.rs#L133-L160

The `--all-targets` flag does not mean all target triples, but rather all compilation targets such as examples, tests, binaries, and libraries that could be compiled. To actually lint all target triples you would do

```yaml
- run: cargo clippy --all-targets --all-features -- -Dwarnings --target x86_64-unknown-linux-gnu x86_64-pc-windows-gnu aarch64-apple-darwin
```
Or whatever targets are desired to check.

Issue #1461, opened 2023-12-27
[BUG] Including hidden files also includes (git)ignored files if the .gitignore is in the parent directory


taeruh:
### Checks

- [X] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

Assume we have the following directory structure:
```
.
├── .abc
├── .gitignore
└── sub_directory
   └── .abc
```
where
```
cat .gitignore
.abc
```
Using `fd` in the `sup_directory` with `--hidden`, I get
```
fd --hidden
.abc
```
The `.abc` file is shown, although it should have been ignored.

### Describe what you expected to happen:

Do not show the `.abc` file.

Note that this only happens in the `sup_directory`. Using `fd --hidden` in the root directory shows the expected output.

Possible Fix:
```
diff --git a/src/cli.rs b/src/cli.rs
index 0007a31..3a7081f 100644
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -643,7 +643,7 @@ impl Opts {
         } else if !self.search_path.is_empty() {
             &self.search_path
         } else {
-            let current_directory = Path::new(".");
+            let current_directory = Path::new("./");
             ensure_current_directory_exists(current_directory)?;
             return Ok(vec![self.normalize_path(current_directory)]);
         };
```
All tests pass in the github workflows with this change.

The bug is not really caused by something `fd` is doing but rather by what the "Walker" in the `ignore` dependency is doing. A simple loop over `WalkBuilder::new(path).hidden(false).build()` with `path = "."` and `path = "./"` leads to the same wrong and correct results, respectively. Maybe one could report it as a bug there? I'm rather confused by this behavior though.

Anyways, the diff above should fix it in `fd`, I think (could open a PR if you want).

### What version of `fd` are you using?

fd 9.0.0

### Which operating system / distribution are you on?

```shell
Linux 6.1.69-1-lts x86_64
```

tavianator:
This definitely seems like a bug in the `ignore` crate.  You should have a `.git` folder in your reproducer, but even with that it fails:

```console
tavianator@graphene $ tree -a .
.
├── .abc
├── .git
├── .gitignore
└── sub_directory
    └── .abc

2 directories, 4 files
tavianator@graphene $ cat .gitignore
.abc
tavianator@graphene $ cd sub_directory
tavianator@graphene $ fd --hidden
.abc
tavianator@graphene $ fd --hidden --search-path ./
tavianator@graphene $ rg --hidden --files .
[...]

PR #1521, opened 2024-03-14
Gitignore workaround

tmccombs:
Workaround for bug with gitinored files

This is a workaround for
https://github.com/BurntSushi/ripgrep/issues/2711

Fixes: #1461

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ## Bugfixes
 
 - Respect NO_COLOR environment variable with `--list-details` option. (#1455)
-- Fix bug that would cause hidden files to be includeddespite gitignore rules
+- Fix bug that would cause hidden files to be included despite gitignore rules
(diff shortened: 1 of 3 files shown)


tavianator (changes requested):
I think this workaround is incomplete: we should do the same thing if the user passes `.` as a path explicitly.

It's also worth adding a test case I think.

tavianator (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 ## Bugfixes
 
 - Respect NO_COLOR environment variable with `--list-details` option. (#1455)
+- Fix bug that would cause hidden files to be includeddespite gitignore rules
```
sharkdp, line 10: minor
```suggestion
- Fix bug that would cause hidden files to be included despite gitignore rules
```

sharkdp (approved):
Thank you!

PR #1522, opened 2024-03-19
Add internationalization function

bin-ly:
Use the fluent library to add internationalization functions to the fd project: the translations directory displays the resources required for translation. The i18.rs file is a file that implements internationalization. It obtains the system language through the get_locale() method and displays text according to the language. For example, if the system language is en-US, Chinese will be displayed. If there is no language translation resource (translations directory) set by the system, English will be displayed by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ readme = "README.md"
 repository = "https://github.com/sharkdp/fd"
 version = "9.0.0"
 edition= "2021"
-rust-version = "1.70.0"
+rust-version = "1.77.2"
 
 [badges.appveyor]
 repository = "sharkdp/fd"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.4"
+version = "0.6.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ab91ebe16eb252986481c5b62f6098f3b698a45e34b5b98200cf20dd2484a44"
+checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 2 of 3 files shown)


src/i18.rs
```diff
@@ -0,0 +1,20 @@
...
+use unic_langid::LanguageIdentifier;
+
+pub fn translate(change: &str) -> String{
+    let locale = get_locale().unwrap_or_else(|| String::from("en-US"));
+
+    let langid: LanguageIdentifier = locale.parse().expect("wrong language"); 
+    let locales = vec![langid.into()];
+    let resources = vec!["message.ftl".into()];
+    let mgr = ResourceManager::new("./translations/{locale}/{res_id}".into()); 
+    let bundle = mgr.get_bundle(locales, resources);
```
tmccombs, line 13: Instead of having to create a new bundle every time we call `translate`, we should only create the bundle once. 

We should use a `OnceCell` to lazily initialize the a `ResoruceBundle` the first time it is needed, and then re-use that instance afterwards.

src/i18.rs
```diff
@@ -0,0 +1,20 @@
...
+    let value = bundle.get_message(change).expect("Message doesn't exist.");
+    let pattern = value.value().expect("Message has no value."); 
+
+    let mut errors = vec![];
+    let msg = bundle.format_pattern(&pattern, None, &mut errors).to_string();
```
tmccombs, line 18: We should avoid the `to_string` if possible. If the bundle is moved into a static `OnceCell` then we can return the `Cow<str>` that `format_pattern` returns.

translations/en-US/message.ftl
```diff
@@ -0,0 +1,9 @@
+Search_hidden_files_and_directories = Search hidden files and directories
+
+A_program_to_find_entries_in_your_filesystem = A program to find entries in your filesystem
+
+Bugs_can_be_reported_on_GitHub = Bugs can be reported on GitHub: https://github.com/sharkdp/fd/issues
+
+Print_helps = Print helps
```
tmccombs, line 7:
```suggestion
Print_help = Print help
```

translations/en-US/message.ftl
tmccombs, line 1: the `Makefile` will need to be updated to include these files in an installation.

tmccombs (changes requested):
I'm concerned about the impact this will have on performance, especially startup performance, since it will require opening, reading, and parsing a fluent translations file. 

Is there any precedent for doing something like this in a simple cli program. 

This PR primarily deals with the help text. I wonder if we could get most of the benefit from this, with less impact on performance, just by translating the man page, and maybe using translated help texts if the locale is non-english and the `--help` or `-h` option is used. 

Another concern I have is maintenance. How will we keep these translations up to date? Especially since, AFAIK, none of the maintainers understands chinese?

Cargo.toml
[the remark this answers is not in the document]
bin-ly: 1.Executable file size: The executable file size is 72M before internationalization is added, and 79M after internationalization is added. The strings that fd needs to be internationalized do not seem to be particularly large and will not cause the executable file to be too large.
2.Regarding performance: I used the hyperfine benchmark tool for comparison, and the performance difference is not particularly big.
result
3.Regarding maintenance: I am from China and am very interested in rust and fd projects. If the project requires maintainers who understand Chinese, I will be very happy to participate.

src/cli.rs
```diff
@@ -22,10 +23,12 @@ use crate::filter::SizeFilter;
 #[command(
     name = "fd",
     version,
-    about = "A program to find entries in your filesystem",
-    after_long_help = "Bugs can be reported on GitHub: https://github.com/sharkdp/fd/issues",
+    about = translate("A_program_to_find_entries_in_your_filesystem"),
```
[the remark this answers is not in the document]
bin-ly: This is okay. I was thinking that developers would be more intuitive when reading the code, but it seems that they can also follow the prompts to look at the translations directory.

src/i18.rs
```diff
@@ -0,0 +1,20 @@
+
+use fluent_resmgr::resource_manager::ResourceManager;
+use sys_locale::get_locale;
+use unic_langid::LanguageIdentifier;
+
+pub fn translate(change: &str) -> String{
```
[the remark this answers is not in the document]
bin-ly: it seems "message" is better

src/i18.rs
```diff
@@ -0,0 +1,20 @@
...
+    let locale = get_locale().unwrap_or_else(|| String::from("en-US"));
+
+    let langid: LanguageIdentifier = locale.parse().expect("wrong language"); 
+    let locales = vec![langid.into()];
+    let resources = vec!["message.ftl".into()];
+    let mgr = ResourceManager::new("./translations/{locale}/{res_id}".into());
```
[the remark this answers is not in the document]
bin-ly: This part is to obtain international resources, that is, the translation files required for internationalization. The current processing method is to store the translations directory in the /usr/share/fd directory, and modify the code here to the absolute path, that is, "/usr/share/fd/translations/{locale}/{res_id}"

src/i18.rs
bin-ly: I may need to research this area again. It seemed like setting a global variable so that it could only be called once did not work.

translations/en-US/message.ftl
bin-ly: fluent不支持翻译的字符串之间有空格，所以采用了‘_’代替

bin-ly: Fluent does not support spaces between translated strings, so ‘_’ is used instead.

tmccombs: My point is it should be "help" not "helps"

src/i18.rs
tmccombs: I think you would have something like

```rust

static resource_manager:  OnceCell<ResourceManager> = OnceCell::new();
static bundle: OnceCell<FluentBundle<&'static FluentResource> = OnceCell::new();


# and then in this function you would have something like:

let mgr = resource_manager.get_or_init(|| {
   // we still need to fix this path...
   ResourceManager::new("./translations/{locale}/{res_id}".into()); 
});

let bundle = bundle.get_or_init(|| {
  let locale = get_locale().unwrap_or_else(|| String::from("en-US"));
   let langid: LanguageIdentifier = locale.parse().expect("wrong language"); 
    let locales = vec![langid.into()];
    let resources = vec!["message.ftl".into()];
   mgr.get_bundle(locales, resources)
});

src/i18.rs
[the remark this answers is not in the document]
tmccombs: > The current processing method is to store the translations directory in the /usr/share/fd directory, and modify the code here to the absolute path, that is, "/usr/share/fd/translations/{locale}/{res_id}"

It would be better if the packaging process doesn't require modifying the code. 

And what if it is just installed from a zip file at an unknown location. How will fd know where the translation files are?

What happens if it can't find the translation files?

Issue #1457, opened 2023-12-21
New .git Handling has no opt out


raphael-ratepay:
I have the following code:

```
fd --type d --base-directory ~/Source -H '\.git' | sed 's|/.git||'
```
which lists all folders that contain a `.git` folder (basically a project search). This worked fine in 8.x.x but breaks in 9.x.x due to the breaking change of the .git handling.

Sadly there is no option to opt out for this. because --no-ignore-vcs does bring back all ignored files (so .terraform, .terragrunt-cache, .node-modules, ...) which I do not want. I just want to include ".git" but not everything .gitignore

How can this be achieved? Could you bring back the old behaviour with some cli flag?

sharkdp:
@skoriop FYI

tmccombs:
@raphael-ratepay in the example you gave I don't think adding `--no-ignore-vcs` would work, unless you had a `.git` directory inside a directory that was itself gitignored.  Although, it would still search ignored files like `node_modules` which could hurt performance.

One possible way we could address is this add a `--ignore-pattern` option that lets you add ignore patterns inline (like the inverse of the `--glob` option in ripgrep), and then you could do `fd --ignore-pattern '!.git' ...` to add it back.

p.s. what you currently have would also find something like a `dir/.github` directory and change it to `dirhub` you probably want a `$` at the end of your pattern.

jlucktay:
I was just bitten by this as well. Would like to be able to discover the paths of git repositories, whilst still honouring their respective `.gitignore` files, please. 🙏

Console32:
Any update on this, I could provide an MR for `--ignore-pattern` if that is desired?

tmccombs:
Actually fd already has an option like `--ignore-pattern`: `--exclude`. 

Unfortunately,  using `--exclude !.git` doesn't work, due to how overrides are implemented in the `ignore` crate (since it takes a list of "include" rules, we prefix the input with "!" to get an exclude rule). 

I tried implementing an `--include` option that negates a previous exclude or ignore rule, but ran into two problems:

1. The excluded !.git rule is higher priority than the included .git priority. I can sort of hack around this by checking if ".git" was given, and if so, don't add "!.git"
2. If I add a non-negated pattern to the overrides, then it restricts results to _only_ directories, and files that match that pattern. 

I'm not sure what the best path forward is with the current API of the `ignore` crate.

See also https://github.com/BurntSushi/ripgrep/issues/2705.

[...]

PR #1540, opened 2024-04-28
Stop ignore .git folders by default

tmccombs:
Fixes: #1457
See: #1396, #1403


tavianator (approved):
LGTM, just needs a rebase for the Clippy failure I think

sharkdp:
Thank you very much. Should we release a new version with this change?

tmccombs:
Yeah. I'm not sure if it should be a major version or not though

PR #1545, opened 2024-05-02
Add --filter argument that filters results based on command exit-code

braden-godley:
Based on the RFC at #400.

This PR adds in a new argument, `-f` / `--filter <command>` which will filter all the results based on the result of the command. If it returns a non-zero exit code, the result will be filtered out.

This is my first Rust PR, please let me know if there is anything I can do to improve this!

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -776,6 +776,11 @@ impl clap::FromArgMatches for Exec {
                     .get_occurrences::<String>("exec_batch")
                     .map(CommandSet::new_batch)
             })
+            .or_else(|| {
+                matches
+                    .get_occurrences::<String>("filter")
+                    .map(CommandSet::new_filter)
+            })
             .transpose()
             .map_err(|e| clap::Error::raw(ErrorKind::InvalidValue, e))?;
         Ok(Exec { command })
@@ -797,7 +802,7 @@ impl clap::Args for Exec {
                 .allow_hyphen_values(true)
                 .value_terminator(";")
                 .value_name("cmd")
-                .conflicts_with("list_details")
+                .conflicts_with_all(["list_details", "exec_batch"])
                 .help("Execute a command for each search result")
                 .long_help(
                     "Execute a command for each search result in parallel (use --threads=1 for sequential command execution). \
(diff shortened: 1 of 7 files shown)


src/cli.rs
```diff
@@ -851,7 +856,35 @@ impl clap::Args for Exec {
...
                            fd -e rs -X wc -l\
-                     "
+                    "
+                ),
+        )
+        .arg(
+            Arg::new("filter")
+                .action(ArgAction::Append)
+                .long("filter")
+                .short('f')
```
tmccombs, line 866: I'm not sure if we should use a short option for this. After all we don't have a short option for `--exec` or `--exec-batch` which I think are probably more common than `--filter`.  And we may want to use `-f` for something else in the future (maybe a `--format` option? ). It is easier to add this alias later if desired than to remove it, so let's leave it off for now.

src/cli.rs
```diff
@@ -851,7 +856,35 @@ impl clap::Args for Exec {
...
+                .num_args(1..)
+                .allow_hyphen_values(true)
+                .value_terminator(";")
+                .value_name("cmd")
+                .conflicts_with_all(["exec", "exec_batch", "list_details"])
```
tmccombs, line 871: Why does this need to conflict? 

Logically, it makes sense to allow filtering items first, and then passing the results that succeed through to exec or exec-batch.

src/cli.rs
```diff
@@ -851,7 +856,35 @@ impl clap::Args for Exec {
...
+                .help("Execute a command to determine whether each result should be filtered")
+                .long_help(
+                    "Execute a command in parallel for each search result, filtering out results where the exit code is non-zero. \
+                     There is no guarantee of the order commands are executed in, and the order should not be depended upon. \
+                     All positional arguments following --filter are considered to be arguments to the command - not to fd. \
```
tmccombs, line 876: Maybe mention that you can end with a `\;`?

src/cli.rs
```diff
@@ -776,6 +776,11 @@ impl clap::FromArgMatches for Exec {
                     .get_occurrences::<String>("exec_batch")
                     .map(CommandSet::new_batch)
             })
+            .or_else(|| {
```
tmccombs, line 779: Could this be implemented in a way that would allow using it along with exec or exec-batch?

src/exec/command.rs
```diff
@@ -93,6 +93,41 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
     ExitCode::Success
 }
 
+pub fn execute_commands_filtering<I: Iterator<Item = io::Result<Command>>>(
+    path: &std::path::Path,
+    cmds: I,
+    out_perm: &Mutex<()>,
+) -> ExitCode {
+    let mut output_buffer = OutputBuffer::new(out_perm);
+    let path = format!("{}\n", path.to_str().unwrap());
```
tmccombs: why  do we add a newline to the path?

src/exec/command.rs
```diff
@@ -93,6 +93,41 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
     ExitCode::Success
 }
 
+pub fn execute_commands_filtering<I: Iterator<Item = io::Result<Command>>>(
+    path: &std::path::Path,
+    cmds: I,
+    out_perm: &Mutex<()>,
+) -> ExitCode {
+    let mut output_buffer = OutputBuffer::new(out_perm);
```
tmccombs, line 107: Writing to this output buffer skips all the logic in output.rs. 

The filter code shouldn't be handling output at all. 

Rather, this should be used as one of the filters that we use to exclude files to be processed  in walk.rs in the `spawn_senders` function.

src/exec/command.rs
```diff
@@ -11,6 +11,8 @@ struct Outputs {
     stdout: Vec<u8>,
     stderr: Vec<u8>,
 }
+
+/// Used to print the results of commands that run on results in a thread-safe way
```
tmccombs, line 15: Do weweant to print output from the filter command, or just suppress it?

tmccombs (changes requested):

src/exec/command.rs
braden-godley: I based this on how the `--exec` command does the same. Should we change that as well?

src/exec/command.rs
braden-godley: I agree it would make more sense for the filter command to simply reduce the results that are matched before they're output to the user as usual.

src/cli.rs
braden-godley: I like this idea, I'll consider how we can make these commands compatible.

src/exec/command.rs
tmccombs: No, with exec, the output of the command itself is the desired output, but with filter, the desired output is the path (but possibly with some transformations done on it).

The question here is what should we do with the output from the filter command, if any (I imagine in most cases there won't be any).

Some options are:

1.  Pipe to /dev/null or equivalent 
2. Read, and immediately drop
3. Collect into buffer, then write to stderr, so that error messages are preserved.
4. Collect all filter output, then print all of it at the end (probably on stderr). 

I do think if we do print the output it should be to stderr.

src/cli.rs
musjj: I'm guessing that this might have changed since this comment was written, but don't we have `-x` and `-X`?

drewgingerich:
@braden-godley I want to say thanks for working on this! I came here looking for this exact functionality and was delighted to see you'd already put so much thought and effort into it!

PR #1547, opened 2024-05-03
ci: Support large page sizes on AArch64

### LINKED ISSUE #1085: Binary releases of fd crash on Asahi Linux because of jemalloc using 8kb page sizes
Binary releases crash because they don't support 16kb-aligned pages.

I get the following error with the latest binary releases (both from github releases, and the Arch Linux ARM repos):
```
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed
[1]    28699 IOT instruction (core dumped)  ./fd-v8.4.0-aarch64-unknown-linux-gnu/fd
```

Building from source fixes the issue (apparently jemalloc has internal logic to detect page size?)

M1 Macs use 16KB pages by default [at a hardware level](https://github.com/AsahiLinux/docs/wiki/Broken-Software#why-not-just-use-4k-pages-then), so this is a common problem with [Ashai Linux](https://github.com/tikv/jemallocator). 

There are two simple fixes that occur to me:
1. disabling jemalloc for AArch64 Linux
2. Compiling jemalloc to use 64K pages (which will work anywhere). This is what the [ALARM Rust](https://github.com/archlinuxarm/PKGBUILDs/commit/19a1393) package does.

The second one will probably have to be filled in the upstream [jemallactor](https://github.com/tikv/jemallocator) project (in fact that could be a more appropriate to file this issue....)

tavianator:
Fixes #1085.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -134,7 +134,11 @@ jobs:
 
     - name: Build
       shell: bash
-      run: $BUILD_CMD build --locked --release --target=${{ matrix.job.target }}
+      run: |
+        case ${{ matrix.job.target }} in
+          aarch64-*) export JEMALLOC_SYS_WITH_LG_PAGE=16 ;;
+        esac;
+        $BUILD_CMD build --locked --release --target=${{ matrix.job.target }}
 
     - name: Set binary name & path
       id: bin


tmccombs:
Will this break any aarch64 systems? Or will it just use more memory?

tavianator:
I don't think it will break any systems, but I'll test it on mine just to be sure.  I don't have one with large pages though so I can't even validate the fix

tmccombs (approved):

mathomp4:
> I don't think it will break any systems, but I'll test it on mine just to be sure. I don't have one with large pages though so I can't even validate the fix

Well, once it's in a release, I can test it. 

Or I guess I can do the `cargo` route! How would someone who's never used `cargo` outside of a tutorial get and test this? 😄

tavianator:
I think just

```console
$ JEMALLOC_SYS_WITH_LG_PAGE=16 cargo install fd-find
```

should work

tmccombs:
I'll try to put together a release tonight

mathomp4:
Sadly, this did not seem to fix it on my NVIDIA GraceGrace node. Let me try a cargo build...

tavianator:
@mathomp4 Oh, that's my mistake, I didn't realize the cross build uses docker.  I made https://github.com/sharkdp/fd/pull/1549 which should fix it for real.

Issue #1563, opened 2024-05-16
Allow clicking on results in terminal to open folder or file viewer


schallm:
Would love to see something similar to `ripgrep`'s new `--hyperlink-format` option.  ( [14.0.0 release](https://github.com/BurntSushi/ripgrep/releases/tag/14.0.0) ) It so helpful to be able to open matching files and even have it jump line number of matched text.  It would be great if `fd` allowed something similar.  

For example:
- Allowing folders to be opened in the OS's default folder viewer (i.e., Finder, Explorer, ...) with `--folder-hyperlink-format default` and allowing overrides like the following to open directory a new default shell `--folder-hyperlink-format shell`
- File handling could match `ripgrep`'s handler and open the file with the OS's default handler with `--file-hyperlink-format default` and allow overrides for common tools like vscode with `--file-hyperlink-format vscode`

tmccombs:
As a workaround until https://github.com/sharkdp/fd/pull/1571 is released you could use the `--format` option to get this, like so:

```
fd --format $'\x1b]8;;file://'"${HOSTNAME}${PWD}/{}"$'\x1b\\{}\x1b]8;;\x1b\\'
```

You may want to use an alias or wrapper script.

tmccombs:
Could we open a new issue for this?

dandavison:
Absolutely. https://github.com/sharkdp/fd/issues/1597

Issue #1295, opened 2023-04-02
Show found files as hyperlinks just as ls does?


097115:
Could it possible that `fd` would return its results as hyperlinks, just as `ls`, [for instance](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), does:

    --hyperlink[=WHEN]
        hyperlink file names; WHEN can be 'always' (default if omitted), 'auto', or 'never'

(Which -- in some terminal emulators -- would allow to open the found files with hotkeys, like via the "Hints" mechanism in `alacritty`.)

dandavison:
I think this might be nice. However, we can do this with a shell wrapper. Here are a couple of quick attempts at that in bash and nushell, preserving fd's colored output, and assuming we want a link to open the path in VSCode.

**bash:** (requires `ansifilter`)

```bash
function hyperlink() {
    local url="$1"
    local text="$2"
    printf '\e]8;;%s\e\\%s\e]8;;\e\\\n' "$url" "$text"
}

function fd() {
    command fd --color=always "$@" \
    | while read path; do
        abspath=$(readlink -f $(echo -n $path | ansifilter))
        url="vscode://file/$abspath"
        hyperlink $url $path
      done
}
```

**nushell**: (nushell currently doesn't support passing arbitrary arguments on to the wrapped executable)

```nu
export def 'str hyperlink' [text: string] {
  let url = $in
  let osc8 = (ansi -o '8;;')
  let st = (ansi st)
  $"($osc8)($url)($st)($text)($osc8)($st)"
}

export def fd [
  pattern: string,
  path: string = ".",
] {
  ^fd --color=always $pattern $path
  | lines
  | each {
    |s| $'vscode://file/($s | ansi strip | path expand)' | str hyperlink $s
  }
  | to text
}
```

thenbe:
another alternative: piping the output to `ls`

```sh
# hyperlinked_fd
hfd() {
	fd "$@" --type=file -X ls -l --hyperlink=auto
}

```

097115:
@thenbe,

Thanks! Way faster than the previous suggestion :). But, on the other hand, still palpably slower than the plain `fd`.

It's really a pity @sharkdp isn't interested in this feature... :(

tmccombs:
As a workaround until https://github.com/sharkdp/fd/pull/1571 is released you could use the `--format` option to get this, like so:

```
[...]

PR #1571, opened 2024-06-08
Add hyperlink support to fd

tmccombs:
Fixes: #1295
Fixes: #1563

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -276,6 +276,12 @@ Do not colorize output.
 Always colorize output.
 .RE
 .TP
+.B "\-\-hyperlink
+Specify that the output should use terminal escape codes to indicate a hyperlink to a
+file url pointing to the path.
+Such hyperlinks will only actually be included if color output would be used, since
+that is likely correlated with the output being used on a terminal.
+.TP
 .BI "\-j, \-\-threads " num
 Set number of threads to use for searching & executing (default: number of available CPU cores).
 .TP
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -277,10 +277,22 @@ Always colorize output.
 .RE
 .TP
 .B "\-\-hyperlink
-Specify that the output should use terminal escape codes to indicate a hyperlink to a
+Specify whether the output should use terminal escape codes to indicate a hyperlink to a
 file url pointing to the path.
-Such hyperlinks will only actually be included if color output would be used, since
-that is likely correlated with the output being used on a terminal.
+
+The value can be auto, always, or never.
+
+Currently, the default is "never", and if the option is used without an argument "auto" is
+used. In the future this may be changed to "auto" and "always".
+.RS
+.IP auto
+Only output hyperlinks if color is also enabled, as a proxy for whether terminal escape
+codes are acceptable.
+.IP never
+Never output hyperlink escapes.
+.IP always
+Always output hyperlink escapes, regardless of color settings.
+.RE
 .TP
 .BI "\-j, \-\-threads " num
 Set number of threads to use for searching & executing (default: number of available CPU cores).
--- a/src/hyperlink.rs
+++ b/src/hyperlink.rs
@@ -1,0 +1,63 @@
+use crate::filesystem::absolute_path;
+use std::fmt::{self, Formatter, Write};
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
+
+pub(crate) struct PathUrl(PathBuf);
+
+#[cfg(unix)]
+static HOSTNAME: OnceLock<String> = OnceLock::new();
+
+impl PathUrl {
+    pub(crate) fn new(path: &Path) -> Option<PathUrl> {
+        Some(PathUrl(absolute_path(path).ok()?))
+    }
+}
+
+impl fmt::Display for PathUrl {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "file://{}", host())?;
+        let bytes = self.0.as_os_str().as_encoded_bytes();
+        for &byte in bytes.iter() {
+            encode(f, byte)?;
+        }
+        Ok(())
+    }
+}
+
+fn encode(f: &mut Formatter, byte: u8) -> fmt::Result {
+    match byte {
+        b'0'..=b'9'
+        | b'A'..=b'Z'
+        | b'a'..=b'z'
+        | b'/'
+        | b':'
+        | b'-'
+        | b'.'
+        | b'_'
+        | b'~'
+        | 128.. => f.write_char(byte.into()),
+        #[cfg(windows)]
+        b'\\' => f.write_char('/'),
+        _ => {
+            write!(f, "%{:X}", byte)
+        }
+    }
+}
+
+#[cfg(unix)]
+fn host() -> &'static str {
+    HOSTNAME
+        .get_or_init(|| {
+            nix::unistd::gethostname()
+                .ok()
+                .and_then(|h| h.into_string().ok())
+                .unwrap_or_default()
+        })
+        .as_ref()
+}
+
+#[cfg(not(unix))]
+const fn host() -> &'static str {
+    ""
+}
--- a/src/hyperlink.rs
+++ b/src/hyperlink.rs
@@ -65,16 +65,23 @@ const fn host() -> &'static str {
 mod test {
     use super::*;
 
+    // This allows us to test the encoding without having to worry about the host, or absolute path
+    struct Encoded(&'static str);
+
+    impl fmt::Display for Encoded {
+        fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+            for byte in self.0.bytes() {
+                encode(f, byte)?;
+            }
+            Ok(())
+        }
+    }
+
     #[test]
     fn test_unicode_encoding() {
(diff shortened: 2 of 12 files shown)


sharkdp:
Thank you! Should we maybe make this an option instead of a flag, to avoid future breaking changes in case we see the need for e.g. `--color=never --hyperlink=always` or potentially other ways of printing hyperlinks?

097115:
Something wrong happens if a filename contains non-latin characters.

097115:
@tmccombs This latest commit seems to fix the bug with non-latin chars, thanks :)

doc/fd.1
```diff
@@ -276,6 +276,24 @@ Do not colorize output.
...
 .RE
 .TP
+.B "\-\-hyperlink
+Specify whether the output should use terminal escape codes to indicate a hyperlink to a
+file url pointing to the path.
+
+The value can be auto, always, or never.
+
+Currently, the default is "never", and if the option is used without an argument "auto" is
+used. In the future this may be changed to "auto" and "always".
```
BuonOmo, line 286: Although I love OSC8 as well (thank you for the PR ❤️ ), I'm not sure they're so mainstream that we should go in that direction :/

BuonOmo (approved):
FWIW, LGTM!

The print separator change is quite nice as well, I hope it won't make too noise for the PR to be unchecked though :/ (please add links!)

doc/fd.1
tmccombs: My question is how do terminals that don't support it handle the sequence? If most terminals silently ignore unrecognized OSC sequences, then it probably doesn't hurt to emit it even if it isn't supported. We'll need to test that.

src/hyperlink.rs
```diff
@@ -0,0 +1,87 @@
...
+
+impl PathUrl {
+    pub(crate) fn new(path: &Path) -> Option<PathUrl> {
+        Some(PathUrl(absolute_path(path).ok()?))
+    }
+}
+
+impl fmt::Display for PathUrl {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "file://{}", host())?;
```
tavianator, line 19: Why do we need the hostname?  Does `file:///home/user/...` not work?

tmccombs: It does. As long as fd is running on the same host as the terminal. Including the hostname is useful if you run the command over ssh or in a container or VM. 

See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname

src/hyperlink.rs
```diff
@@ -0,0 +1,87 @@
...
+    // byte.
+    // Percent encoding these bytes is probably safer anyway.
+    match byte {
+        b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'/' | b':' | b'-' | b'.' | b'_' | b'~' => {
+            f.write_char(byte.into())
+        }
+        #[cfg(windows)]
+        b'\\' => f.write_char('/'),
+        _ => {
+            write!(f, "%{:X}", byte)
```
tavianator, line 42: [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) says that the byte `1` should become `%01` not `%1`.  But to avoid having to know that, maybe we could just use https://crates.io/crates/urlencoding?

tmccombs: Unfortanetely, I don't think that `urlencoding` would work, because it would  percent encode "/", which isn't what we want.

doc/fd.1
tmccombs: weston-terminal at least simply ignores the OSC 8 sequence.

sharkdp (approved):
Thank you very much. Looks great.

dandavison:
Hi, this is a great development, but I think that it would be very valuable to allow more customization of the format of the hyperlink. For example, in my case I'd like to be able to specify the protocol (e.g. `vscode://` instead of `file://`). I think that we could re-use a lot of the design work (and some code maybe?) that was done for ripgrep in https://github.com/BurntSushi/ripgrep/pull/2483.

Here's the current full text of the entry for `--hyperlink-format=FORMAT` in `rg --help`:

<details>
<summary>`rg --help` output</summary>
<pre>
    --hyperlink-format=FORMAT
        Set the format of hyperlinks to use when printing results. Hyperlinks
        make certain elements of ripgrep's output, such as file paths,
        clickable. This generally only works in terminal emulators that support
        OSC-8 hyperlinks. For example, the format file://{host}{path} will emit
        an RFC 8089 hyperlink. To see the format that ripgrep is using, pass
        the --debug flag.

        Alternatively, a format string may correspond to one of the following
        aliases: default, none, file, grep+, kitty, macvim, textmate, vscode,
        vscode-insiders, vscodium. The alias will be replaced with a format
        string that is intended to work for the corresponding application.

        The following variables are available in the format string:

        {path}: Required. This is replaced with a path to a matching file. The
        path is guaranteed to be absolute and percent encoded such that it is
        valid to put into a URI. Note that a path is guaranteed to start with a
        /.

        {host}: Optional. This is replaced with your system's hostname. On
        Unix, this corresponds to calling gethostname. On Windows, this
        corresponds to calling GetComputerNameExW to fetch the system's
        "physical DNS hostname." Alternatively, if --hostname-bin was provided,
        then the hostname returned from the output of that program will be
        returned. If no hostname could be found, then this variable is replaced
        with the empty string.

        {line}: Optional. If appropriate, this is replaced with the line number
        of a match. If no line number is available (for example, if
        --no-line-number was given), then it is automatically replaced with the
        value 1.

        {column}: Optional, but requires the presence of {line}. If
        appropriate, this is replaced with the column number of a match. If no
        column number is available (for example, if --no-column was given),
        then it is automatically replaced with the value 1.

        {wslprefix}: Optional. This is a special value that is set to
        wsl$/WSL_DISTRO_NAME, where WSL_DISTRO_NAME corresponds to the value of
        the equivalent environment variable. If the system is not Unix or if
        the WSL_DISTRO_NAME environment variable is not set, then this is
        replaced with the empty string.

        A format string may be empty. An empty format string is equivalent to
        the none alias. In this case, hyperlinks will be disabled.

        At present, ripgrep does not enable hyperlinks by default. Users must
        opt into them. If you aren't sure what format to use, try default.

        Like colors, when ripgrep detects that stdout is not connected to a
        tty, then hyperlinks are automatically disabled, regardless of the
        value of this flag. Users can pass --color=always to forcefully emit
        hyperlinks.

        Note that hyperlinks are only written when a path is also in the output                                                                                                                    11:55:48 [409/3626]
        and colors are enabled. To write hyperlinks without colors, you'll need
        to configure ripgrep to not colorize anything without actually
        disabling all ANSI escape codes completely:

            --colors 'path:none' \
            --colors 'line:none' \
            --colors 'column:none' \
            --colors 'match:none'

        ripgrep works this way because it treats the --color flag as a proxy
        for whether ANSI escape codes should be used at all. This means that
        environment variables like NO_COLOR=1 and TERM=dumb not only disable
        colors, but hyperlinks as well. Similarly, colors and hyperlinks are
        disabled when ripgrep is not writing to a tty. (Unless one forces the
        issue by setting --color=always.)

        If you're searching a file directly, for example:

            rg foo path/to/file

        then hyperlinks will not be emitted since the path given does not
        appear in the output. To make the path appear, and thus also a
        hyperlink, use the -H/--with-filename flag.

        For more information on hyperlinks in terminal emulators, see:
        https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda        
</pre>
</details>

tmccombs:
It's sort of possible to do that today using the `--format` option. That is, you could do something like:


```
fd --absolute-path --format $'\e]8;;vscode://file/{}\e\\{}\e]8;;\e\\' ...
````

I'm not completely opposed to implement multiple formats, but I didn't want to add that level of complexity in the initial implementation.

**Comment by dandavison** [comment]
Thanks! OK, but there are at least two problems with that (which is probably why you said "sort of possible"):

1. There's no color
2. The file path is displayed as absolute, whereas we will usually want relative path displayed, and the absolute path in the hyperlink.
 
Here's the best I've come up with so far (thanks for showing the `$'` trick)

```bash
fd() {
    command fd --color=always --hyperlink=always "$@" |
        rg -r $'\e]8;;vscode://file$1' $'^\e]8;;file://'$(hostname)'(.*)'
}
```

Some examples of editors/IDEs supporting custom protocols I'm aware of are
- [vscode](https://code.visualstudio.com/docs/editor/command-line#_opening-vs-code-with-urls)
- pycharm, intelliJ, webstorm and presumably other JetBrains IDEs
- [zed](https://github.com/zed-industries/zed/issues/8482)

But I imagine there are others. Definitely understand opting for a simple initial implementation but given `fd`'s role of "display paths in terminal for user to do things with",  I suspect `fd` will want to do this ultimately.

So I guess the main question is forwards-compatibility. Seems like e.g. `fd --hyperlink=vscode` would be natural, convenient, and would work fine with the current `always|never|auto` enum.

And then for a fully custom protocol, perhaps `--format` is all that's needed? Sorry if I missed it but where's the documentation of the `--format` template? Would it make sense for it to support both `{absolute-path}` and `{relative-path}` placeholders?

dandavison:
Thanks @tmccombs, I've moved this discussion to the issue: https://github.com/sharkdp/fd/issues/1563

PR #1596, opened 2024-07-22
Fix up some lints on Windows

RossSmyth:
Lints are not run for Windows in CI so they are missed. There are a couple that have been showing for a few months now so here are some fixes.

The reasoning behind the unused field is because of here:
https://github.com/sharkdp/fd/blob/9f0fea6e21f5dfb192f456dd6fb1ff10cd1c1aa0/src/output.rs#L159-L167

Currently in CI you are only running Clippy for Linux. `cargo clippy --all-targets` does not lint for all target triples, but rather all compilation targets (binaries, libraries, tests, examples). So for each target you have cfg'd code, to lint that code you need to run Clippy on that target. Usually in the CI matrix in another step. Cross Clippy is theoretically possible but I've never gotten it to work reliably.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -75,6 +75,7 @@ pub struct Config {
     pub ls_colors: Option<LsColors>,
 
     /// Whether or not we are writing to an interactive terminal
+    #[cfg_attr(not(unix), allow(unused))]
     pub interactive_terminal: bool,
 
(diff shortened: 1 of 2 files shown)


RossSmyth:
oops had the wrong email on my commits

tmccombs (approved):

PR #1608, opened 2024-08-23
Add WSL2 support for hyperlinks

TiltedToast:
I noticed that the hyperlinks generated by fd don't quite work inside wsl2 as you have to use a very specific hostname in this case. 

Thankfully as far as I'm aware every officially supported wsl2 distro comes with the `wslpath` utility that can quickly convert between windows and wsl paths so getting the hostname is as easy as calling that with the root of the file system as an argument. 

Performance wise this shouldn't cause any issues, since it only runs once regardless of how many files are being printed. Other unix environments should be unaffected by this, since it'll just silently fail and use `gethostname` as usual.

### DIFF
--- a/src/hyperlink.rs
+++ b/src/hyperlink.rs
@@ -42,12 +42,23 @@ fn encode(f: &mut Formatter, byte: u8) -> fmt::Result {
 
 #[cfg(unix)]
 fn host() -> &'static str {
-    use std::sync::OnceLock;
+    use std::{process::Command, sync::OnceLock};
 
     static HOSTNAME: OnceLock<String> = OnceLock::new();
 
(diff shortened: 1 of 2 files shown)


tavianator:
Perhaps we can copy how ripgrep does it?  https://github.com/BurntSushi/ripgrep/blob/ea99421ec896fcc9a9cd5367d42a80f89316822f/crates/core/flags/hiargs.rs#L1384-L1406

TiltedToast:
Yeah you're right, that is definitely a better way of doing it. It also seems that ripgrep automatically detects common terminals and uses their specific schemes, seems like something that might be worth looking into hmm

PR #1609, opened 2024-08-30
Support absolute path in exclude 

### LINKED ISSUE #1577: exclude and absolute path as argument
Hello, I have noticed that `--exclude` flag does not accept an absolute path as argument. Indeed

both commmands
```
$ pwd
/home/user
fd --hidden --exclude /home/user/beta kate
fd --hidden --exclude=/home/user/beta kate
```
were still searching `/home/user/beta`. I have noticed that because such path is a SSHFS remote mount and I could notice activity on such path.
To make sure `fd` was not searching inside such path I had to use the relative name

```
fd --hidden --exclude beta kate
fd --hidden --exclude=beta kate
```

I think it would be nice to be able to use absolute path too.
Thank your for your work on fd and have a nice day

cedricbuild:
fixes #1577

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -313,7 +313,23 @@ fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config
             .map(crate::fmt::FormatTemplate::parse),
         command: command.map(Arc::new),
         batch_size: opts.batch_size,
-        exclude_patterns: opts.exclude.iter().map(|p| String::from("!") + p).collect(),
+        exclude_patterns: opts
+            .exclude
+            .iter()
+            .map(|p| {
+                let path = Path::new(p);
+                let relative_path = if path.is_absolute() {
+                    let cwd = env::current_dir().unwrap_or("".into());
+                    path.strip_prefix(cwd)
+                        .unwrap()
+                        .to_string_lossy()
+                        .to_string()
+                } else {
+                    p.to_string()
+                };
+                String::from("!") + &relative_path
+            })
+            .collect(),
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -313,7 +313,23 @@ fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config
...
         command: command.map(Arc::new),
         batch_size: opts.batch_size,
-        exclude_patterns: opts.exclude.iter().map(|p| String::from("!") + p).collect(),
+        exclude_patterns: opts
+            .exclude
+            .iter()
+            .map(|p| {
+                let path = Path::new(p);
+                let relative_path = if path.is_absolute() {
+                    let cwd = env::current_dir().unwrap_or("".into());
```
tmccombs, line 322: This won't work if one or more search paths is given to fd, since it needs to be relative to the search path, which may not be the current directory.

src/main.rs
```diff
@@ -313,7 +313,23 @@ fn construct_config(mut opts: Opts, pattern_regexps: &[String]) -> Result<Config
             .map(crate::fmt::FormatTemplate::parse),
         command: command.map(Arc::new),
         batch_size: opts.batch_size,
-        exclude_patterns: opts.exclude.iter().map(|p| String::from("!") + p).collect(),
+        exclude_patterns: opts
+            .exclude
+            .iter()
+            .map(|p| {
+                let path = Path::new(p);
```
tmccombs, line 320: It isn't a path, it's a pattern.

What if there is a glob that occurs between the root and the search path?

tmccombs (changes requested):

PR #1641, opened 2024-11-29
Add support to find dead symlinks

cgzones:
Find dead symbolic links by reusing the `--type empty` flag.


sorairolake:
I think this can be achieved by running `fd -t symlink -L`. So I don't think there is a need for another way to do this.

See also #1298.

GNU findutils describes about `find -L -type l` which is similar to `fd -t symlink -L`:[^1]

> symbolic link; if '**-L**' is in effect, this is true only for broken symbolic links. If you want to search for symbolic links when '**-L**' is in effect, use '**-xtype**' instead of '**-type**'.

So it may be useful to explain how to search broken links in `README.md` or `fd(1)`.

[^1]: <https://www.gnu.org/software/findutils/manual/html_node/find_html/Type.html>

PR #1652, opened 2024-12-26
feat: adds (no-)ignore-vsc-exclude

### LINKED ISSUE #1612: Only ignore .git/info/exclude
It would be great if there could be a flag such as `--no-ignore-vcs-exclude`, which would be similar to `--no-ignore-vcs` except it would still take `.gitignore` into account but only not `.git/info/exclude`. Maybe this seems like a too specific feature but the reason why I think this would make sense is that I would often use `.git/info/exclude` for files that I actually _do_ want to see and use locally, but only for myself and not commit them to the repo. It's different from `.gitignore` which most likely no one wants to see (therefore including me), and is the reason that this is actually checked into a repo.

Would this make sense? If this feature sounds okay and no one is able to work on it atm I could also take a stab but first wanted to check if it sounds reasonable?

AckslD:
closes #1612 

Let me know if this looks okay, if so I can add tests and update the docs.


src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct Opts {
     #[arg(long, overrides_with = "no_ignore_vcs", hide = true, action = ArgAction::SetTrue)]
     ignore_vcs: (),
 
+    ///Show search results from files and directories that
+    ///would otherwise be ignored by '.git/info/exclude'.
+    ///Git ignore files are still respected unless --no-ignore-vsc is given.
```
tmccombs, line 83:
```suggestion
    ///Git ignore files are still respected unless --no-ignore-vcs is given.
```

src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct Opts {
     #[arg(long, overrides_with = "no_ignore_vcs", hide = true, action = ArgAction::SetTrue)]
     ignore_vcs: (),
 
+    ///Show search results from files and directories that
```
tmccombs, line 81: Nit: Add space between the final "/" and thext of the comment

src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct Opts {
...
+    ///would otherwise be ignored by '.git/info/exclude'.
+    ///Git ignore files are still respected unless --no-ignore-vsc is given.
+    ///The flag can be overridden with --ignore-vcs.
+    #[arg(
+        long,
+        hide_short_help = true,
+        help = "Do not respect .git/info/exclude",
+        long_help
+    )]
+    pub no_ignore_vcs_exclude: bool,
```
tmccombs, line 91: Let's call this "no_ignore_exclude" to match the name ripgrep uses.

tmccombs (changes requested):
This seems fine to me. 

But before merging it should have some tests and a changelog entry.

Issue #1337, opened 2023-06-15
--exclude is not working


UnknownSuperficialNight:
**What version of `fd` are you using?**
[paste the output of `fd --version` here]

version='fd 8.4.0'

im having issues with syntax or fd is just broken

`fd --exclude "/mnt/1.9tbhdd/SteamLibrary" --exclude "/mnt/1.9tbhdd/Back In Time Snapshots" -t f -e mkv -e mp4 --search-path '/mnt/1.9tbhdd' -x bash -c 'echo {}'`

the above command will search for files with the extensions `.mkv` and `.mp4` within the directory `/mnt/1.9tbhdd`, while excluding the two specified directories (`/mnt/1.9tbhdd/SteamLibrary` and `/mnt/1.9tbhdd/Back In Time Snapshots`).

The `bash -c 'echo {}'` is just a example before anyone comments u don't need it it's just here for context

problem im having is `/mnt/1.9tbhdd/SteamLibrary` is still being found even with the --exclude

i tried `! -path 'Dir'` instead of `--exclude 'Dir'` that wont work with `--search-path` so i tried `--full-path` and that does nothing just exits the program

any ideas on what im doing wrong?

tavianator:
`--exclude` is relative to the starting directory, as if it's a `.gitignore` rule basically.  So this should work:

```console
$ fd --exclude /SteamLibrary --exclude "/Back In Time Snapshots" ... /mnt/1.9tbhdd ...
```

IzhakJakov:
### How do you do exclude any directory matching pattern?
e.g: exclude any dir named `templates`?
I tried `fd -E '*/templates/*'` but that only worked on the top level (relative) dirs and not recursive

tmccombs:
Just use "templates"

tmccombs:
So fd's concept of ignore/exclude comes from how .gitignore files work. In git, it doesn't really make sense to use absolute paths in ignore files, because such paths wouldn't be portable when a git repo is used on a different host, and the root of the git repo serves as the base for the ignore rules. 

[...]

PR #1674, opened 2025-02-06
doc: Clarify behavior of --exclude

tmccombs:
Relates-To: #1337

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -268,6 +268,14 @@ as a normal search pattern.
 Exclude files/directories that match the given glob pattern.
 This overrides any other ignore logic.
 Multiple exclude patterns can be specified.
+This uses the same syntax as ignore files (i.e. gitignore) and behaves as if
+.I pattern
+was a line in the global ignore file.
+
+Note that for purposes of ignore rules, the current directory
+is treated as the root, so patterns that begin with / will be anchored
+against the current directory, not treated as absolute paths.
+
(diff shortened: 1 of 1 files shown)


doc/fd.1
```diff
@@ -255,6 +255,14 @@ as a normal search pattern.
 Exclude files/directories that match the given glob pattern.
 This overrides any other ignore logic.
 Multiple exclude patterns can be specified.
+This uses the same syntax as ignore files (i.e. gitignore) and behaves as if
+.I pattern
+was a line in the global ignore file.
+
+Note that for purposes of ignore rules, the current directory
```
tavianator, line 262: "Current directory" is kind of ambiguous.  It's really the *search path* that is treated as the root, not the cwd.

tmccombs: hmmm, this is.... interesting. 

It looks like if there are multiple search paths, then it use the _first_ one  for all:

```
❯ fd
a/
a/other
a/test/
b/
b/a
b/test/
test/

❯ fd --exclude '/test' --search-path a
a/other

❯ fd --exclude '/test' --search-path b
b/a

❯ fd --exclude '/test' --search-path b --search-path b
b/a
b/a

❯ fd --exclude '/test' --search-path b --search-path a
a/other
a/test/
b/a

❯ fd --exclude '/test' --search-path a --search-path b
a/other
b/a
b/test/
```

PR #1690, opened 2025-03-12
chore(deps): Replace `humantime` and `chrono` with `jiff`

### LINKED ISSUE #1689: Replace `humantime` crate
[`humantime`](https://crates.io/crates/humantime) crate seems to be unmaintained and [`jiff`](https://crates.io/crates/jiff) crate provides the same functionality.

ref: rustsec/advisory-db#2249

sorairolake:
Closes #1689

I think the behavior of `fd` is probably the same as before this change, but I don't know if [`jiff`](https://crates.io/crates/jiff) crate is a drop-in replacement for [`humantime`](https://crates.io/crates/humantime) crate, so I added an entry to the changelog.

see also: <https://docs.rs/jiff/0.2.4/jiff/fmt/friendly/index.html#comparison-with-the-humantime-crate>

### DIFF
--- a/src/filter/time.rs
+++ b/src/filter/time.rs
@@ -1,5 +1,4 @@
-use chrono::{DateTime, Local, NaiveDate, NaiveDateTime};
-use jiff::{Span, Zoned};
+use jiff::{civil::DateTime, tz::TimeZone, Span, Timestamp, Zoned};
 
 use std::time::SystemTime;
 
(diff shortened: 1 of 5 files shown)


src/filter/time.rs
```diff
@@ -11,7 +12,8 @@ pub enum TimeFilter {
 
 impl TimeFilter {
     fn from_str(ref_time: &SystemTime, s: &str) -> Option<SystemTime> {
-        humantime::parse_duration(s)
+        s.parse::<Span>()
+            .and_then(Duration::try_from)
```
BurntSushi, line 16: This is going to result in errors if the duration contains calendar units (years, months, weeks or days).

Looking at this code holistically, it would be better to do arithmetic using the `jiff::Span` itself with a `jiff::Zoned` (which you can get via `Zoned::try_from(*ref_time)`). And you might as well replace the `chrono` code below with Jiff. But IDK how much that snowballs into a bigger change.

BurntSushi: You can't parse calendar units directly into a `jiff::SignedDuration`. [This section in the Jiff docs](https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html#comparison-with-the-humantime-crate) might help understanding here.

The short story here is that `humantime` would silently treat things like `1 month` as `30.44` days, regardless of which month it is. Jiff won't let you do that. So if you were previously supporting parsing calendar units via `humantime`, code might need some refactoring.

Some people switching to jiff only used durations for timeouts, where timeouts bigger than hours are not really needed, and so you can get away without refactoring anything and just reject calendar durations (which is what Jiff will do when you parse into a `SignedDuration`). But in this context, my guess is that calendar durations are actually useful.

sorairolake: Fixed to not return an error when the duration contains calendar units.

src/filter/time.rs
```diff
@@ -11,8 +12,11 @@ pub enum TimeFilter {
...
-        humantime::parse_duration(s)
-            .map(|duration| *ref_time - duration)
+        s.parse::<Span>()
+            .and_then(|duration| {
+                Zoned::try_from(*ref_time).and_then(|zoned| zoned.checked_sub(duration))
+            })
+            .map(SystemTime::from)
             .ok()
             .or_else(|| {
                 DateTime::parse_from_rfc3339(s)
```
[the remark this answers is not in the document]
sorairolake: I just want to replace `humantime` with `jiff`. It would be better if someone could replace `chrono` with `jiff`, but I don't want to do that as part of this PR.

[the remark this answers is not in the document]
sorairolake: I'm not sure if I can do it well, but I'll try to replace `chrono` with `jiff` later. If it's successful, I'll add it as part of this PR.

src/filter/time.rs
```diff
@@ -11,32 +11,30 @@ pub enum TimeFilter {
...
+                        s.parse::<DateTime>()
+                            .map(|dt| local_tz.to_ambiguous_zoned(dt))
+                            .and_then(|zdt| zdt.later())
+                            .ok()
                     })
                     .or_else(|| {
                         let timestamp_secs = s.strip_prefix('@')?.parse().ok()?;
-                        DateTime::from_timestamp(timestamp_secs, 0).map(Into::into)
+                        Timestamp::from_second(timestamp_secs)
+                            .map(|ts| ts.to_zoned(TimeZone::UTC))
```
tmccombs, line 33: It's a little unfortunate that we go from a timestmp to a Zoned, and then  back to a SystemTime, which is basically the same as a timestamp. 

But I don't think converting everything else to Timestamp first would be a lot better.

tmccombs (approved):
This looks great. Thanks!

tmccombs:
Oh, one thing that should probably be mentioned in the changelog is that this changes it so fd will no longer accept something like `"1M"` as 1 month.  It now needs to be specified as "1mo" or "1month" etc.

I think that's  probably ok, but should be in the changelog

src/filter/time.rs
BurntSushi: Yeah I think this is fine. I think it might read a little clearer/simpler by just unifying to `Timestamp`, but behaviorally, this is all the same.

PR #1692, opened 2025-03-19
Some small improvements to time filter code after switch to jiff

tmccombs:


src/filter/time.rs
```diff
@@ -58,107 +67,96 @@ mod tests {
     use super::*;
     use std::time::Duration;
 
+    fn set_test_time(time: Zoned) -> SystemTime {
```
tavianator, line 70: To avoid leaking state between tests, I think an API like this would be better:

```rust
fn with_test_time(time: Zoned, f: impl FnOnce()) -> SystemTime
```

And have it reset TESTTIME to `None` with a `Drop` handler.

tavianator: ... or, since you need a drop handler anyway, maybe just

```rust
struct TestTimeGuard;

impl TestTimeGuard {
    fn new(time: Zoned) -> Self { ... }
}

impl Drop for TestTimeGuard { ... }
```

tavianator (approved):

PR #1711, opened 2025-04-24
Update etcetera from 0.9 to 0.10

musicinmybrain:
Pin `home` to `<0.5.11` to avoid MSRV problems. As described in https://github.com/lunacookies/etcetera/releases/tag/v0.10.0. This keeps the MSRV at 1.70.0 (less than `fd`’s 1.77.2), rather than increasing it to 1.81.0 for `etcetera` 0.10.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ regex-syntax = "0.8"
 ctrlc = "3.2"
 globset = "0.4"
 anyhow = "1.0"
-etcetera = "0.9"
+etcetera = "0.10"
 normpath = "1.1.1"
 crossbeam-channel = "0.5.14"
 clap_complete = {version = "4.5.47", optional = true}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,9 +238,9 @@ dependencies = [
 
 [[package]]
 name = "etcetera"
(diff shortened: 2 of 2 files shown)


tavianator:
Thanks!  I PR'd the clippy fix as https://github.com/sharkdp/fd/pull/1712, we can rebase this once that's merged

Cargo.toml
```diff
@@ -44,7 +44,11 @@ regex-syntax = "0.8"
 ctrlc = "3.2"
 globset = "0.4"
 anyhow = "1.0"
-etcetera = "0.9"
+etcetera = "0.10"
+# Temporarily pin home to keep the MSRV of etcetera from increasing to 1.81.0.
+# This direct dependency can be removed after our MSRV is at least 1.81.0.
+# https://github.com/lunacookies/etcetera/releases/tag/v0.10.0
+home = "<0.5.11"
```
tavianator, line 51: This doesn't really work as desired.  If I run `cargo update` I get a duplicated `home` dependency, and the MSRV is still not respected:

```
├── etcetera v0.10.0
│   ├── cfg-if v1.0.0
│   └── home v0.5.11
├── home v0.4.2
```

PR #1716, opened 2025-04-29
Add aarch64 Windows to CI.

### LINKED ISSUE #1658: arm binary for windows
I am using an aarch64 windows instance and would like to have a native binary

jcotton42:
Closes #1658.

I tested the CI run in my fork; you can see the results here https://github.com/jcotton42/fd/actions/runs/14741919681.

I do not have an ARM Windows device to test with at the moment, but CI was successful, and a check with `file` from Git Bash shows the binary is arm64.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -91,6 +91,7 @@ jobs:
           - { target: aarch64-apple-darwin        , os: macos-14                      }
           - { target: x86_64-pc-windows-gnu       , os: windows-2022                  }
           - { target: x86_64-pc-windows-msvc      , os: windows-2022                  }
+          - { target: aarch64-pc-windows-msvc     , os: windows-11-arm                }
           - { target: x86_64-unknown-linux-gnu    , os: ubuntu-22.04, use-cross: true }
           - { target: x86_64-unknown-linux-musl   , os: ubuntu-22.04, use-cross: true }
     env:


jcotton42:
Something I wanted to ask about (unfortunately GitHub doesn't let you leave comments on unchanged lines)

```yaml
    - name: Install Rust toolchain
      uses: dtolnay/rust-toolchain@stable
      with:
        targets: ${{ matrix.job.target }}
        # On windows, for now build with 1.77.2, so that it works on windows 7.
        # When we update the MSRV again, we'll need to revisit this, and probably drop support for Win7
        toolchain: "${{ contains(matrix.job.target, 'windows-') && '1.77.2' || 'stable' }}"
```

Given how new 64-bit ARM Windows is, and that Windows 7 for ARM does not exist, are you still going to use the older 1.77.2 toolchain for aarch64 Windows builds?

tmccombs:
Would it be better to wait until windows-11-arm is out of public preview for this? 

Or at least not fail the build if this fails until then?

I would be fine with using a more recent rust for aarch64 windows.

jcotton42:
> Would it be better to wait until windows-11-arm is out of public preview for this?

I personally think it valuable to have easy access to ARM builds for Windows fd users. As for the image's preview status, while its construction quality is definitely... interesting, the Rust tooling is reasonably mature as far as I can tell.

> I would be fine with using a more recent rust for aarch64 windows.

@sharkdp your thoughts?

PR #1717, opened 2025-04-30
Doc better

tmccombs:


tavianator:
LGTM, save for that typo.  I wonder if we can fix the `fd --help` output easily

PR #1736, opened 2025-06-01
chore: Update several packages

tmccombs:
Supercedes https://github.com/sharkdp/fd/pull/1734


tavianator:
I'm okay with bumping the MSRV even though I think it will lose us Windows 7 support.  We can't stay behind forever.

Issue #1741, opened 2025-06-21
[BUG] Changes to build fd on Illumos


nealsid:
### Checks

- [x] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

When the build system is running `configure` for `jemalloc`, an error is reported:

```
  configure: exit 1

  --- stderr
  Invalid configuration `x86_64-unknown-illumos': OS `illumos' not recognized
  configure: error: /bin/sh build-aux/config.sub x86_64-unknown-illumos failed

  thread 'main' panicked at /export/home/neal/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tikv-jemalloc-sys-0.6.0+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7/build.rs:\
388:9:
```

`fd` already disables `jemalloc` based on certain OS/architectures so Illumos can be added to this list.  The GitHub project for `jemalloc` has been archived (https://github.com/jemalloc/jemalloc) so I don't know if it's possible to fix this upstream.  

### Describe what you expected to happen:

Nothing really, since Illumos isn't officially supported.

### What version of `fd` are you using?

bac535ea

### Which operating system / distribution are you on?

```shell
SunOS openindiana 5.11 illumos-23690bf77a i86pc i386 i86pc
```

nealsid:
FYI, there are two test failures on Illumos I'm looking into (I know Illumos isn't a supported platform, just making a best effort to get it to work).  The first one looks strange, it's as if `-n` is being printed to the console instead of processed as a command line switch to `echo`.  If I run `echo -n foo` in a terminal on Illumos it behaves as expected (i.e. it suppresses printing a newline)

```
---- test_exec_multi stdout ----

thread 'test_exec_multi' panicked at tests/testenv/mod.rs:284:13:
`fd foo --exec echo -n {/}:  ; --exec echo {//}` did not produce the expected output.
Showing diff between expected and actual:
-C.Foo2: ./one/two
-a.foo: .
-b.foo: ./one
-c.foo: ./one/two
-d.foo: ./one/two/three
-directory_foo: ./one/two/three
+-n C.Foo2:
+-n a.foo:
+-n b.foo:
+-n c.foo:
+-n d.foo:
+-n directory_foo:
+.
+./one
+./one/two
+./one/two
+./one/two/three
+./one/two/three

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

[...]

PR #1742, opened 2025-06-21
Changes to support building/testing fd on Illumos

nealsid:
Fix for #1741: jemalloc doesn't build on Illumos, so add it to the list of excluded platforms.


tmccombs (approved):

nealsid:
@tmccombs ready to review/merge, thanks!

Issue #1744, opened 2025-06-28
Test refactoring


pwithams:
I saw https://github.com/sharkdp/fd/issues/382 and was interested in helping out.

Specifically regarding https://github.com/sharkdp/fd/issues/382#issuecomment-894977261 I see that integration tests are all under a single `tests/tests.rs`. I took a look at some possible simple test groupings, such as:

```
and.rs  exec.rs  glob.rs  ignore.rs  owner.rs  symlink.rs  tests.rs
```
Shared static globals could be moved to `testenv` or a separate `utils` helper module.

Is there still any interest in splitting these into separate files? Or any other refactorings?

tmccombs:
That seems reasonable to me.

pwithams:
Sounds good, I created a PR: https://github.com/sharkdp/fd/pull/1745

PR #1745, opened 2025-06-29
Split integration tests into separate files

pwithams:
PR for https://github.com/sharkdp/fd/issues/1744.

Total test runtime appears very slightly slower (by ~100ms for me locally), presumably a constant due to the overhead of creating separate binaries for each test file. However, I think the improved organization is probably worth it as it makes it a little clearer how the tests are distributed and to know where to put new tests.

As each test file is created as a separate binary now clippy warns when parts of `testenv` are not used by every test file, so I added `#![allow(dead_code)]` at the top of `testenv/mod.rs` to suppress that. Not sure if there's a better way to handle that.

I'm open to any suggestions for different file layouts too. For example, if there any gaps in current test types a file could be created to highlight that with the goal of filling it out.

### DIFF
--- a/tests/integration/and.rs
+++ b/tests/integration/and.rs
@@ -1,5 +1,3 @@
-mod testenv;
-
 use crate::testenv::{TestEnv, DEFAULT_DIRS, EXTRA_FILES};
 
 /// AND test
--- a/tests/integration/exec.rs
+++ b/tests/integration/exec.rs
@@ -1,5 +1,3 @@
-mod testenv;
-
 use crate::testenv::{get_test_env_with_abs_path, TestEnv, DEFAULT_DIRS, DEFAULT_FILES};
 
 /// Shell script execution (--exec)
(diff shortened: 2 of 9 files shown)


tmccombs:
We could potentially make a single create with mutiple files for the integration tests, like https://momori.dev/posts/organize-rust-integration-tests-without-dead-code-warning/#one-integration-test-crate-with-modules

I'm not sure if the is the right thing to do here or not.

pwithams:
Yeah I like that idea. I see that cargo uses a similar approach:

https://github.com/rust-lang/cargo/pull/5022
https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

Some timings after reorganizing - seems runtime is looking good again:

Original HEAD: `cargo test  3.70s user 1.46s system 539% cpu 0.956 total`
Splitting tests into separate files/binaries: `cargo test  3.51s user 1.40s system 448% cpu 1.096 total`
Using a single crate for tests: `cargo test  3.74s user 1.39s system 584% cpu 0.879 total`

I renamed the original `tests/tests.rs` file to `tests/integration/general.rs` to try make it clear each file is a related grouping of tests.

PR #1766, opened 2025-08-05
feat: use ANSI colors for default LS_COLORS

arkriny:
Currently, if LS_COLORS is unset, `fd` falls back to a hardcoded color scheme. While these hardcoded colors look fine on most dark backgrounds, they are not guaranteed to work well on light themes - or even all dark themes - because the fixed colors may clash with the user's terminal palette.

Even when contrast is acceptable, hardcoded colors can conflict with the overall tone or saturation of the theme, whereas the terminal-defined colors will always match overall look.

<details>
<summary>Comparison using the Adwaita theme(default theme/ANSI):</summary>
1754407651_screenshot
1754407666_screenshot
</details>

ANSI colors tend to provide better contrast because they are used by many tools, including `ls` itself. Here’s a clearer demonstration of the problem: even without LS_COLORS, `ls` falls back to reasonable highlighting, while `fd` does not. And it's not about the specific colors looking better - it's about consistency.
1754408558_screenshot

I'm using dircolors simply because it's the easiest way to generate an ANSI-based LS_COLORS table, but other sources could work too (e.g., extended themes for more filetypes). btw, `vivid` currently lacks support for a ANSI themes, AFAIK. The key thing is to rely on terminal-provided colors by default.

Would be happy to hear thoughts on this!


arkriny:
Closed in favor of #1779.

Issue #1031, opened 2022-06-05
Faded colors with solarized light?


djbpitt:
**What version of `fd` are you using?**
fd 8.2.1

In iterm2 (Build 3.4.15) `fd` renders several colors in pale pastel hues that cannot easily be read against the background because of insufficient contrast. `ls -G` does not have this issue. My iterm2 settings are:

* General tab, Theme: Light (High Contrast). "Brighten bold text" is checked.
* Profiles tab, Default profile, Colors panel, Color presets drop-down: Solarized light

LS_COLORS is not set.

I'm not wedded to "Light (High Contrast)" or "Solarized light", but my vision is imperfect and I can't read the screen with dark themes, so I do need something light (but not a bright white background) with strong colors and strong color contrasts. I don't have this issue with any other application that produces colored text output in iterm2. Are there iterm2 settings or `fd` options that would enable me to render output from `fd` that is no less legible than the output of `lg -G`?

tavianator:
Can you paste the output of

```
$ ls --color=always path/to/file | xxd
$ fd --color=always -p path/to/file | xxd
```

where `path/to/file` is a file with different colors for `ls` and `fd`?

djbpitt:
Thank you for the quick response. Here is the output:

***@***.*** pr-app % ls --color=always | xxd
00000000: 4c49 4345 4e53 450a 5245 4144 4d45 2e6d  LICENSE.README.m
00000010: 640a 1b5b 3334 6d62 7569 6c64 1b5b 3339  d..[34mbuild.[39
00000020: 3b34 396d 1b5b 306d 0a62 7569 6c64 2e78  ;49m.[0m.build.x
00000030: 6d6c 0a63 6f6c 6c65 6374 696f 6e2e 7863  ml.collection.xc
00000040: 6f6e 660a 636f 6e74 726f 6c6c 6572 2e78  onf.controller.x
00000050: 716c 0a63 7970 7265 7373 2e6a 736f 6e0a  ql.cypress.json.
00000060: 1b5b 3334 6d64 6174 611b 5b33 393b 3439  .[34mdata.[39;49
00000070: 6d1b 5b30 6d0a 1b5b 3334 6d64 6576 1b5b  m.[0m..[34mdev.[
00000080: 3339 3b34 396d 1b5b 306d 0a65 7870 6174  39;49m.[0m.expat
00000090: 682d 706b 672e 786d 6c0a 6963 6f6e 2e70  h-pkg.xml.icon.p
[...]

PR #1779, opened 2025-08-20
Change from hardcoded colors to env default colors

joelparkerhenderson:
The `fd` application displays hard-to-read colors on some systems. This commit replaces the hard-to-read colors with the system env default colors.

See full discussion here:

https://github.com/sharkdp/fd/issues/1031#issuecomment-3206229342

I'm submitting this PR on behalf of the full discussion group, because there seems to be some technical glitch with the original author submitting a PR.


FelikZ:
@sharkdp seem good PR

tmccombs:
One significant downside to this is the lscolors default doesn't have colors for as many different file types as what fd currently uses. And given the limited set of colors available in the set of 16 ANSI colors, there probably isn't much that can be done about that.

PR #1800, opened 2025-09-30
feat(output): implement YAML output

### LINKED ISSUE #1765: Request for `--json` flag
This would help people who like:

- jq
- nushell

Thanks!

Dustin-Jiang:
Implement `--yaml` switch for YAML format output, which allows using `yq` or nushell to interact with the result. 

nushell result

Actually I was initially working on #1765, but could not reach to a perfect state. JSON *afraids* trailing commas, making it *not so easy* to statelessly stream the result without a buffer. Considering a large number of potiential results, it would be memory-consuming to store all lines and print them at last. 

On the other side, the YAML format, while infamous for its complexity on parsing, is friendly for streaming output. No need for serializing or extra dependencies, the simple and fast `write!` is all you need.

There are tools like `yq` that work just like beloved `jq`, and nushell supports YAML as well, so I suggest this PR might be able to close #1765.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -647,6 +647,16 @@ pub struct Opts {
     )]
     search_path: Vec<PathBuf>,
 
+    /// Print results as YAML objects so you can use it with tools like yq and nushell.
+    #[arg(
+        long,
+        value_name = "yaml",
+        conflicts_with("format"),
+        conflicts_with("list_details"),
+        help = "Print results as YAML objects so you can use it with tools like yq and nushell."
+    )]
+    pub yaml: bool,
+
     /// By default, relative paths are prefixed with './' when -x/--exec,
     /// -X/--exec-batch, or -0/--print0 are given, to reduce the risk of a
     /// path starting with '-' being treated as a command line option. Use
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::io::{self, Write};
+use std::os::unix::fs::PermissionsExt;
 
 use lscolors::{Indicator, LsColors, Style};
 
@@ -24,6 +25,8 @@ pub fn print_entry<W: Write>(stdout: &mut W, entry: &DirEntry, config: &Config) 
 
     if let Some(ref format) = config.format {
         print_entry_format(stdout, entry, config, format)?;
+    } else if config.yaml {
+        print_entry_yaml_obj(stdout, entry, config)?;
     } else if let Some(ref ls_colors) = config.ls_colors {
         print_entry_colorized(stdout, entry, config, ls_colors)?;
     } else {
@@ -172,4 +175,59 @@ fn print_entry_uncolorized<W: Write>(
         stdout.write_all(entry.stripped_path(config).as_os_str().as_bytes())?;
         print_trailing_slash(stdout, entry, config, None)
     }
+}
+
+fn print_entry_yaml_obj<W: Write>(
+    stdout: &mut W,
+    entry: &DirEntry,
+    config: &Config,
+) -> io::Result<()> {
+    let path = entry.stripped_path(config);
+    let path_string = path.to_string_lossy();
+    let file_type = entry
+        .file_type()
+        .map(|ft| {
+            if ft.is_dir() {
+                "directory"
+            } else if ft.is_file() {
+                "file"
+            } else if ft.is_symlink() {
+                "symlink"
+            } else {
+                "other"
+            }
+        })
+        .unwrap_or("unknown");
+
+    // Manually construct a simple YAML representation
+    // to avoid adding a dependency on serde_yaml (deprecated).
+    //
+    // A little bit dirty, but safe enough for buffered output.
+    let mut result = format!("- path: \"{}\"\n  type: {}\n", path_string, file_type);
+    let metadata = entry.metadata();
+    if !metadata.is_none() {
+        if let Some(meta) = metadata {
+            result.push_str(&format!("  size: {}\n", meta.len()));
+            result.push_str(&format!(
+                "  mode: {:o}\n",
+                meta.permissions().mode() & 0o7777
+            ));
+            if let Ok(modified) = meta.modified() {
+                if let Ok(duration) = modified.duration_since(std::time::UNIX_EPOCH) {
+                    result.push_str(&format!("  modified: {}\n", duration.as_secs()));
+                }
+            }
+            if let Ok(accessed) = meta.accessed() {
+                if let Ok(duration) = accessed.duration_since(std::time::UNIX_EPOCH) {
+                    result.push_str(&format!("  accessed: {}\n", duration.as_secs()));
+                }
+            }
+            if let Ok(created) = meta.created() {
+                if let Ok(duration) = created.duration_since(std::time::UNIX_EPOCH) {
+                    result.push_str(&format!("  created: {}\n", duration.as_secs()));
+                }
+            }
+        }
+    }
+    write!(stdout, "{}", result)
 }
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,6 +1,5 @@
 use std::borrow::Cow;
 use std::io::{self, Write};
-use std::os::unix::fs::PermissionsExt;
 
 use lscolors::{Indicator, LsColors, Style};
 
@@ -8,6 +7,9 @@ use crate::config::Config;
 use crate::dir_entry::DirEntry;
 use crate::fmt::FormatTemplate;
 use crate::hyperlink::PathUrl;
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
 
 fn replace_path_separator(path: &str, new_path_separator: &str) -> String {
     path.replace(std::path::MAIN_SEPARATOR, new_path_separator)
@@ -183,7 +185,7 @@ fn print_entry_yaml_obj<W: Write>(
     config: &Config,
 ) -> io::Result<()> {
     let path = entry.stripped_path(config);
-    let path_string = path.to_string_lossy();
+    let path_string = path.to_string_lossy().escape_default().to_string();
     let file_type = entry
         .file_type()
         .map(|ft| {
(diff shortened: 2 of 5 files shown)


tmccombs:
I think i would really prefer json output. There are more tools that would support that as input. 

As for the streaming problem, I think there are a couple of ways to handle that:


1. Write the comma at the beginning of each entry instead of the end 
2. Use newline delimited json (ndjson) instead of a json array. Tools like jq can still parse this.
3. Just don't write the last item until we either have the next one, or have reached the end, so we know if we need a comma or not

Dustin-Jiang:
> I think i would really prefer json output. There are more tools that would support that as input.

Yeah definitely, but the original functions in `output.rs` were stateless, so I have to refactor it to make it stateful. 

The good news is, fortunately, that we gain a little bit performance improvement after refactoring, maybe because of less param passing (I guess)?

图片

tmccombs (changes requested):
Should also have some tests, and the man page should be updated. 

As well as the things I commented on above.

doc/fd.1
```diff
@@ -503,6 +503,23 @@ Maximum number of arguments to pass to the command given with -X. If the number
...
+yaml.
+
+Currently, the default is "plain", and if the option is used without an argument "plain"
+is used. The available options are:
+.RS
+.IP plain
+Output the results as plain text (default).
+.IP json
+Output the results as a JSON array.
+.IP ndjson
```
tmccombs: I'm sorry I misdirected you, but after doing some more research it seems like "jsonl" or "JSON Lines" (mime type application/jsonl) is a more common name for this format. So maybe we should use that format.

src/cli.rs
```diff
@@ -647,26 +647,17 @@ pub struct Opts {
...
-        conflicts_with("list_details"),
-        help = "Print results as YAML objects so you can use it with tools like yq and nushell."
-    )]
-    pub yaml: bool,
-
-    /// Print results as a JSON array so you can use it with tools like jq and nushell.
-    #[arg(
-        long,
-        value_name = "json",
+        value_name = "output",
```
tmccombs: I'm still on the fence on this. `--output` allows a more general framework for output formats, but `--json` or `--jsonl` is easier to use. And `ripgrep` uses `--json` (but outputs jsonl).

I'm also not totally convinced that we need more than just jsonl. What is the use case for using yaml or a full json array? I suppose yaml is probably more human readable.  Maybe a tabular format could make sense at some point?

I'm curious what @sharkdp's thoughts are.

src/output.rs
```diff
@@ -1,33 +1,65 @@
...
-}
-
 fn replace_path_separator(path: &str, new_path_separator: &str) -> String {
     path.replace(std::path::MAIN_SEPARATOR, new_path_separator)
 }
 
+#[cfg(unix)]
+fn encode_path(path: &std::path::Path) -> PathEncoding {
+    use std::os::unix::ffi::OsStrExt;
+    let bytes = path.as_os_str().as_bytes();
```
tmccombs: As of rust 1.74 we can actually use [as_encoded_bytes](https://doc.rust-lang.org/stable/std/ffi/struct.OsStr.html#method.as_encoded_bytes) to get a slice of bytes in a platform-independent way.

src/output.rs
```diff
@@ -55,27 +115,38 @@ impl<'a, W: Write> Printer<'a, W> {
...
 
         if has_hyperlink {
             write!(self.stdout, "\x1B]8;;\x1B\\")?;
         }
 
         self.started = true;
+
+        if matches!(
+            self.config.output,
+            OutputFormat::Json | OutputFormat::Ndjson
```
tmccombs:
```suggestion
            OutputFormat::Json 
```
I think we should allow nul output for `Ndjson`. Although then it is really nul separated instead of newline separated.

src/cli.rs
Dustin-Jiang: In my opinion, JSONL, as an extension to JSON format, naturally gets much fewer support than JSON and YAML.

For example, I googled some time and found out that I have to use `jq --stream` , not `--jsonl` or `--ndjson` to parse a JSONL stdin. As another example, the `nushell` supports JSON, YAML and TOML out-of-box, but can only work with JSONL with an extension. 

Looking forward to sharkdp's thoughts.

BurntSushi: As another data point here, ripgrep's `--json` flag emits JSONL. It has been doing so for years and I don't think I've ever heard anyone complain that was using JSONL instead of just JSON. I indeed chose JSONL because it made streaming the output essentially trivial.

src/output.rs
```diff
@@ -1,33 +1,65 @@
...
 
+#[cfg(unix)]
+fn encode_path(path: &std::path::Path) -> PathEncoding {
+    use std::os::unix::ffi::OsStrExt;
+    let bytes = path.as_os_str().as_bytes();
+
+    // Try to convert to UTF-8 first
+    match std::str::from_utf8(bytes) {
+        Ok(utf8_str) => {
+            let escaped: String = utf8_str.escape_default().collect();
```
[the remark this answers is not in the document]
BurntSushi: I would strongly encourage y'all to look at how [ripgrep encodes paths](https://docs.rs/grep-printer/latest/grep_printer/struct.JSON.html#text-encoding). And in particular, you really really really should not be using WTF-8 in your output format. From the [WTF-8 spec](https://wtf-8.codeberg.page/#intended-audience):

> Any [WTF-8](https://wtf-8.codeberg.page/#wtf-8) data must be converted to a [Unicode](https://wtf-8.codeberg.page/#unicode-text) encoding at the system’s boundary before being emitted. [UTF-8](https://wtf-8.codeberg.page/#utf-8) is recommended. [WTF-8](https://wtf-8.codeberg.page/#wtf-8) must not be used to represent text in a file format or for transmission over the Internet.

You'll also want to check how `PathEncoding` serializes. Sometimes just serializing a `&[u8]` leads to something sub-optimal (like an array of integers or something). This is why ripgrep base64 encodes data that isn't valid UTF-8.

tavianator:
Just want to throw out a reference to [libxo](https://juniper.github.io/libxo/libxo-manual.html) as my favourite way for command line programs to support structured output formats.  Not sure there's something like that in the Rust ecosystem.

src/output.rs
```diff
@@ -1,175 +1,290 @@
...
+        if let Some(accessed) = &detail.accessed {
+            write!(self.stdout, ",\"accessed\":\"{}\"", accessed)?;
+        }
+        if let Some(created) = &detail.created {
+            write!(self.stdout, ",\"created\":\"{}\"", created)?;
+        }
+        write!(self.stdout, "}}")
+    }
+
+    fn print_entry_detail(&mut self, format: OutputFormat, entry: &DirEntry) -> io::Result<()> {
```
tmccombs: nit: We could potentially combine this with the  print_entry_json_obj method now, since it is only used for jsonl. 

OTOH, it may be useful in the future if we add a yaml format later, or maybe a tabular format?

phiresky:
It would be useful if there was a distinction between fd --json invocations that do stat calls vs ones that don't. Because for large directories, the stat calls are what dominates runtime, and fd can run e.g. 10x faster without them.

Compare for example

```
strace -f --summary-only fd -I >/dev/null
```

with

```
strace -f --summary-only fd -I -S +0B >/dev/null
```

The second call is much slower, because the `-S` forces fd to use stat to find file sizes, without it does not.

For JSON output, there's no reason to not output all the info we have - but we don't have the stat info by default.

For example the param could be called basic `--json=basic` and `--json=stat`.

tmccombs:
There isn't much we can output without a full stat call. Just the filename, the file type, and the inode number (on unix).

tmccombs:
@sharkdp since this is a pretty significant change, I'd like to confirm you are ok with this change before merging it.

sharkdp:
> I'd like to confirm you are ok with this change before merging it.

Thank you for asking. This looks like a great feature to have! And thank you for your contribution, @Dustin-Jiang!

While I'm here… the format that we introduce here is something that users will depend on, so it is worth investing some time to come up with a first version of this format that we can hopefully depend on for a long time:
* In particular, I would like us to consider using the same format as ripgrep for paths (see [this comment](https://github.com/sharkdp/fd/pull/1800#discussion_r2469282754) by BurntSushi).
* Also, I would really like to see a unit being included in the size-field (`size_bytes`).
* I was also wondering if `"mode"` should be a string? It's not supposed to be read as a decimal number, so serializing it as such feels wrong(?)

tmccombs:
> I was also wondering if "mode" should be a string? It's not supposed to be read as a decimal number, so serializing it as such feels wrong

I'm kind of split on this. Decimal is a pretty bad format for human consumption. But if this json is then consumed programatically, it is probably more convenient to have it as a number than a string. 

Perhaps using the octal encoding as a string could be a good middle graound?

tmccombs:
@Dustin-Jiang, if you want, I could help make those changes to this PR.

sharkdp:
> > I was also wondering if "mode" should be a string? It's not supposed to be read as a decimal number, so serializing it as such feels wrong
> 
> I'm kind of split on this. Decimal is a pretty bad format for human consumption. But if this json is then consumed programatically, it is probably more convenient to have it as a number than a string.

Yes, exactly. The problem with human consumption is that, even if unlikely, it is technically possible for a mode to be `----rwxrwx`, and the current format would serialize this as `"mode": 77`, which is confusing/ambiguous. `"mode": "077"` would be pretty clear.

I agree that we should probably optimize for programmatic consumption, though. But even then, the string seems more practical to me? Because I will probably want to split the number by owner/group/others, and that is much easier if it's already in this string notation. Converting a single digit from a character to an integer should be an operation that is easily available in every programming language.

It gets more tricky actually if we consider stricky/setgid/setuid where the mode can be something like `4777`. That suggests that we should always serialize it as a four-character string, possibly with a leading zero?

> Perhaps using the octal encoding as a string could be a good middle graound?

How is that different from what I am proposing? Something like "0644" or "4777" would be the octal encoding as a (fixed length) string, right? Or would you suggest "0o0644" / "0o4777"?

tmccombs:
> How is that different from what I am proposing? 

Sorry, I meant as opposed to something like "rwxr-x---", which would probably be the most uset friendly but not very computer friendly in most cases.

PR #1812, opened 2025-10-21
build(deps): bump ignore to 0.4.24

tmccombs:
This should hopefully fix some bugs around gitignore.

Also run `cargo update` to update the lockfile


tmccombs:
well, looks like this requires a newer MSRV, so I'll close this in favor of  #1813

PR #1828, opened 2025-11-06
ci: resolve Zizmor-identified workflow issues

HastD:
* Set dependabot cooldown timer of 7 days, which helps mitigate stability and supply-chain security risks. For more info, see https://docs.zizmor.sh/audits/#dependabot-cooldown
* Restrict write permissions to the job that actually needs them.
* Set `persist-credentials: false` for `actions/checkout`. See https://docs.zizmor.sh/audits/#artipacked
* Use environment variables instead of template expansions in code contexts. See https://docs.zizmor.sh/audits/#template-injection
* Pin action dependencies to SHA hashes.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -14,9 +14,7 @@ on:
       - '*'
 
 permissions:
-  id-token: write
-  contents: write
-  attestations: write
+  contents: read
 
 jobs:
   crate_metadata:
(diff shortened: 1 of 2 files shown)


.github/workflows/CICD.yml
```diff
@@ -262,7 +282,7 @@ jobs:
         subject-digest: sha256::${{ steps.upload-deb.outputs.artifact-digest }}
 
     - name: Publish archives and packages
-      uses: softprops/action-gh-release@v2
+      uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090 # v2.4.1
```
tmccombs: I actually intended to replace this with calling `gh` directly. I just haven't gotten around to it yet.

PR #1841, opened 2025-11-19
Add support for filtering orphaned files with --owner orphan

devmanuelli:
Extends the --owner filter to support finding files with no valid
user/group by using the 'orphan' keyword. This provides equivalent
functionality to find's -nouser and -nogroup flags.

Examples:
  fd --owner orphan       # equivalent to find -nouser
  fd --owner :orphan      # equivalent to find -nogroup
  fd --owner orphan:orphan # both -nouser and -nogroup

Implementation:
- Added Check::Orphaned variant to owner filter
- Extended OwnerFilter::from_string to recognize 'orphan' keyword
- Updated OwnerFilter::matches to check if uid/gid maps to valid user/group
- Added unit tests for orphan parsing

### DIFF
--- a/src/filter/owner.rs
+++ b/src/filter/owner.rs
@@ -1,34 +1,28 @@
 use anyhow::{Result, anyhow};
 use nix::unistd::{Group, User};
-use std::collections::HashSet;
-use std::fs;
-use std::sync::{LazyLock, Mutex};
+use std::{cell::RefCell, collections::HashMap, fs};
 
-static VALID_UIDS: LazyLock<Mutex<HashSet<u32>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
-static VALID_GIDS: LazyLock<Mutex<HashSet<u32>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
+thread_local! {
+    static UID_CACHE: RefCell<HashMap<u32, bool>> = RefCell::new(HashMap::new());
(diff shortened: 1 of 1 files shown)


src/filter/owner.rs
```diff
@@ -134,6 +147,10 @@ mod owner_parsing {
         both_negate:"!4:!3" => Ok(OwnerFilter { uid: NotEq(4), gid: NotEq(3)   }),
         uid_not_gid:"6:!8"  => Ok(OwnerFilter { uid: Equal(6), gid: NotEq(8)   }),
 
+        orphan_uid: "orphan"       => Ok(OwnerFilter { uid: Orphaned, gid: Ignore   }),
```
tmccombs: It is possible that a user or group is actually named "orphan". I think it would probably be better to use a a different option, or maybe use something that isn't a valid username like "-".

devmanuelli:
Addressed all feedback: changed keyword to `-`, added validator function to Orphan variant for self-contained checking, added caching, and updated docs.

devmanuelli:
bump @tmccombs

devmanuelli:
@tmccombs I should have fixed it

devmanuelli:
bump @tmccombs is there sth I should fix?

devmanuelli:
@tavianator I should have addressed ur review in the last commit

devmanuelli:
> Mostly looks good

@tmccombs 
if there is something smelly just tell me

devmanuelli:
@tavianator bump. ty in advance

PR #1846, opened 2025-11-23
Add colors to --help

starsep:
~I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.~
User can disable colors via `NO_COLOR=1` environment variable

| before | after | `NO_COLOR=1` |
| - | - | - |
| fd-before | fd-colors | fd-no-color |


tmccombs:
fd isn't a clap plugin, so using clap-cargo for it doesn't seem appropriate.

Also, doesn't fd already have colored help output?

starsep:
> fd isn't a clap plugin, so using clap-cargo for it doesn't seem appropriate.

Ok, I will remove dependency as it is actually quite easy to do without it.

> Also, doesn't fd already have colored help output?

At the moment it doesn't. It had before upgrade to clap v4 which is using bold for headers and no longer uses colors by default, even with color feature enabled

tmccombs (approved):

PR #1852, opened 2025-11-28
`--ignore-contain`: ignore directories containing a named entry.

### LINKED ISSUE #1727: feature request: add `--prune-parent` flag
I am building a tool to list git repos.  Currently, its implementation is

```sh
fd '^\.(git|sl)$' "$@" \
	--max-depth=10 --unrestricted --hidden --type=d --follow --prune \
	--exec dirname {} \; \
	| sort

```

On my machine this still is pretty slow (~1m).  Most of the time is being spent looking for repos in repos.  I don't care if my tool finds submodules, so I'd like to exclude them.  `--prune` will prevent `fd` from looking inside `.git/` but I want `fd` to not look at the whole repo. 

Would you accept a patch that adds `--prune-parent`, which would do the behavior I described?

fischman:
Example use-cases include:
- [`CACHEDIR.TAG`](https://bford.info/cachedir/)
- Ignoring entire git checkouts (including submodules) (see linked issue).

Fixes #1727.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -461,11 +461,22 @@ impl WorkerState {
                     return WalkState::Quit;
                 }
 
-                let entry = match entry {
-                    Ok(ref e) if e.depth() == 0 => {
+                if let Ok(e) = &entry {
+                    let entry_path = e.path();
+                    if entry_path.is_dir()
+                        && config
+                            .ignore_contain
+                            .iter()
+                            .any(|ic| entry_path.join(ic).exists())
+                    {
+                        return WalkState::Skip;
+                    }
+                    if e.depth() == 0 {
                         // Skip the root directory entry.
                         return WalkState::Continue;
                     }
+                }
+                let entry = match entry {
                     Ok(e) => DirEntry::normal(e),
                     Err(ignore::Error::WithPath {
                         path,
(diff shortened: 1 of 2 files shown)


fischman:
(this is the only Rust I've ever written; please review carefully)

fischman:
@tmccombs please take a look, thanks.

src/cli.rs
```diff
@@ -528,6 +528,16 @@ pub struct Opts {
     )]
     pub hyperlink: HyperlinkWhen,
 
+    /// Ignore directories containing the named entry.
+    #[arg(
+        long,
+        value_name = "name",
+        hide_short_help = true,
```
tmccombs: I think it is fine to include this in the short help.

src/cli.rs
```diff
@@ -528,6 +528,16 @@ pub struct Opts {
...
+        help = "Ignore directories containing this name",
```
tmccombs: This isn't necessary. By default the help content will be use the doc comment (due to macro magic)

src/cli.rs
```diff
@@ -528,6 +528,16 @@ pub struct Opts {
...
+        long_help
+    )]
+    pub ignore_contain: Option<String>,
```
tmccombs: `ignore_parents_of` might be a clearer name. But it's also quite long.

tmccombs:
Sorry it took so long for me to get to reviewing this. And thanks for putting this together. I think it is pretty close.

src/cli.rs
fischman: Done.

src/walk.rs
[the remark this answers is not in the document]
fischman: Now skipping (and added test). Sorry for not catching the hint earlier.

PR #1860, opened 2025-12-10
feat[completions]: Add value_hints

tmccombs:
Add hints for some arguments to generate better completions.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 
 use anyhow::anyhow;
 use clap::{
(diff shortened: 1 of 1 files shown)


awsms:
could we also have hints for `-x`/`-X` so it auto-completes with commands?
https://github.com/awsms/fd/commit/3c113a1b562a9e7e003521acef49b42d844b6d86

tmccombs:
I'd like to. But unfortunately, the correct hint to use there would be CommandWithArguments, and it isn't currently possible to use that there, due to https://github.com/clap-rs/clap/issues/3266.

PR #1861, opened 2025-12-13
build: Reduce debug info for dev builds

tmccombs:
Hopefully this will sped up dev builds. Adds a new debugging profile if
full debug symbols are needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,16 @@ tempfile = "3.23"
 filetime = "0.2"
 test-case = "3.3"
 
+[profile.dev]
+debug = "line-tables-only"
+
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -76,6 +76,16 @@ tempfile = "3.23"
 filetime = "0.2"
 test-case = "3.3"
 
+[profile.dev]
+debug = "line-tables-only"
+
+[profile.dev.package."*"]
+debug = false
+
+[profile.debugging]
```
tavianator: Why not just `profile.debug`?

tavianator (approved):

PR #1865, opened 2025-12-26
Add unified Filter trait for consistent filtering interface

sergii4:
This commit introduces a Filter trait with a common `matches()` methodto replace the inconsistent `applies_to()` and `is_within()` methods.

  - Add Filter trait to filter/mod.rs with associated type Item
  - Implement Filter for TimeFilter (Item = SystemTime)
  - Implement Filter for SizeFilter (Item = u64)
  - Update walk.rs to use matches() instead of specific methods

  Addresses #382
  
### Benchmark Setup
  - 200 runs with 10 warmups using hyperfine
  - Test: `fd --type f --size +1k .`

### Results
  Master:  12.3 ms ± 2.9 ms  [User: 3.4 ms, System: 3.5 ms]
  This PR: 13.0 ms ± 2.8 ms  [User: 3.4 ms, System: 3.5 ms]
  Ratio: 1.06x ± 0.34

### Analysis
  - **CPU time is identical** (User/System both 3.4/3.5ms)
  - The ~6% difference has ±34% confidence (ranges overlap)
  - Likely measurement noise rather than real overhead
  - Static dispatch with `#[inline(always)]`should be zero-cost

### DIFF
--- a/src/filter/size.rs
+++ b/src/filter/size.rs
@@ -79,7 +79,7 @@ impl SizeFilter {
 impl Filter for SizeFilter {
     type Item = u64;
 
-    #[inline]
+    #[inline(always)]
     fn matches(&self, item: &Self::Item) -> bool {
         self.is_within(*item)
     }


sergii4:
Hey @tmccombs I have seen your comment regarding performance then it disappeared 🤷‍♂️ Anyway I played with benchmark(updated my initial comment) and made a change.

Is this acceptable for the unified interface? Happy to explore alternatives if needed!

src/filter/size.rs
```diff
@@ -74,6 +76,15 @@ impl SizeFilter {
     }
 }
 
+impl Filter for SizeFilter {
+    type Item = u64;
+
+    #[inline(always)]
+    fn matches(&self, item: &Self::Item) -> bool {
+        self.is_within(*item)
```
tmccombs: Same here. Why not just move `is_within` to the body of the matches method?

src/filter/mod.rs
```diff
@@ -9,3 +9,9 @@ mod time;
 
 #[cfg(unix)]
 mod owner;
+
+pub trait Filter {
```
tmccombs: I don't really get the benefit of this trait. Does this give us anything beyond having a consistent name for the filtering methods? 

Is this intended to be used more in an additional refactor?

sergii4:
You're right @tmccombs , this needs more thought. Thanks for reviewing!

PR #1867, opened 2026-01-01
feat(build): enable jemalloc on musl and riscv64

JustSoup312:
Alpine Linux vendors a custom patch to enable jemalloc for all arches, which seems to work fine in practice.


JustSoup312:
I wonder if the 32bit failures might be an Ubuntu issue specifically, because on Alpine the arches build completely fine with jemalloc...?

tmccombs:
Maybe a mismatch between musl and jemalloc?

PR #1872, opened 2026-01-14
Fix Windows hyperlinks for paths with spaces

### LINKED ISSUE #1843: [BUG] Hyperlinks fail on Windows when file path contains spaces
### Checks

Setup (pwsh):
```pwsh
mkdir C:\test
pushd C:\test
New-Item hi.txt
New-Item 'hi there.txt'
```

Run:
```
C:\test
> fd -e txt--hyperlink=auto -d1

hi there.txt
hi.txt
# entries are underlined in Windows Terminal to indicate hyperlinks
```

Image

```
# Observed hyperlinks from fd==10.2.0 on Windows
file://C:/test/hi%20there.txt  # FAILURE
file://C:/test/hi.txt  # SUCCESS
```

The `file://` part is missing the third slash for local host. The issue affects filepaths with spaces (and maybe other escape chars; I haven't checked). Filepaths without spaces open fine, even with two slashes. 

I'm not a Rust person but it looks like adding a slash '/' to the Windows `host` function in `hyperlinks.rs` would fix it for my use case, which is drive-based paths. No idea how UNC paths will behave (they're also broken right now, see below). 

```rust
#[cfg(windows)]
const fn host() -> &'static str {
    "/"
}
```

Other notes:
- if I run from a mapped network drive `N:\test`, I get the same results - 2 slashes - `file://N:/test/hi%20there.txt`
- if I run from a bare UNC path on the network `\\192.168.0.20\share\test`, I'll get 4 slashes `file:////192.168.0.20/share/test/hi%20there.txt`, and Terminal declares the link as unsupported. I know UNC paths are not supported (https://github.com/microsoft/terminal/issues/19236 ) and I personally don't care about this use case at all. But the correct hyperlink would be 2 slashes `file://192.168.0.20/share/test/hi%20there.txt` 

(Terminal Error Message)
```
This link type is currently not supported:
file:////192.168.0.20/share/test/hi%20there.txt 
```

```
# Expected RFC 3986 URLs
file:///C:/test/hi%20there.txt 
file:///C:/test/hi.txt
```

10.2.0

```shell
- Windows 10 and 11
- pwsh 7.4 and 7.5
- Windows Terminal
```

josephadamsdev:
Fixes #1843

This PR fixes hyperlink generation on Windows when file paths contain spaces or other characters that require percent-encoding.

## Problem
On Windows, fd was generating file URLs like `file://C:/test/hi%20there.txt` (2 slashes) which don't work in Windows Terminal. Paths without spaces worked fine, but paths with spaces failed to open.

## Solution  
Changed the `host()` function for Windows to return `"/"` instead of `""`, generating RFC 3986 compliant URLs with three slashes: `file:///C:/test/hi%20there.txt`

## Changes
- Modified `src/hyperlink.rs` line 61 to return `"/"` for Windows
- This matches the RFC 3986 file URL format: `file:///` + drive path

## Testing
The fix addresses the exact issue reported:
- **Before**: `file://C:/test/hi%20there.txt` ❌ (broken)
- **After**: `file:///C:/test/hi%20there.txt` ✅ (works)

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=105917501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ## Bugfixes
 
-- Fix Windows hyperlink generation for paths with spaces. (#1843)
+- Fix Windows hyperlink generation for paths with spaces. (#1872)
 
 - `--print0` combined with `--exec` will now print a `\0` between the output of each entry. Note that if there are multiple instances
   of `--exec`, the `\0` will be between each _set_ of commands, _not_ between each individual command run. Fixes #1797.


crackedsdev:
> It looks like some tests are failing. Pro

checking it immediately

crackedsdev:
Thanks for the heads-up @tmccombs updated the hyperlink test and pushed. CI ran successfully now.

tmccombs:
Thanks for fixing this!

crackedsdev:
> We should probably also add an entry to the changelog as well (sorry, I should have mentioned that in my previous review), but looks good.

Good call @tmccombs I’ll add a changelog entry and push an update shortly. Thanks

Issue #1873, opened 2026-01-15
[BUG] The 'pattern contains path separator' error is only shown when pattern is an existing directory


greenwastaken:
### Checks

- [x] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

If invoked with `fd /`, fd fails with the following error 
```
[fd error]: The search pattern '/' contains a path-separation character ('/') and will not lead to any search results.

If you want to search for all files inside the 'Programs/' directory, use a match-all pattern:

  fd . '/'

Instead, if you want your pattern to match the full file path, use:

  fd --full-path '/'
```

However, this behavior is inconsistent. The error message is only printed when the pattern also happens to be an existing directory. Otherwise, `fd` returns no error, and silently finds no files. 

### Describe what you expected to happen:

I would expect the error about how a pattern with a '/' in it will not lead to any search results to be displayed regardless of the pattern possibly being a directory path.  

### What version of `fd` are you using?

fd 10.3.0

### Which operating system / distribution are you on?

```shell
Linux 6.18.4-arch1-1 x86_64

LSB Version:    n/a
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a
```

marcospb19:
Confirmed, repro steps:

```sh
mkdir -p dir1/dir2
fd r1/dir2   # no complaint
fd dir1/dir2 #    complains
```

tmccombs:
The difficulty in fixing this is that on windows "\" is both the path separator, and the eacape character for regex, so if you detect a path separator it could just be part of a valid regex.

cl2t:
I’d like to work on this.\n\nI have a small fix ready that makes the path-separator error consistent on Unix and adds a regression test.\nI left Windows behavior unchanged because \ can also be a regex escape there.

PR #1876, opened 2026-01-26
Fix: Show path separator error regardless of directory existence

Richman018:
Fixes #1873

The error message about patterns containing path separators was only shown when the pattern was also an existing directory. This change makes the error consistent by showing it whenever a pattern contains a path separator (and `--full-path` is not used), regardless of whether the pattern is an existing directory.

This ensures users get helpful feedback even when they accidentally use a path separator in a pattern that doesn't exist as a directory.

**Changes:**
- Removed the `Path::new(&opts.pattern).is_dir()` check from `ensure_search_pattern_is_not_a_path()`
- The error is now shown consistently whenever a pattern contains a path separator

**Testing:**
- Before: `fd /nonexistent/path` would silently return no results
- After: `fd /nonexistent/path` now shows the helpful error message


Richman018:
Hi, @goyox86 @wezm @reima @cuviper .
Could you review my PR?
Thank you.

PR #1886, opened 2026-02-17
feat: add --override-ignore flag to selectively bypass .gitignore rules

### LINKED ISSUE #1885: --override-ignore flag to selectively bypass .gitignore
Currently, `fd` respects `.gitignore` rules (great for most searches), but sometimes you need to find files inside gitignored directories like `node_modules`, `build`, or `.cache`. The only option today is `--no-ignore` / `-I`, which disables **all** ignore rules.

  This forces users into an awkward workaround, disabling all ignores and then manually re-excluding every other pattern from `.gitignore`:

  ```bash
  # Just want to search inside node_modules, but .gitignore also has
  # build/, dist/, *.pyc, .env*, coverage/, .cache/, tmp/ ...
  # Now you have to re-add ALL of them as --exclude:
  fd -I --exclude build --exclude dist --exclude '*.pyc' \
       --exclude '.env*' --exclude coverage --exclude .cache \
       --exclude tmp pattern
```

  This is tedious, error-prone, and breaks every time .gitignore changes.
  ## Proposed Solution

 Add `--override-ignore <pattern>` flag that force-includes entries matching the given glob pattern, even when they'd be excluded by `.gitignore`, while keeping `.gitignore` active for everything else.

```bash
  # Instead of the mess above, just:
  fd --override-ignore node_modules pattern
```

benodiwal:
Closes #1885

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -305,6 +305,22 @@ pub struct Opts {
     )]
     pub exclude: Vec<String>,
 
+    /// Force-include entries matching the given glob pattern, even if they
+    /// would otherwise be ignored by '.gitignore' files. This overrides
+    /// VCS ignore rules for specific patterns while keeping them active
+    /// for everything else. Multiple patterns can be specified.
+    ///
+    /// Examples:
+    /// {n}  --override-ignore node_modules
+    /// {n}  --override-ignore '*.log'
+    #[arg(
+        long,
+        value_name = "pattern",
+        help = "Override .gitignore for entries matching the given glob pattern",
+        long_help
+    )]
+    pub override_ignore: Vec<String>,
+
     /// Do not traverse into directories that match the search criteria. If
     /// you want to exclude specific directories, use the '--exclude=…' option.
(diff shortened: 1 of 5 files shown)


tmccombs:
This looks like it adds a lot of complexity for a fairly niche feature. 

Is there a reason we can't just use [WalkBuilder::override](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.overrides) for this?

benodiwal:
I initially tried using `WalkBuilder::overrides` for this, but the ignore crate's `Override::matched()` has an implicit blacklist behavior that prevents it.

When you add any whitelist pattern to `OverrideBuilder`, every non-matching file is force-excluded `(Match::Ignore)` instead of falling through `(Match::None)`. So for example whitelisting `node_modules` would hide every other file in the tree. There's no flag to disable this it's hardcoded.

The two-walk approach avoids this by keeping Walk 1 completely unchanged and only running Walk 2 when `--override-ignore` is actually used, so there's zero cost for normal usage.

Open to simpler approaches if I'm missing something!

PR #1887, opened 2026-02-19
Show path-separator error for all patterns containing '/', not just directories

danielalanbates:
## Summary

The "pattern contains path separator" error was only shown when the pattern also happened to be an existing directory (e.g., `fd Programs/` where `Programs/` exists). Patterns like `fd foo/bar` that aren't directories silently returned no results with no diagnostic.

Now the error is always shown when the pattern contains a path separator and `--full-path` is not used. The "search inside directory" suggestion is conditionally shown only when the pattern is an actual directory.

## Example

Before (pattern is not a directory — no error, silent empty results):
```
$ fd foo/bar
$
```

After:
```
$ fd foo/bar
[fd error]: The search pattern 'foo/bar' contains a path-separation character ('/') and will not lead to any search results.

If you want your pattern to match the full file path, use:

  fd --full-path 'foo/bar'
```

Closes #1873

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,7 +147,10 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
 
 /// Detect if the user accidentally supplied a path instead of a search pattern
 fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
-    if !opts.full_path && opts.pattern.contains(std::path::MAIN_SEPARATOR) {
+    if !opts.full_path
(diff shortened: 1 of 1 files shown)


tmccombs:
There is already a [PR for this](https://github.com/sharkdp/fd/pull/1876).

This also doesn't update the changelog, or add any tests.

And it doesn't address the issue that on windows the path separator is also the regex escape character.

danielalanbates:
Thanks for pointing out the existing PR. Apologies for the duplicate!

PR #1889, opened 2026-02-20
Fix symlink for bash completions

### LINKED ISSUE #1888: [BUG] Bash completions do not work for alias fdfind in deb release
### Checks

When using the alias `fdfind` bash completions do not work. It does work when using `fd`.

Steps to reproduce:

1. Download latest deb release for amd64: `fd_10.3.0_amd64.deb`
2. Install with `dbkg -i  fd_10.3.0_amd64.deb`
3. Restart bash shell
4. Try completion for `fdfind` in shell.

`fdfind` should have the same bash completions as `fd`.

fd 10.3.0

```shell
Linux 6.6.87.2-microsoft-standard-WSL2 x86_64

Distributor ID: Ubuntu
Description:    Ubuntu 24.04.4 LTS
Release:        24.04
Codename:       noble
```

skane-lukas:
### Problem
In the deb package a broken symlink is created for bash completions to the alias command `fdfind`.

### Solution
Create static completion files that sources fd completions for fdfind for bash, zsh and fish
Install the completions in Makefile and create-deb.sh

### Changes
- Create static completion files in contrib/completion that sources fd completions for fdfind for bash, zsh and fish
- Changed symlink target for completion `fdfind` in `create-deb` from `./fd.bash` to `./fd`
- Update Makefile and create-deb.sh to install the new static completions.
- Updated `CHANGELOG.md` with the fix.

fixes #1888

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
     - #1667
     - #1813
 
+- Fix completions for alias `fdfind` in deb release, see #1888 (@skane-lukas)
+
 ## Changes
 
 - Minimum required rust version has been increased to 1.90.0. Notably, this means dropping fully support for intel Mac and Windows 7.
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ $(EXE): Cargo.toml src/**/*.rs
 	cargo build --profile $(PROFILE) --locked
 
 .PHONY: completions
-completions: autocomplete/fd.bash autocomplete/fd.fish autocomplete/_fd.ps1 autocomplete/_fd
+completions: autocomplete/fd.bash autocomplete/fd.fish autocomplete/_fd.ps1 autocomplete/_fd autocomplete/_fdfind autocomplete/fdfind.bash autocomplete/fdfind.fish
 
 comp_dir=@mkdir -p autocomplete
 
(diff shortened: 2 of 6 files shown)


skane-lukas:
The completions still dont work because they are generated for `fd` and not `fdfind`.


I can only think of changing the symlinks in `create-deb.sh` to something like:

```bash
ln -sf "/usr/bin/fd" "${DPKG_DIR}/usr/bin/fdfind"

"${DPKG_DIR}/usr/bin/fdfind" --gen-completions bash > autocomplete/fdfind.bash
"${DPKG_DIR}/usr/bin/fdfind" --gen-completions fish > autocomplete/fdfind.fish
"${DPKG_DIR}/usr/bin/fdfind" --gen-completions powershell > autocomplete/_fdfind.ps1
# Generate Zsh completion by substituting top row of static completion file
sed '1s/^#compdef fd/#compdef fdfind/' autocomplete/_fd > autocomplete/_fdfind

install -Dm644 autocomplete/fdfind.bash "${DPKG_DIR}/usr/share/bash-completion/completions/fdfind"
install -Dm644 autocomplete/fdfind.fish "${DPKG_DIR}/usr/share/fish/vendor_completions.d/fdfind.fish"
install -Dm644 autocomplete/_fdfind.ps1 "${DPKG_DIR}/usr/share/powershell/Modules/fdfind/_fdfind.ps1"
install -Dm644 autocomplete/_fdfind "${DPKG_DIR}/usr/share/zsh/vendor-completions/_fdfind"
```


I am not sure if there is a better way to do this.

tmccombs:
I wonder if it would be better for the fdfind completions to call the fd completions, rather than duplicating them

skane-lukas:
I agree that would be ideal.

For bash the fdfind could look like:
```bash
# Source fd completions
source /usr/share/bash-completion/completions/fd

# Complete with same options as fd
eval "$(complete -p fd | sed 's/ fd$/ fdfind/')"
# Or
#if [[ "${BASH_VERSINFO[0]}" -eq 4 && "${BASH_VERSINFO[1]}" -ge 4 || "${BASH_VERSINFO[0]}" -gt 4 ]]; then
#    complete -F _fd -o nosort -o bashdefault -o default fdfind
#else
#    complete -F _fd -o bashdefault -o default fdfind
#fi
```
Bash completions are evaluated lazily so a side efffect  would be that the completions for `fd` would be registered when completions for `fdfind` are triggered in the terminal.


similar apply to zsh
```zsh
#compdef fdfind

# Source completions from fd
source /usr/share/zsh/vendor-completions/_fd

# Register completions for fdfind
compdef _fd fdfind
```

For fish the `--wraps` flag could be used and a completion file could look like:
```fish
complete -c fdfind --wraps fd
```
This seem to work without sourcing anything else.

I am not sure about powershell.
Is this a better approach?

Could these completion files be generated in the `create-deb.sh` script or should they be static files in `contrib/completions`?


Another direction could be to append the code snippets for registering the completions in the fd completion files and keeping the symlinks.

tmccombs:
That seems like a better approach to me, although I don't know how to do it for Powershell either. 

I think putting those in contrib/completions would be fine

villarroelgj:
The current implementation looks good directionally, but there's a build gap that needs fixing before merge.

`create-deb.sh` now installs from `autocomplete/fdfind.bash`, `autocomplete/fdfind.fish`, and `autocomplete/_fdfind`, but the `Makefile` was not updated to generate those files. The `autocomplete/` directory is built by `make  it is not checked into the repo.completions` 

You'll need to add Makefile targets similar to:

```makefile
autocomplete/fdfind.bash: contrib/completion/fdfind.bash
knowledge.md $(comp_dir)
knowledge.md cp $< $@

autocomplete/fdfind.fish: contrib/completion/fdfind.fish
knowledge.md $(comp_dir)
knowledge.md cp $< $@

autocomplete/_fdfind: contrib/completion/_fdfind
knowledge.md $(comp_dir)
knowledge.md cp $< $@
```

And add `autocomplete/fdfind.bash autocomplete/fdfind.fish autocomplete/_fdfind` as dependencies of the `completions` target. Without this, the deb build will fail with "no such file."

Otherwise the delegation approach (bash source + complete, zsh `_fd "$@"`, fish `--wraps`) looks clean. Once this is fixed, could @tmccombs do a formal review?

skane-lukas:
I updated the Makefile and tried to "mirror" the cicd to create and install the deb package successfully. I am not sure how to test the proper cicd.

tmccombs (approved):

PR #1890, opened 2026-02-23
Add --filter and --reject options

### LINKED ISSUE #400: Proposal: add `--filter <cmd>` option

veeceey:
Adds two new flags for command-based result filtering:

- `--filter <cmd> ;` - runs a command for each match, keeps the result only if it exits 0
- `--reject <cmd> ;` - inverse of --filter, excludes results where the command exits 0

These use the same placeholder syntax as `--exec` (`{}`, `{/}`, `{//}`, `{.}`, `{/.}`).

**Example usage:**
```bash
# Find files containing "TODO"
fd -t f --filter grep -q TODO {} \;

# Find files that are NOT symlinks
fd --reject test -L {} \;
```

The commands run with stdout/stderr suppressed so they don't pollute the output. Filter/reject runs in the walker threads alongside the other built-in filters (size, time, type, etc.) so it integrates naturally with the existing pipeline.

Closes #400

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -859,7 +859,11 @@ impl clap::FromArgMatches for Exec {
             .map(CommandSet::new)
             .transpose()
             .map_err(|e| clap::Error::raw(ErrorKind::InvalidValue, e))?;
-        Ok(Exec { command, filter_command, reject_command })
+        Ok(Exec {
+            command,
+            filter_command,
+            reject_command,
+        })
     }
 
     fn update_from_arg_matches(&mut self, matches: &ArgMatches) -> clap::error::Result<()> {
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -608,23 +608,17 @@ impl WorkerState {
                 }
 
                 // Apply --filter: only include entries where the command exits 0.
-                if let Some(ref filter_cmd) = config.filter_command {
-                    if !filter_cmd.matches_filter(
-                        entry_path,
-                        config.path_separator.as_deref(),
-                    ) {
-                        return WalkState::Continue;
-                    }
+                if let Some(ref filter_cmd) = config.filter_command
+                    && !filter_cmd.matches_filter(entry_path, config.path_separator.as_deref())
+                {
+                    return WalkState::Continue;
                 }
 
                 // Apply --reject: exclude entries where the command exits 0.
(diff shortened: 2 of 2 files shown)


veeceey:
**Test results** (macOS, local build):

```
cargo test
test result: ok. 103 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

**Manual testing:**

```bash
# Setup: src/exec/ has 3 files: command.rs, job.rs, mod.rs
# Only job.rs and mod.rs contain "CommandSet"

# --filter: keep files containing "CommandSet"
$ fd . src/exec/ -t f --filter grep -q 'CommandSet' \;
src/exec/job.rs
src/exec/mod.rs
# ✓ Correct - command.rs excluded

# --reject: exclude files containing "CommandSet"  
$ fd . src/exec/ -t f --reject grep -q 'CommandSet' \;
src/exec/command.rs
# ✓ Correct - only command.rs (which doesn't have CommandSet)
```

veeceey:
hey, just bumping this — anything else needed on my end?

veeceey:
any thoughts on this? happy to make adjustments

PR #1902, opened 2026-03-07
fix: always warn when pattern contains path separator, not only for existing dirs

kadaliao:
## Problem

Fixes #1873.

`ensure_search_pattern_is_not_a_path` only showed the path-separator warning when the pattern happened to be an existing directory (`Path::new(&opts.pattern).is_dir()`). Patterns containing `/` that do *not* correspond to an existing directory silently produced no results.

**Reproducer (before fix):**
```sh
mkdir -p dir1/dir2
fd dir1/dir2   # ✅ warns (existing dir)
fd r1/dir2     # ❌ silent, no results (non-existing path)
fd some/nonexistent/pattern  # ❌ silent, no results
```

## Fix

Remove the `is_dir()` guard so the warning fires whenever the pattern contains a path separator and `--full-path` is not set.

**Windows caveat:** `\` is simultaneously a path separator and a valid regex escape character (`\d`, `\w`, etc.) on Windows. To avoid false-positive warnings on legitimate regex patterns, the fix only checks for `/` on Windows.

```rust
let pattern_has_separator = if cfg!(windows) {
    opts.pattern.contains('/')
} else {
    opts.pattern.contains(std::path::MAIN_SEPARATOR)
};
```

## Tests

Added 4 unit tests for `ensure_search_pattern_is_not_a_path`:

- Pattern with separator, non-existent path → now errors ✅
- Pattern with separator, existing dir → still errors ✅
- Pattern without separator → `Ok(())` ✅
- `--full-path` flag suppresses warning → `Ok(())` ✅

All 242 tests pass (`cargo test`).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,10 +147,15 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
 
 /// Detect if the user accidentally supplied a path instead of a search pattern
 fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
-    if !opts.full_path
-        && opts.pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(&opts.pattern).is_dir()
-    {
+    // On Windows, `\` is both a path separator and a regex escape character (e.g. `\d`),
+    // so we only check for forward slashes there to avoid false positives.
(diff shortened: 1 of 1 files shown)


kadaliao:
太久没有回应，先关闭此 PR。

PR #1917, opened 2026-03-09
fix: handle invalid working directories gracefully with --full-path

### LINKED ISSUE #1900: Panic when using `--full-path` if current working directory is invalid or removed
## Description

When `--full-path` (or `-f`) is used, fd builds absolute paths for each match. If the process’s current working directory becomes invalid during the search (e.g. it was deleted or permissions changed), the program can **panic** instead of reporting an error.

**Location:** `src/walk.rs`, in the worker closure that builds `search_str` for full-path matching (around lines 521–524).

**Cause:** `filesystem::path_absolute_form(entry_path)` can return an `Err` (for example when `env::current_dir()` fails). The code uses `.expect("Retrieving absolute path succeeds")`, which turns that error into a panic.

**Reproduction (conceptually):**
1. Start fd with `--full-path` in a directory.
2. From another process/shell, remove or make the current directory inaccessible (e.g. `rmdir` the CWD or revoke access) while the search is running.
3. When fd tries to resolve a relative path, `path_absolute_form` fails and the process panics.

**Suggested fix:** Do not use `.expect()`. Either propagate the error (e.g. with `?`) so the search can fail gracefully, or send a `WorkerResult::Error` and continue, so the user gets an error message and a non-zero exit code instead of a panic.

Xavrir:
Fixes #1900

When using `--full-path`, fd panics if the current working directory becomes invalid (e.g., deleted or inaccessible) during execution. This occurs because the code attempts to resolve relative paths after the cwd is no longer accessible.

This fix replaces the panic with a graceful error message, allowing fd to exit cleanly instead of crashing.

Changes:
- Modified path resolution logic to handle invalid cwd gracefully
- Added regression test to prevent future panics in this scenario
- Verified with `cargo test` and `cargo clippy` locally


src/walk.rs
```diff
@@ -524,18 +531,18 @@ impl WorkerState {
...
-                        ),
+                let search_str = match search_str_for_entry(entry_path, config.search_full_path) {
+                    Ok(search_str) => search_str,
+                    Err(err) => {
+                        let error = ignore::Error::WithPath {
+                            path: entry_path.to_path_buf(),
+                            err: Box::new(ignore::Error::Io(err)),
+                        };
+
+                        return match tx.send(WorkerResult::FatalError(error)) {
```
tmccombs: We don't need the match here. We can just ignore the result and always return `WalkSTate::Quit`

tmccombs (changes requested):
This is a decent attempt, although there are some minor issues. 

however, I'm not sure this is the best approach. One thing I notice is that we are calling `current_dir()` every time we construct an absolute path. 

I wonder if maybe it would be better to re-architect this so that we only call current_dir() once, at the beginning, and if that succeeds, then we can create absolute paths from that infallibly afterwords, and if it fails we can fail early, without needing to check for every entry. And it would probably be more performant as well. 

It would mean that if the current directory disappears while fd is running you will continue to get results that may no longer be valid, which may be good or bad depending on the situation.

Xavrir:
You're right, caching cwd once at startup is the better approach. I reworked the PR: Config now gets a `cwd: Option<PathBuf>` populated when `--full-path` is set, and `search_str_for_entry` uses it infallibly via a new `make_absolute(path, cwd)` function. If the cwd can't be retrieved, fd fails early before the walk begins. This let me drop FatalError entirely, revert the batch() Vec allocation, and delete the test that was mutating process-global state.

src/filesystem.rs
```diff
@@ -20,6 +20,17 @@ pub fn path_absolute_form(path: &Path) -> io::Result<PathBuf> {
...
+/// Construct an absolute path from a potentially relative path and a
+/// pre-resolved working directory. Unlike `path_absolute_form`, this
+/// does not call `env::current_dir()` and cannot fail.
+pub fn make_absolute(path: &Path, cwd: &Path) -> PathBuf {
+    if path.is_absolute() {
+        return path.to_path_buf();
+    }
+    let path = path.strip_prefix(".").unwrap_or(path);
+    cwd.join(path)
+}
```
tmccombs:
```suggestion
pub fn make_absolute<'a, 'b>(path: &'a Path, cwd: &'b Path) -> Cow<'a, Path> {
    if path.is_absolute() {
        return Cow::Borrowed(path);
    }
    let path = path.strip_prefix(".").unwrap_or(path);
    Cow::Owned(cwd.join(path))
}
```

tmccombs: nit: it seems slightly more intuitive to me for the cwd to come first, but that might just be me.

PR #1922, opened 2026-03-12
Clarify that -e/--extension matches all entry types

### LINKED ISSUE #1693: [BUG] `fd -e <ext>` extension search matches `foo.<ext>` in directory names
### Checks

Given the description of the `-e` option I think it could be reasonable to expect only matching *files* with a given extension, instead directories with `.<ext>` suffix are also matched e.g.

```shell
~/temp/bar » fd -e b
a.b/
```

An example of where those kinds of directory names can be found would be package names in some langs e.g. `clojure.zip`.

_No response_

fd 10.2.0

```shell
Darwin 21.5.0 x86_64
```

rexkirshner:
## Summary

Fixes #1693

The help text said "Filter by file extension" which led users to expect only files would be matched. However, directories with matching extensions (e.g. `a.b/`) are also returned, which is correct behavior per UNIX convention.

Updated the help text to "Filter by extension (applies to all entry types)" in:
- CLI help text
- Man page (also added a note about using `--type` to restrict)
- README

## Test plan

- No behavioral change, documentation only
- Wording aligns with collaborator and community feedback on the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ Options:
   -t, --type <filetype>            Filter by type: file (f), directory (d/dir), symlink (l),
                                    executable (x), empty (e), socket (s), pipe (p), char-device
                                    (c), block-device (b)
-  -e, --extension <ext>            Filter by file extension
+  -e, --extension <ext>            Filter by extension (applies to all entry types)
   -S, --size <size>                Limit results based on the size of files
       --changed-within <date|dur>  Filter by file modification time (newer than)
       --changed-before <date|dur>  Filter by file modification time (older than)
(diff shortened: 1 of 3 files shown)


tmccombs (changes requested):

wyf027:
Opened an alternative that incorporates @tmccombs's review (short `-h` text kept minimal; details in long help + man page only). Happy to close this in favor of #1922 if you prefer to update here instead.

wyf027:
For convenience: [#2001](https://github.com/sharkdp/fd/pull/2001) now matches @tmccombs's review feedback (minimal `-h` text, man/long-help details only) and no longer includes unrelated `--full-path` edits (#1997).

PR #1926, opened 2026-03-16
fix: normalize exclude patterns in build_overrides 

### LINKED ISSUE #1919: [BUG] `--exclude` quirk, only seems to affect first `--search-path` in some cases (without glob)
### Checks

### Bug

I believe this is a bug. It's probably at least tangentially related to #851, #374, and #1577.

When multiple `--search-path` arguments are supplied, fd appears to ignore (aka not respect) the `--exclude` directive when processing dirs after the first search path.

### Repro

I wrote a very simple bash script to repro this:

```
#!/usr/bin/env bash

fd --version

echo "only 'myfile1.txt' should be found"
echo

# dir1
mkdir -p ~/test/abc/def
mkdir -p ~/test/ghi/def
touch ~/test/abc/def/myfile1.txt
touch ~/test/ghi/def/myfile2.txt

# dir2
mkdir -p ~/test2/abc/def
mkdir -p ~/test2/ghi/def
touch ~/test2/abc/def/myfile1.txt
touch ~/test2/ghi/def/myfile2.txt

echo "this works: --exclude 'ghi'"
fd --type file --search-path ~/test --search-path ~/test2 --exclude 'ghi'
echo

echo "this also works: --exclude '**/ghi/def'"
fd --type file --search-path ~/test --search-path ~/test2 --exclude '**/ghi/def'
echo

echo "this seems to NOT work as expected: --exclude 'ghi/def'"
echo "(~/test2/ghi/def/myfile2.txt will be shown)"
fd --type file --search-path ~/test --search-path ~/test2 --exclude 'ghi/def'
```

### Output

```
$ ./fd_repro.sh
fd 10.4.2
only 'myfile1.txt' should be found

this works: --exclude 'ghi'
/Users/luke/test/abc/def/myfile1.txt
/Users/luke/test2/abc/def/myfile1.txt

this also works: --exclude '**/ghi/def'
/Users/luke/test/abc/def/myfile1.txt
/Users/luke/test2/abc/def/myfile1.txt

this seems to NOT work as expected: --exclude 'ghi/def'
(~/test2/ghi/def/myfile2.txt will be shown)
/Users/luke/test/abc/def/myfile1.txt
/Users/luke/test2/abc/def/myfile1.txt
/Users/luke/test2/ghi/def/myfile2.txt
```

Did not expect any of the commands to surface `/Users/luke/test2/ghi/def/myfile2.txt`

```
fd 10.4.2
```

```
macOS 26.3.2 (25D2140)
```

FidelSch:
Fixes #1919

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -330,12 +330,20 @@ impl WorkerState {
     fn build_overrides(&self, paths: &[PathBuf]) -> Result<Override> {
         let first_path = &paths[0];
         let config = &self.config;
-
         let mut builder = OverrideBuilder::new(first_path);
 
         for pattern in &config.exclude_patterns {
+            // pattern is "!<glob>". If the glob contains "/" the ignore crate
+            // would anchor it to first_path, which is wrong for --exclude.
+            // Always make such patterns unanchored with a "**/" prefix.
+            let glob = &pattern[1..]; // strip leading "!"
+            let normalized = if glob.contains('/') && !glob.starts_with("**/") {
+                format!("!**/{}", glob)
+            } else {
+                pattern.clone()
+            };
             builder
-                .add(pattern)
+                .add(&normalized)
                 .map_err(|e| anyhow!("Malformed exclude pattern: {}", e))?;
         }
 


src/walk.rs
```diff
@@ -330,12 +330,20 @@ impl WorkerState {
...
-
         let mut builder = OverrideBuilder::new(first_path);
 
         for pattern in &config.exclude_patterns {
+            // pattern is "!<glob>". If the glob contains "/" the ignore crate
+            // would anchor it to first_path, which is wrong for --exclude.
+            // Always make such patterns unanchored with a "**/" prefix.
+            let glob = &pattern[1..]; // strip leading "!"
+            let normalized = if glob.contains('/') && !glob.starts_with("**/") {
+                format!("!**/{}", glob)
```
tmccombs: This will significantly change the behavior. That means that if you exclude something like "/a/b" that will now start also excluding

"foo/a/b"  from the base directory. 

I'll put some more thoughts on how we could approach solving this on the issue itself, but I don't thik this is a good solution for this.

PR #1930, opened 2026-03-18
fix: always error on patterns containing path separators

qianchongyang:
## Problem

`fd` only shows the "pattern contains path separator" error when the pattern also happens to be an existing directory. In all other cases it silently returns no results, which is inconsistent and confusing.

## Solution

Remove the directory existence check so any pattern containing a path separator triggers the same guidance unless `--full-path` is used.

## Validation

- Reviewed the existing path validation branch
- Verified the patch is a minimal logic change in `src/main.rs`
- Full `cargo build/test` could not be run on the Oracle host because the Rust toolchain is not installed there

Fixes #1873


tmccombs:
This doesn't address https://github.com/sharkdp/fd/issues/1873#issuecomment-4001391637.

PR #1943, opened 2026-03-25
build: Create archive script

tmccombs:
Move the code for creating the archive from the GHA yaml into a dedicated script file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,16 @@
-PROFILE=release
+PROFILE?=release
 EXE=target/$(PROFILE)/fd
 prefix=/usr/local
 bindir=$(prefix)/bin
 datadir=$(prefix)/share
 exe_name=fd
+ifdef VERSION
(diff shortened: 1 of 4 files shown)


sharkdp (approved):

Makefile
```diff
@@ -1,9 +1,16 @@
...
 prefix=/usr/local
 bindir=$(prefix)/bin
 datadir=$(prefix)/share
 exe_name=fd
+ifdef VERSION
+	ARCHIVE_NAME = fd-v$(VERSION)
+else
+	ARCHIVE_NAME = fd
+endif
+export ARCHIVE_NAME
```
tavianator, line 12: Do you have to export this variable?

tavianator (approved):

Makefile
tmccombs: Yes. it needs to be exported in order for create-archive.sh to be able to see it.

PR #1946, opened 2026-03-28
chore: Update os matrix

### LINKED ISSUE #1918: x86_64-apple-darwin.tar.gz binary missing since v10.4.1
I noticed that the x86_64-apple-darwin.tar.gz binary has been missing from the release assets since v10.4.1.

Previously (up to v10.3.0), binaries for both architectures were available.

Is the removal of the Intel Mac (x86_64) pre-built binary intentional starting from v10.4.1?

### LINKED ISSUE #1945: Please release a binary for Darwin amd64
please cargo compile for `x86_64-apple-darwin` and release binary, thanks

this is for older intel macs

tmccombs:
Update the versions of Windows and Mac OSes used in build matrix.

Also add back intel Mac since on the latest version that is still supported until 2027.

Fixes: #1945
Fixes: #1918


gjoseph:
@tmccombs would you consider releasing a 10.4.3 for the sake of re-adding these binaries? 🥺

PR #1962, opened 2026-04-12
fix: avoid skipping broken symlinks with --min-depth and --follow

cuiweixie:
## Summary

- **Bug:** With `--min-depth` and `--follow` (`-L`), dangling symlinks were never listed. `DirEntry::broken_symlink` has `depth() == None`; `Option::is_none_or` treats `None` as matching the predicate, so every broken symlink was incorrectly skipped at the min-depth gate.
- **Fix:** Use `is_some_and` so we only skip when depth is known and below `min_depth`.

## Test

- Integration test `test_min_depth_follow_broken_symlink_nested` reproduces `fd --min-depth 1 -L . testdir` with `testdir/sub/broken` → `/noexistent`.

## Changelog

- Entry added under *Unreleased* → *Bugfixes* per CONTRIBUTING.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
+- Do not skip dangling symlinks when using `--min-depth` together with `--follow` (`-L`): broken-symlink entries have no walk depth and were incorrectly filtered. see #1962 (@cuiweixie).
 
 # 10.4.2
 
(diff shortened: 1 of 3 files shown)


tmccombs:
Doesn't this mean that any broken symlinks with a depth less than min-depth will still be printed?

cuiweixie:
> Doesn't this mean that any broken symlinks with a depth less than min-depth will still be printed?

It means any use of --min-depth silently filters out all broken symlinks, regardless of their actual depth in the directory tree.

tmccombs:
I mean that with your fix `is_some_and(...)` will be false if the depths is `None`,  which means that any broken symlink found will be included regardless of the depths. 

So, with your changes if you have a directory tree like:
```
❯ tree
.
├── a
│   └── foo
└── b -> /noexist
```

Then run `fd --min-depth 2` then `b` will be printed, even though it is at depth 1. I'm not sure that's an improvement.

PR #1968, opened 2026-04-16
feat: add --ignore-parent flag to undo --no-ignore-parent

### LINKED ISSUE #1958: Add --ignore-parent option
Add an `--ignore-parent` option to undo the `--no-ignore-parent` option, similar to how `--ignore-vcs` undoes `--no-ignore-vcs`.

Ai-chan-0411:
I noticed that `--no-ignore-vcs` has a corresponding `--ignore-vcs` override, and `--no-ignore` has `--ignore`, but `--no-ignore-parent` had no counterpart. This makes it awkward to undo `--no-ignore-parent` in shell aliases or config files where earlier flags need to be selectively reverted.

This adds a hidden `--ignore-parent` flag following the exact same pattern already used by `--ignore-vcs` and `--ignore`: it uses `overrides_with` to cancel `--no-ignore-parent` when both are present, and it's hidden from help output since it's only needed as a negation toggle.

The change is two lines of attribute macro plus the unit field — no runtime logic changes since clap handles the override resolution automatically.

Closes #1958

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Features
+- Add `--ignore-parent` option to undo `--no-ignore-parent`, see #1958 (@Ai-chan-0411)
+
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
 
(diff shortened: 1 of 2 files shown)


tmccombs:
Duplicate of existing pr

PR #1970, opened 2026-04-17
Fix fdfind bash completion symlink in Debian packages

### LINKED ISSUE #1888: [BUG] Bash completions do not work for alias fdfind in deb release

lawrence3699:
Fixes #1888.

Before this change, `scripts/create-deb.sh` installed the bash completion file as `usr/share/bash-completion/completions/fd` but created the `fdfind` symlink as `./fd.bash`, which leaves a dangling symlink in the Debian package.

This updates the bash completion symlink to `./fd`, matching the installed file name. The fish and zsh `fdfind` symlinks already point to their installed completion file names, so this makes the bash packaging consistent with the sibling paths.

Validation:
- `bash -n scripts/create-deb.sh`
- `git diff --check`
- local packaging-layout fixture confirming `fdfind -> ./fd` resolves alongside the existing fish and zsh completion symlinks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
+- Fix the `fdfind` bash completion symlink in Debian packages, see #1888 (@lawrence3699).
 
 # 10.4.2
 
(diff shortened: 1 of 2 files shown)


tmccombs:
This was fixed by #1889

PR #1973, opened 2026-04-17
fix: show path-separator error for any pattern containing a separator

SAY-5:
Fixes #1873.

## Problem

\`fd pattern\` with a path-separator character in the pattern silently returns zero results most of the time. The user only sees the helpful \"pattern contains a path-separation character\" error when the pattern also happens to match an existing directory on disk:

```
$ fd /        # pattern happens to be a real directory → error fires
$ fd 'src/foo'  # not an existing directory → silent zero results
```

That's because \`ensure_search_pattern_is_not_a_path\` in \`src/main.rs\` checks both:

```rust
if !opts.full_path
    && opts.pattern.contains(std::path::MAIN_SEPARATOR)
    && Path::new(&opts.pattern).is_dir()
```

## Fix

Drop the \`is_dir()\` check. By fd's design, the search is file-name based unless \`--full-path\` is set, so a pattern containing a separator will never match anything. The error should always fire.

```diff
-    if !opts.full_path
-        && opts.pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(&opts.pattern).is_dir()
-    {
+    if !opts.full_path && opts.pattern.contains(std::path::MAIN_SEPARATOR) {
```

The existing error message already generalizes: it suggests \`fd . 'pattern'\` or \`fd --full-path 'pattern'\`, both of which make sense whether the pattern is a real directory or not.

## Behavior change

Before: silent zero results for patterns with separators unless the pattern happens to be a real directory.
After: consistent helpful error for any such pattern, pointing the user to \`--full-path\` or a corrected match-all invocation.

This is the change the issue author explicitly asked for (\"I would expect the error about how a pattern with a '/' in it will not lead to any search results to be displayed regardless of the pattern possibly being a directory path\").

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -145,12 +145,15 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
     Ok(())
 }
 
-/// Detect if the user accidentally supplied a path instead of a search pattern
+/// Detect if the user accidentally supplied a path instead of a search pattern.
+///
+/// A pattern containing a path-separation character will never match because
+/// the regex engine sees the separator literally and fd's search is file-name
+/// based (not full-path based). Fire the helpful error whenever the pattern
+/// contains a separator, regardless of whether it also happens to match an
+/// existing directory on disk. See https://github.com/sharkdp/fd/issues/1873.
 fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
-    if !opts.full_path
-        && opts.pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(&opts.pattern).is_dir()
-    {
+    if !opts.full_path && opts.pattern.contains(std::path::MAIN_SEPARATOR) {
         Err(anyhow!(
             "The search pattern '{pattern}' contains a path-separation character ('{sep}') \
              and will not lead to any search results.\n\n\
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,22 +147,28 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
 
 /// Detect if the user accidentally supplied a path instead of a search pattern.
 ///
-/// A pattern containing a path-separation character will never match because
-/// the regex engine sees the separator literally and fd's search is file-name
-/// based (not full-path based). Fire the helpful error whenever the pattern
-/// contains a separator, regardless of whether it also happens to match an
-/// existing directory on disk. See https://github.com/sharkdp/fd/issues/1873.
+/// A pattern containing '/' will never match because the regex engine sees
+/// the separator literally and fd's search is file-name based (not full-path
+/// based). Fire the helpful error whenever the pattern contains '/',
+/// regardless of whether it also happens to match an existing directory on
+/// disk. See https://github.com/sharkdp/fd/issues/1873.
+///
+/// We intentionally only flag '/' (not the platform MAIN_SEPARATOR): on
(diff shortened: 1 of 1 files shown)


SAY-5:
Pushed a follow-up ([8c9f814](https://github.com/sharkdp/fd/pull/1973/commits/8c9f814e10f5e048500a2069b7f416408a9bd7b8)) that restricts the path-separator check to '/'.

The CI failure on windows-\*-msvc / windows-gnu was caused by me using `std::path::MAIN_SEPARATOR`, which is '\\' on Windows. That collides with regex escape characters — `test_smart_case` uses `\\Ac` as a regex anchor pattern, and my check was incorrectly rejecting it. The original issue #1873 is specifically about '/', which is always a path separator (including on Windows) and has no regex meaning, so checking only '/' fixes the bug on every platform without breaking Windows regex patterns.

SAY-5:
Hi @tmccombs — I noticed this was closed without comment. Was the approach wrong, or is there something I can adjust? Happy to rework if you'd prefer a different approach to fixing #1873. Apologies if the initial breaking of Windows CI caused frustration — the follow-up (8c9f814) limits the check to '/' only, which sidesteps the regex-escape collision entirely.

tmccombs:
There have been many PRs created for this already that do basically the same thing, and from the style of the comments, I'm suspicious this is AI generated.

SAY-5:
I am doing it manually, sorry if it felt that way @tmccombs

SAY-5:
Understood, @tmccombs — thanks for the context on the existing PRs. I'll leave it at that. Appreciate you looking.

tmccombs:
If you want to make this mergeable, the issues with it are:

1. On windows, it will no longer print the warning if you use a pattern that is an actual directory. Maybe we're ok with losing that, but it warrants some discussion, and my initial thought is that is behavior we probably want to keep. 
2. We should add tests for the new behavior
3. You'll need an entry in the changelog, although you can wait to do that until it looks like the rest is ready to be merged.

SAY-5:
Hi @tmccombs — followed up on your three points. Force-pushed onto the same branch:

1. Preserved the Windows `is_dir()` behaviour. The new check is:

   ```rust
   let has_forward_slash = pattern.contains('/');
   let has_windows_dir_backslash =
       cfg!(windows) && pattern.contains('\\') && Path::new(pattern).is_dir();
   if has_forward_slash || has_windows_dir_backslash { ... }
   ```

   So `/` always warns (the original issue #1873), and on Windows a backslash pattern that also resolves to a real directory still warns the way it did before. Regex patterns like `\Ac` or `\d+` no longer hit the check since `/` is not present and the pattern is not a directory.

2. Added three integration tests in `tests/tests.rs`:
   - `test_pattern_forward_slash_always_errors` - asserts the warning fires for `src/foo` and `/`.
   - `test_pattern_full_path_allows_separator` - `--full-path 'one/two'` succeeds.
   - `test_pattern_regex_anchor_not_flagged` - `\Ac` runs as a regex anchor without triggering the warning.

3. Added a CHANGELOG entry under Unreleased / Bugfixes.

Would appreciate a reopen so CI can run on the updated commit. Happy to further adjust the Windows check (e.g. also include forward-slash-free but backslash-and-path patterns) if you'd prefer different coverage there.

tmccombs:
Thanks you. Looks like I can't reopen because there was a force push, but if you open a new PR, I can look at it. Sorry.

PR #1975, opened 2026-04-17
fix: flag any pattern containing a path separator, not just ones naming a directory

SAY-5:
Reopening per @tmccombs ("if you open a new PR, I can look at it" — the
previous PR #1973 force-push blocked reopening). Addresses the three
review items raised there:

1. **Preserve the Windows "actual directory" diagnostic.** The old check
   fired only when the pattern contained `MAIN_SEPARATOR` AND the pattern
   named an existing directory. This PR now fires in two cases:
   - the pattern contains `/` (safe on every platform because `/` is
     always a path separator and has no regex meaning), and
   - on Windows, the pattern contains the native `\` separator AND names
     a real directory (kept intact from master, since `\` is also the
     regex escape character and we can't treat it as a pure path signal).
2. **Tests added** in `tests/tests.rs`:
   - `test_pattern_with_forward_slash_is_rejected` covers both a
     non-existing path (previously silent, now errors) and an existing
     directory (must still error).
   - `test_pattern_with_forward_slash_allowed_with_full_path` guards
     the `--full-path` escape hatch.
3. **Changelog entry** under `# Unreleased / ## Bugfixes`, referencing
   #1873.

### Context on #1873

Without `--full-path`, fd matches the pattern against the file *name*,
never the full path. A pattern like `src/foo.rs` therefore cannot match
anything regardless of whether `src/foo.rs` exists on disk. The old
check only fired when `Path::new(&opts.pattern).is_dir()` was true, so
the common typo (pasting a path, or using a non-existing directory
prefix) silently returned zero results:

```
$ fd nonexistent/path
$   # zero output, no hint that fd is looking at file names only
```

After this change:

```
$ fd nonexistent/path
[fd error]: The search pattern 'nonexistent/path' contains a path-separation character
            and will not lead to any search results.

            If you want to search for all files inside the 'nonexistent/path' directory, use a match-all pattern:
              fd . 'nonexistent/path'

              fd --full-path 'nonexistent/path'
```

The `--full-path` opt-in is unchanged.

Refs #1873.

cc @tmccombs

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -145,21 +145,56 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
     Ok(())
 }
 
-/// Detect if the user accidentally supplied a path instead of a search pattern
+/// Detect if the user accidentally supplied a path instead of a search pattern.
+///
+/// Without `--full-path`, fd matches patterns against file names, so any pattern
+/// containing a path separator can never match. Two cases are worth a friendly
+/// error rather than silent "no results":
+///
+/// 1. The pattern contains '/'. '/' is always a path separator (including on
+///    Windows) and has no regex meaning, so flagging it is safe and catches the
+///    common Linux/macOS mistake of pasting a full path as the pattern.
+/// 2. On Windows only, the pattern contains the native `\` separator *and*
+///    names an existing directory on disk. We can't treat `\` as a pure
+///    path-separator signal there because it is also the regex escape char,
+///    so valid regex patterns like `\Ac` or `\d+` must still run. Requiring
+///    that the pattern resolves to a real directory avoids those false
+///    positives while preserving the legacy diagnostic for operators who
+///    literally typed a directory path.
+///
+/// See https://github.com/sharkdp/fd/issues/1873.
 fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
-    if !opts.full_path
-        && opts.pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(&opts.pattern).is_dir()
+    if opts.full_path {
+        return Ok(());
+    }
+
+    // Start with the cheap check: '/' is always a path separator, including on
+    // Windows, and has no regex meaning, so flagging it is safe and catches the
+    // Linux/macOS mistake of pasting a full path as the pattern.
+    #[cfg_attr(not(windows), allow(unused_mut))]
+    let mut should_warn = opts.pattern.contains('/');
+
+    // On Windows we additionally accept the native `\` separator, but only when
+    // the pattern actually resolves to an existing directory - `\` is also the
+    // regex escape char there, so valid patterns like `\Ac` or `\d+` must still
+    // run. The is_dir syscall is only needed when `should_warn` is still false,
+    // so short-circuit via `||` to avoid the stat call on the happy path.
+    #[cfg(windows)]
     {
+        should_warn = should_warn
+            || (opts.pattern.contains(std::path::MAIN_SEPARATOR)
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -145,21 +145,43 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
...
+    if opts.full_path {
+        return Ok(());
+    }
+
+    let has_forward_slash = opts.pattern.contains('/');
+    let looks_like_existing_windows_dir = cfg!(windows)
         && opts.pattern.contains(std::path::MAIN_SEPARATOR)
-        && Path::new(&opts.pattern).is_dir()
-    {
+        && Path::new(&opts.pattern).is_dir();
```
tmccombs, line 174: Minor performance note. If `has_forward_slash` is true, it would probably be a little more efficient to avoid checking if the path is a directory, since we already know it will be an error.

tests/tests.rs
```diff
@@ -374,6 +374,38 @@ fn test_multi_file_with_missing() {
...
+/// pattern happened to resolve to a real directory, so the common typo of
+/// pasting a full path silently returned zero matches.
+#[test]
+fn test_pattern_with_forward_slash_is_rejected() {
+    let te = TestEnv::new(DEFAULT_DIRS, DEFAULT_FILES);
+
+    // Pattern that is NOT a real directory; old behaviour: no warning.
+    te.assert_failure_with_error(
+        &["nonexistent/path"],
+        "[fd error]: The search pattern 'nonexistent/path' contains a path-separation character",
```
tmccombs, line 389:
```suggestion
        "[fd error]: The search pattern 'nonexistent/path' contains a path-separation character and will not lead to any search results.",
```

I think why there are failed tests is because the way `assert_failure_with_error` work is it (possibly unintentionally) adds a "\n" at the end of the expected string, which doesn't match if you don't have the full line. 

Although I'm confused why the test is currently succeeding on some platforms.

tmccombs (changes requested):
This seems pretty good to me. Just need to fix the failing tests. 

Although I wouldn't mind a second opinion on if this is a good way of addressing the problem.

src/main.rs
```diff
@@ -145,21 +145,43 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
...
+
+    if has_forward_slash || looks_like_existing_windows_dir {
```
tmccombs, line 176: Could be:

```suggestion
    #[cfg_attr(not(windows), allow(unused_mut))]
    let mut  should_warn = opts.pattern.contains('/');
    
    #[cfg(windows)]
    should_warn = should_warn || (
      opts.pattern.contains(std::path::MAIN_SEPARATOR)
      && Path::new(&opts.pattern).is_dir());

    if should_warn{
```

tmccombs (approved):

src/main.rs
```diff
@@ -145,21 +145,56 @@ fn set_working_dir(opts: &Opts) -> Result<()> {
...
+
+    // Start with the cheap check: '/' is always a path separator, including on
+    // Windows, and has no regex meaning, so flagging it is safe and catches the
+    // Linux/macOS mistake of pasting a full path as the pattern.
+    #[cfg_attr(not(windows), allow(unused_mut))]
+    let mut should_warn = opts.pattern.contains('/');
+
+    // On Windows we additionally accept the native `\` separator, but only when
+    // the pattern actually resolves to an existing directory - `\` is also the
+    // regex escape char there, so valid patterns like `\Ac` or `\d+` must still
```
tavianator, line 179: I feel like we should be able to use `regex-syntax` or something to distinguish between escape sequences and literal backslashes, but this is a good start

tavianator (approved):

PR #1976, opened 2026-04-18
fix: sanitize escape sequence injection

curious-rabbit:
- Filenames can contain terminal escape codes (like ESC). fd prints them raw to your terminal, so a file named `innocent\x1b]52;c;...\x1b\.txt `silently rewrites your clipboard. Fix: strip control bytes before writing to a TTY.

- `fd -x rm {/} `in a dir with a file named -rf runs `rm -rf`. Fix: prepend ./ when a path placeholder produces a leading dash.

- `fd -x '{}'` lets you use {} as the command to run, not just an argument to one. It's inconsistent with fd -X which rejects this. The patch rejects placeholder in args[0] for both modes.

Also added tests to ensure changes don't break this in the future

### DIFF
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::cell::OnceCell;
 use std::ffi::OsString;
 use std::fs::{FileType, Metadata};
@@ -54,18 +55,26 @@ impl DirEntry {
     }
 
     /// Returns the path as it should be presented to the user.
-    pub fn stripped_path(&self, config: &Config) -> &Path {
+    /// When stripping `./` would leave the path starting with `-`, keep the `./` so
+    /// downstream tools don't interpret the filename as an option.
+    pub fn stripped_path(&self, config: &Config) -> Cow<'_, Path> {
+        let path = self.path();
         if config.strip_cwd_prefix {
-            strip_current_dir(self.path())
+            let stripped = strip_current_dir(path);
+            if starts_with_dash(stripped) {
+                Cow::Owned(Path::new(".").join(stripped))
+            } else {
+                Cow::Borrowed(stripped)
+            }
         } else {
-            self.path()
+            Cow::Borrowed(path)
         }
     }
 
     /// Returns the path as it should be presented to the user.
     pub fn into_stripped_path(self, config: &Config) -> PathBuf {
         if config.strip_cwd_prefix {
-            self.stripped_path(config).to_path_buf()
+            self.stripped_path(config).into_owned()
         } else {
             self.into_path()
         }
@@ -99,6 +108,10 @@ impl DirEntry {
             .get_or_init(|| ls_colors.style_for(self).cloned())
             .as_ref()
     }
+}
+
+fn starts_with_dash(path: &Path) -> bool {
+    path.as_os_str().as_encoded_bytes().first() == Some(&b'-')
 }
 
 impl PartialEq for DirEntry {
@@ -151,5 +164,27 @@ impl Colorable for DirEntry {
 
     fn metadata(&self) -> Option<Metadata> {
         self.metadata().cloned()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::starts_with_dash;
+    use std::path::Path;
+
+    #[test]
+    fn dash_prefixed_paths_detected() {
+        assert!(starts_with_dash(Path::new("-rf")));
+        assert!(starts_with_dash(Path::new("--delete")));
+        assert!(starts_with_dash(Path::new("-")));
+    }
+
+    #[test]
+    fn safe_paths_not_flagged() {
+        assert!(!starts_with_dash(Path::new("foo")));
+        assert!(!starts_with_dash(Path::new("./foo")));
+        assert!(!starts_with_dash(Path::new("sub/-rf")));
+        assert!(!starts_with_dash(Path::new("")));
+        assert!(!starts_with_dash(Path::new(" -rf")));
     }
 }
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -55,14 +55,14 @@ impl DirEntry {
     }
 
     /// Returns the path as it should be presented to the user.
-    /// When stripping `./` would leave the path starting with `-`, keep the `./` so
-    /// downstream tools don't interpret the filename as an option.
+    /// When stripping `./` would leave the path starting with `-`, keep the original
+    /// (with `./`) so downstream tools don't interpret it as an option.
     pub fn stripped_path(&self, config: &Config) -> Cow<'_, Path> {
         let path = self.path();
         if config.strip_cwd_prefix {
             let stripped = strip_current_dir(path);
             if starts_with_dash(stripped) {
-                Cow::Owned(Path::new(".").join(stripped))
+                Cow::Borrowed(path)
             } else {
                 Cow::Borrowed(stripped)
             }
--- a/src/dir_entry.rs
+++ b/src/dir_entry.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::cell::OnceCell;
 use std::ffi::OsString;
 use std::fs::{FileType, Metadata};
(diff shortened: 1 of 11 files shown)


src/fmt/mod.rs
```diff
@@ -134,6 +165,16 @@ impl FormatTemplate {
                         Text(string) => s.push(string),
                     }
                 }
+
+                if protect_leading_dash
+                    && tokens.first().is_some_and(|t| t.is_path_derived())
+                    && s.as_encoded_bytes().first() == Some(&b'-')
+                {
+                    let mut prefixed = OsString::from("./");
```
tmccombs: This is almost certainly wrong if it is `Basename` or `BasenameNoExt`, since it won't have the rest of the path.

src/main.rs
```diff
@@ -65,7 +66,13 @@ fn main() {
             exit_code.exit();
         }
         Err(err) => {
-            eprintln!("[fd error]: {err:#}");
+            // Gap fix: anyhow errors bubbled from run() may embed user paths/args.
```
tmccombs: Could we just use the `print_error` function here?

src/sanitize.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/// Replace control characters with `?` (matches `ls -q`). Borrows on the fast path.
+pub fn sanitize_for_terminal(s: &str) -> Cow<'_, str> {
+    if !s.chars().any(is_dangerous_control) {
+        return Cow::Borrowed(s);
+    }
+    Cow::Owned(
+        s.chars()
+            .map(|c| if is_dangerous_control(c) { '?' } else { c })
```
tmccombs: Could we escape them with an escape code instead, so that we provide enough information that the user can unambiguously determine what the actual filename was (and maybe escape "\" as well? although that's kind of problematic on windows... maybe we could use a different escape character that is less likely to show up in paths?)

src/fmt/mod.rs
tmccombs: On windows, with a path separator of "\" shouldn't this be ".\" ?

src/fmt/mod.rs
```diff
@@ -134,6 +165,16 @@ impl FormatTemplate {
                         Text(string) => s.push(string),
                     }
                 }
+
+                if protect_leading_dash
```
tmccombs: I think I might prefer if this was done as part of the substitution above. In part to avoid an additional allocation.

tmccombs (changes requested):

src/dir_entry.rs
```diff
@@ -54,18 +55,26 @@ impl DirEntry {
...
-    pub fn stripped_path(&self, config: &Config) -> &Path {
+    /// When stripping `./` would leave the path starting with `-`, keep the `./` so
+    /// downstream tools don't interpret the filename as an option.
+    pub fn stripped_path(&self, config: &Config) -> Cow<'_, Path> {
+        let path = self.path();
         if config.strip_cwd_prefix {
-            strip_current_dir(self.path())
+            let stripped = strip_current_dir(path);
+            if starts_with_dash(stripped) {
+                Cow::Owned(Path::new(".").join(stripped))
```
tavianator: Why not just `Cow::Owned(path)`?

src/sanitize.rs
```diff
@@ -0,0 +1,101 @@
+//! TTY-output sanitization to prevent terminal escape injection via filenames.
+
+use std::borrow::Cow;
+use std::fmt::Write;
+
+#[inline]
+fn is_dangerous_control(c: char) -> bool {
+    // C0 (except HT), DEL, and C1 controls (U+0080..=U+009F can act as
+    // single-byte CSI/OSC initiators on 8-bit-control terminals).
+    matches!(c, '\x00'..='\x08' | '\x0A'..='\x1F' | '\x7F' | '\u{80}'..='\u{9F}')
```
tavianator: This is okay, but for `bfs` I went even more aggressive and just escaped everything that isn't a printable character or whitespace: https://github.com/tavianator/bfs/blob/9220b27ae0d7e802525e16496317ed891fe7ac35/src/bfstd.c#L1169

src/exec/mod.rs
```diff
@@ -245,6 +242,12 @@ impl CommandTemplate {
             bail!("No executable provided for --exec or --exec-batch");
         }
 
+        if args[0].has_tokens() {
```
tavianator: Why did this move?

curious-rabbit: Moved it to CommandTemplate::new so the same check covers both --exec and --exec-batch. Previously only -X rejected placeholder-as-executable; `fd -x '{}' `would happily try to execute each matched file as a program. The duplicate check in CommandSet::new_batch is removed because CommandTemplate::new is on the path for both constructors.

src/sanitize.rs
curious-rabbit: I tried to apply at least the same threat model that is used by `find` 
Now the filter is even a bit more aggressive but still reasonable.

src/dir_entry.rs
curious-rabbit: Switched to Cow::Borrowed(path) so it's allocation-free and preserves the original bytes/separator verbatim.

src/dir_entry.rs
```diff
@@ -54,18 +55,26 @@ impl DirEntry {
     }
 
     /// Returns the path as it should be presented to the user.
-    pub fn stripped_path(&self, config: &Config) -> &Path {
+    /// When stripping `./` would leave the path starting with `-`, keep the original
+    /// (with `./`) so downstream tools don't interpret it as an option.
+    pub fn stripped_path(&self, config: &Config) -> Cow<'_, Path> {
```
tmccombs: We always return a `Borrowed` now. Could we change this back to returning a `&Path`?

src/error.rs
```diff
@@ -1,3 +1,12 @@
...
+
+use crate::sanitize::sanitize_for_terminal;
+
 pub fn print_error(msg: impl Into<String>) {
-    eprintln!("[fd error]: {}", msg.into());
+    let msg = msg.into();
+    if std::io::stderr().is_terminal() {
+        eprintln!("[fd error]: {}", sanitize_for_terminal(&msg));
+    } else {
+        eprintln!("[fd error]: {msg}");
```
tmccombs: Could we make use of `maybe_sanitize` here?

src/sanitize.rs
```diff
@@ -0,0 +1,147 @@
...
+            | '\u{202A}'..='\u{202E}'   // bidi embedding/override
+            | '\u{2060}'..='\u{206F}'   // word joiner, invisibles, deprecated formats
+            | '\u{FEFF}'                // BOM / zero-width no-break space
+            | '\u{FFF9}'..='\u{FFFB}'   // interlinear annotation
+            | '\u{E0000}'..='\u{E007F}' // language tags
+        )
+}
+
+/// Replace dangerous chars with `\xNN` / `\u{NNNN}` so the original is recoverable.
+pub fn sanitize_for_terminal(s: &str) -> Cow<'_, str> {
```
tmccombs, line 28: I'm not sure if it is worth the additional complexity, but it would be possible to use a struct that implements `Display` that passes characters or escapes through as it processes them, without needing to allocate an intermediate String.

src/sanitize.rs
tmccombs, line 1: Have you done any benchmarking to see what impact, if any, this has on performance?

tmccombs (approved):
Would you mind adding an entry to the changelog?

src/sanitize.rs
curious-rabbit: I tried it but it made the parsing slower and I skipped it.

src/sanitize.rs
curious-rabbit: I benchmarked this current PR vs the unpatched version and could not find any measurable difference

tmccombs (approved):

src/exec/mod.rs
tavianator: If `fd -x {}` currently works, maybe it should keep working.  `fd -X {}` is unpredictable and almost always nonsense, but I can imagine something like

```console
$ fd some-script.sh -x {} arg arg
```

where I want to run that script wherever I put it

curious-rabbit: Done

curious-rabbit:
Is there anything left to do for the merge?

decathorpe:
It looks like this PR removed jemalloc from the default features. I don't see any references to this happening so it appears to have been accidental? Note that this has leaked into the 10.5.0 release too.

PR #1979, opened 2026-04-20
docs: document PowerShell placeholder quoting requirement

### LINKED ISSUE #735: Unexpected behavior for the Placeholder syntax on windows powershell
```powershell
$ fd --version
fd 8.2.1
```

**Context**
I want to list all files recursively in a directory while printing only the basename.

So instead of this:
```powershell
dir1\test1.txt
dir1\test2.txt
dir2\testA.txt
```
I want this:
```powershell
test1.txt
test2.txt
testA.txt
```
To my understanding, the best way to do so is to use `fd -t f -x echo {/}`.

On windows `cmd.exe` it works fine (but it's very slow).
```cmd
fd -t f -x echo {/}
test1.txt
test2.txt
testA.txt
```

On windows `powershell.exe`, the output is not as expected.
```
$ fd -t f -x echo {/}
-encodedCommand LwA= -inputFormat xml -outputFormat text dir1\test1.txt
-encodedCommand LwA= -inputFormat xml -outputFormat text dir1\test2.txt
-encodedCommand LwA= -inputFormat xml -outputFormat text dir2\testA.txt
```

**Question**
Is this a bug?
Maybe a documentation problem?
Is there a better way than `fd -t f -x echo {/}`  ?

YoshKoz:
Closes #735.

In #735 the maintainer asked for the troubleshooting section to mention that `{}` placeholders need to be escaped in PowerShell. Adds:

- A new "Placeholders (`{}`, `{/}`, `{//}`, `{.}`, `{/.}`) in PowerShell" subsection under **Troubleshooting** with a working/non-working example.
- A matching changelog entry.

No code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,6 @@
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
 
-## Documentation
-- Document PowerShell placeholder quoting requirement in the troubleshooting section, see #735 (@YoshKoz)
-
 # 10.4.2
 
 ## Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
 
+## Documentation
+- Document PowerShell placeholder quoting requirement in the troubleshooting section, see #735 (@YoshKoz)
+
 # 10.4.2
 
 ## Bugfixes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
tmccombs, line 1: I don't think we need a changelog entry for changes to readme

README.md
```diff
@@ -444,6 +444,23 @@ you can use `export -f my_function` to make available to child processes. You wo
 need to call `fd -x bash -c 'my_function "$1"' bash`. For other use cases or shells, use
 a (temporary) shell script.
 
+### Placeholders (`{}`, `{/}`, `{//}`, `{.}`, `{/.}`) in PowerShell
```
tmccombs: Could we just say something like "depending on your shell you may need to quote the placeholder"?

I don't think Powershell is the only shell where this is required, and I don't want the readme to have to teach people how to use their shell.

tmccombs (changes requested):

tmccombs (approved):

PR #1982, opened 2026-04-20
feat: Add --sort arg to CLI to sort by path/size/dates

### LINKED ISSUE #1875: Add a `--sort <path|size|created|modified>` option, especially for use with `-x <cmd>`
I want to run a command for a bunch of files, but it's important that the output is alphabetical by file name. 

As far as I know, the only way to do this is to switch to using 

I believe the old `sort` command supported this behaviour with a similar technique.

I'm aware `-l` / `--list-details` exists and does a sort, but it is incompatible with `-x`.

DeflateAwning:
Resolves #1875


src/filter/mod.rs
```diff
@@ -1,10 +1,12 @@
 pub use self::size::SizeFilter;
+pub use self::sort::SortKey;
 pub use self::time::TimeFilter;
 
 #[cfg(unix)]
 pub use self::owner::OwnerFilter;
 
 mod size;
+mod sort; // Arguably not a "filter", but more of an augmentation on search results.
```
tmccombs: Yeah, I don't think this is the right place to put this.

src/cli.rs
```diff
@@ -549,6 +549,20 @@ pub struct Opts {
     #[arg(long, hide = true, value_parser = parse_millis)]
     pub max_buffer_time: Option<Duration>,
 
+    /// Sort search results by the given key before printing or executing commands.
+    ///
+    /// Note: this buffers all results in memory before outputting.
```
tmccombs: Perhaps there should be more of a warning that this can result in increased memory usage and prevents some level of parallelization.

We should probably also have documentation on how it interacts with `--exec` and `--exec-batch`. At the very least in the man page, but probably in the help text as well.

tests/tests.rs
```diff
@@ -2023,6 +2023,49 @@ fn test_exec_batch_with_limit() {
...
+
+    files
+}
+
+#[test]
+fn test_sort_by_path() {
+    let te = TestEnv::new(DEFAULT_DIRS, &shuffle_files(DEFAULT_FILES, 42));
+
+    // --sort=path should produce deterministic alphabetical output
+    te.assert_output(
```
tmccombs, line 2053: assert_output is not suitable for testing this, because it is meant to check output for non-deterministic order.

src/walk.rs
```diff
@@ -411,7 +412,18 @@ impl WorkerState {
         // This will be set to `Some` if the `--exec` argument was supplied.
         if let Some(ref cmd) = config.command {
```
tmccombs, line 426: Your changes only actually affect the order when using `--exec` or `--exec-batch`, you aren't changing anything when printing results.

src/walk.rs
```diff
@@ -411,7 +412,18 @@ impl WorkerState {
...
             if cmd.in_batch_mode() {
-                exec::batch(rx.into_iter().flatten(), cmd, config)
+                let mut results: Vec<WorkerResult> = rx.into_iter().flatten().collect();
```
tmccombs: This creates an intermediate buffer of all results, regardless of if we are sorting. That will increase memory usage and possibly execution time, even if sort isn't used.  Yes, in the case of `exec-batch` we do eventually effectively create a `Vec` of all the arguments, but  with this, we are now creating a Vec of all the DirEntries AND a Vec of all the arguments. 

We should avoid this buffering if the user doesn't specify to sort.

src/walk.rs
```diff
@@ -663,6 +675,35 @@ impl WorkerState {
...
+fn sort_worker_results(results: &mut [WorkerResult], sort_key: SortKey) {
+    results.sort_by(|a, b| {
+        // Errors sort to the end; two errors are considered equal
+        let (WorkerResult::Entry(a), WorkerResult::Entry(b)) = (a, b) else {
+            return match (a, b) {
+                (WorkerResult::Error(_), WorkerResult::Entry(_)) => std::cmp::Ordering::Greater,
+                (WorkerResult::Entry(_), WorkerResult::Error(_)) => std::cmp::Ordering::Less,
+                _ => std::cmp::Ordering::Equal,
+            };
+        };
```
tmccombs:
```suggestion
        let (a, b) = match (a, b) {
            (WorkerResult::Entry(a), WorkerResult::Entry(b)) => (a, b) 
            (WorkerResult::Error(_), WorkerResult::Entry(_)) => return std::cmp::Ordering::Greater,
            (WorkerResult::Entry(_), WorkerResult::Error(_)) => return std::cmp::Ordering::Less,
            _ => return std::cmp::Ordering::Equal,
        };

```

Although, it might actually be better to split the sequence into entries and errors _before_ sorting it. 

Maybe just use a `filter` that reports any errors it find along the way first?

src/walk.rs
```diff
@@ -663,6 +675,35 @@ impl WorkerState {
...
+
+        match sort_key {
```
tmccombs: I'm not sure, but it may be more performant if we passed in a different function depending on the value of the sort key,  rather than checking the value of the sort key on every iteration. Probably worth benchmarking that.

src/walk.rs
tmccombs, line 1: This implementation requires sorting to be handled by every output path. That adds complexity and makes it harder to maintain. 

It would be better if there is some way we could avoid needing separate sorting logic for `--exec`, `--exec-batch`, and printing.

tmccombs (changes requested):
If a sorting feature is added to fd, it will need to have very minimal impact on flows that don't use sort, and shouldn't significantly add to the complexity and maintenance burden of the project. 

This isn't there (yet).

src/filter/mod.rs
DeflateAwning: Lol okay, any advice? Hoping to get it right on round 2 instead of trial-and-erroring. 

There is no precedent for enums like this shared perfectly between the CLI Opts and Config structs, as far as I can tell (other than filters).

src/walk.rs
DeflateAwning: Gotcha, my mistake. Was tough being certain about the scope of what was being modified. 

Do you recommend updating the docs to explain that `--sort` only applies to `-x`, or do you recommend fixing it to make `--sort` apply to all (including printing)?

src/walk.rs
DeflateAwning: You're totally right, sorry. Was an oversight. Will fix.

src/walk.rs
DeflateAwning: My hunch is that branch prediction makes this a non-issue (after a branch is taking one way many times, a proper modern branch predictor will guess that it'll continue going that way). 

Fair enough though - a perfect solution would do as you suggest.

DeflateAwning:
Thank you for the review! One or two days and I'll have another version to look over.

src/filter/mod.rs
DeflateAwning: Moved to `config.rs`

src/walk.rs
DeflateAwning: Fixed in ba9cb06

src/walk.rs
DeflateAwning: Improved in 676be36

src/walk.rs
DeflateAwning: Decision: Made it so that `--sort` works on all code paths (i.e., not just `--exec`).

src/cli.rs
DeflateAwning: Improved in 90e6539. I'm open to further suggestions though.

DeflateAwning:
@tmccombs Thanks again for the review/suggestions. Would you mind taking another look? It's getting closer, I think.

Regarding the "one code path" suggestion - I'm open to ideas. I think it's reasonably DRYed right now. The alternative implementation I thought of involves piping the iterator through something like `itertools::sorted`, but I'm not sure I see how to make that happen, and I'm not convinced it'd be much cleaner.

DeflateAwning:
@tmccombs Would you please take another look at this one? I did all of your suggestions, and I'd hate for it to get too out-of-date with the trunk.

ShadiZade:
Any progress on this?

PR #1983, opened 2026-04-23
feat: Add --exact argument to match exact filename

### LINKED ISSUE #1963: Add --exact option
This was already requested by @KuroSensei in #627 back in 2020, specifically because `--fixed-strings` performs substring matching even when undesirable, and there's no way to turn that off. @sharkdp suggested that `--glob` can be used instead and closed the issue, but I say that this is not a good solution because globbing syntax can interfere and cause undesirable behavior.

`--exact` would perform `--fixed-strings` without substring matching.

DeflateAwning:
Resolves #1963.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -173,16 +173,29 @@ pub struct Opts {
 
     /// Treat the pattern as a literal string instead of a regular expression. Note
     /// that this also performs substring comparison. If you want to match on an
-    /// exact filename, consider using '--glob'.
+    /// exact filename, consider using '--glob' or '--exact' instead.
     #[arg(
         long,
         short = 'F',
         alias = "literal",
         hide_short_help = true,
-        help = "Treat pattern as literal string stead of regex",
+        help = "Treat pattern as literal string instead of regex",
         long_help
     )]
     pub fixed_strings: bool,
+
+    /// Perform an exact match. This is equivalent to '--fixed-strings' but requires
+    /// the pattern to match the entire filename (or path if '--full-path' is used),
+    /// rather than a substring. Special regex characters in the pattern are treated
+    /// as literal characters.
+    #[arg(
+        long,
+        conflicts_with("glob"),
+        hide_short_help = true,
+        help = "Match the entire filename exactly (literal, non-substring)",
+        long_help
+    )]
+    pub exact: bool,
 
     /// Add additional required search patterns, all of which must be matched. Multiple
     /// additional patterns can be specified. The patterns are regular
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -190,7 +190,7 @@ pub struct Opts {
     /// as literal characters.
     #[arg(
         long,
-        conflicts_with("glob"),
+        conflicts_with_all(["glob", "fixed_strings"]),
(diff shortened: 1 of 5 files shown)


ShadiZade:
Works as intended and solves a headache I've been having for a long time.
image
It might be worth it to add `--exact` to the regex error message, though.

DeflateAwning:
Just added the regex comment, as suggested.

src/main.rs
```diff
@@ -170,6 +170,10 @@ fn build_pattern_regex(pattern: &str, opts: &Opts) -> Result<String> {
     Ok(if opts.glob && !pattern.is_empty() {
         let glob = GlobBuilder::new(pattern).literal_separator(true).build()?;
         glob.regex().to_owned()
+    } else if opts.exact {
+        // Anchor the escaped pattern so the full filename (or path) must match exactly.
+        // Literal. No substring matching.
+        format!("^{}$", regex::escape(pattern))
```
tmccombs, line 176: it might be a little more performant if we used `regex_syntax::escape_into` to write into the same String we use for the anchors. 
but that would be less readable, and we only do this once, so probably not worth it.

I kind of wish there was something like `regex::escape` that returned something that implemented Display so it could be used in formats like this without having to allocate.

tests/tests.rs
```diff
@@ -2135,14 +2135,49 @@ fn test_fixed_strings() {
...
+    // Combine with --case-sensitive (matched).
+    te.assert_output(
+        &["--fixed-strings", "--case-sensitive", "Download (1)"],
+        "test2/Download (1).tar.gz",
+    );
+}
+
+/// Literal search, non-substring (--exact)
+#[test]
+fn test_exact_literal_nonsubstring() {
```
tmccombs, line 2156: We should also test that it only matches the exact file name and not a substring of it.

That is `a.foo` shouldn't match "aa.foo" , "a.food" or "ca.food"

src/cli.rs
```diff
@@ -173,17 +173,30 @@ pub struct Opts {
...
     )]
     pub fixed_strings: bool,
 
+    /// Perform an exact match. This is equivalent to '--fixed-strings' but requires
+    /// the pattern to match the entire filename (or path if '--full-path' is used),
+    /// rather than a substring. Special regex characters in the pattern are treated
+    /// as literal characters.
+    #[arg(
+        long,
+        conflicts_with("glob"),
```
tmccombs: This should also conflict with fixed_strings.

tmccombs (changes requested):
Overall, this looks pretty good to me, there are just a couple of minor things to fix.

Also, please add an entry to the changelog, and update the man page.

tests/tests.rs
DeflateAwning: Resolved in b77f9b8

src/cli.rs
DeflateAwning: Fixed in d642c44

DeflateAwning:
@tmccombs Fixed those issues, thanks! Sorry for the delay.

tmccombs (approved):

PR #1990, opened 2026-05-04
Fix --min-depth for broken symlinks

### LINKED ISSUE #1017: --min-depth is broken for broken symlinks

mitre88:
Fixes #1017

Broken symlinks were returning None for depth(), causing min-depth checks to fail. This commit tracks depth for BrokenSymlink entries and returns it properly.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -495,7 +495,8 @@ impl WorkerState {
                                     .ok()
                                     .is_some_and(|m| m.file_type().is_symlink()) =>
                         {
-                            DirEntry::broken_symlink(path)
+                            let broken_path = path.clone();
+                            DirEntry::broken_symlink(broken_path, 1)
                         }
                         _ => {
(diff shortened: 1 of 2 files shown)


src/walk.rs
```diff
@@ -495,7 +495,8 @@ impl WorkerState {
                                     .ok()
                                     .is_some_and(|m| m.file_type().is_symlink()) =>
                         {
-                            DirEntry::broken_symlink(path)
+                            let broken_path = path.clone();
+                            DirEntry::broken_symlink(broken_path, 1)
```
tavianator, line 499: What if the depth is not 1?

tavianator (changes requested):

mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (changes-requested) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

PR #1991, opened 2026-05-04
fix: exclude directories from extension filter (issue #1693)

mitre88:
Fixes issue #1693 - directories with matching extension suffix were incorrectly included in results. Now extension filter only applies to files, not directories.


tmccombs:
It looks like this includes changes from https://github.com/sharkdp/fd/pull/1990 as well.

I'm also not sure this is something we actually want. Directories can have extensions too, and limiting it to files is just a matter of using `-t f` in the fd invocation.

PR #1993, opened 2026-05-05
fix: prepend **/ to glob patterns when using --full-path

### LINKED ISSUE #1650: [BUG] Glob pattern that uses ** does not seem to work
### Checks

```
> fd -p foo.txt
bar/baz/foo.txt

> fd -p -g foo.txt
(no results) <-- ??

> fd -p -g '**/foo.txt'
bar/baz/foo.txt

> fd -p -g '/**/foo.txt'
bar/baz/foo.txt

> fd -p -g 'bar/**/foo.txt'
(no results) <-- ??
```

Are the two `(no results)` above supposed to work?

_No response_

fd 10.2.0

```shell
Darwin 23.6.0 arm64
```

argon17:
When using '--glob' with '--full-path' together, **globset** always generates fully-anchored regexes (^...$), so patterns like 'c.foo' or 'one/**/c.foo' never matches against the absolute path, because the regex requires the match to start at the very beginning of the string, but the full path will always start with '/'.

Fixed by prepending **/ to the glob pattern when '--full-path' is set and the pattern doesn't already start with / or **/.

Fixes #1650

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Add `--ignore-parent` option to override `--no-ignore-parent`, see #1958 (@tmchow)
 
 ## Bugfixes
+- Fix `--glob` patterns not working correctly when used with `--full-path`, see #1650 (@argon17)
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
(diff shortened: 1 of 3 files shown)


tavianator (changes requested):
Automatically prepending `**/` is dangerous.  Maybe the user meant something more like `./path/to/file`, but `**/path/to/file` can match more files than expected.

Issue #839, opened 2021-08-29
`--abs-path` and `--rel-path` instead of `--full-path`


andreavaccari:
`fd --full-path` matches a pattern against the *absolute* path instead of the *relative* one. This is not well documented and it's a possible cause of confusion (e.g. #769). In my opinion, it is also less ergonomic than matching against the relative path of a candidate.

My suggestion is:

* Replace `--full-path` with `--abs-path` to make it clear this matches against the absolute path
* Add a new flag `--rel-path` to allow matching against the relative path

Thank you for the great work and for your focus on usability!

tmccombs:
I like that idea. Though we would probably want to keep `--full-path` as a (probably hidden) alias for `--abs-path` for backwards compatibility.

andreavaccari:
Thank you for taking this into consideration! It makes sense to maintain backward compat.

Separately, I realized there is `--absolute-path` to modify how results are printed but not how they are matched. Introducing `--abs-path` next to `--absolute-path` could itself cause confusion. One solution would be to rename the latter to `--show-abs-path`, which would align with `--show-errors`, but would break backward compat. Another solution would be to introduce the longer `--match-abs-path` and `--match-rel-path`.

sharkdp:
> `fd --full-path` matches a pattern against the _absolute_ path instead of the _relative_ one. This is not well documented and it's a possible cause of confusion (e.g. #769). 

Thank you for your request. I am definitely in favor of improving the documentation. It should be clear that we match on the full absolute path.

> In my opinion, it is also less ergonomic than matching against the relative path of a candidate.

Could you explain this in more detail? What is your use case and why is it unergonomic to solve it with `--full-path`?

andreavaccari:
> Could you explain this in more detail? What is your use case and why is it unergonomic to solve it with `--full-path`?

Say you have the following directory structure:
```
~
├── project/
│   ├── docs/
│   │   ├── some-doc.pdf
│   │   ├── some-doc.txt
│   │   └── secret.pdf
│   ├── repo/
│   │   ├── ...
```

Let's say I'm in `project` and I want to find all `.pdf` files in `docs`:
- I could issue `cd docs && fd ".*pdf$"`, but let's say I don't want to change directory.
- Intuitively I'd want to issue `fd --rel-path "^docs/.*pdf$"`
- Instead I'd have to issue `fd --full-path ".*docs/.*pdf$"` but this could also match `docs` dirs under `repo`.
[...]

PR #1996, opened 2026-05-14
docs: clarify full path help text

### LINKED ISSUE #1686: [BUG] The help page is confusing for --full-path option
### Checks

-p, --full-path Search full abs. path (default: filename only)

-p, --full-path Search full path (default: the last element of the path)

NOTES:
1) "full" and "absolute" are the same terms.
2) no reason to say "file or directory name" (like it is done in man page) either.
    "the last element of the path" clearly describes what it does.

10.2.0

```shell
Windows: 26100.3194
```

fizyxbt:
## Summary

- clarify that `--full-path` searches the full path, not an absolute output path
- describe the default search target as the final path component
- update the README help snapshot to match

Fixes #1686.

## Checks

- `cargo fmt -- --check`
- `cargo clippy --all-targets --all-features -- -Dwarnings`
- `cargo test --locked`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -239,12 +239,12 @@ pub struct Opts {
     #[arg(long, overrides_with = "follow", hide = true, action = ArgAction::SetTrue)]
     no_follow: (),
 
-    /// By default, the search pattern is only matched against the filename (or directory name). Using this flag, the pattern is matched against the full (absolute) path. Example:
+    /// By default, the search pattern is only matched against the final path component. Using this flag, the pattern is matched against the full path. Example:
     ///   fd --glob -p '**/.git/config'
     #[arg(
         long,
         short = 'p',
-        help = "Search full abs. path (default: filename only)",
+        help = "Search full path (default: final path component only)",
         long_help,
         verbatim_doc_comment
     )]
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ Options:
   -a, --absolute-path              Show absolute instead of relative paths
   -l, --list-details               Use a long listing format with file metadata
   -L, --follow                     Follow symbolic links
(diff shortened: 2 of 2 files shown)


src/cli.rs
```diff
@@ -239,12 +239,12 @@ pub struct Opts {
     #[arg(long, overrides_with = "follow", hide = true, action = ArgAction::SetTrue)]
     no_follow: (),
 
-    /// By default, the search pattern is only matched against the filename (or directory name). Using this flag, the pattern is matched against the full (absolute) path. Example:
+    /// By default, the search pattern is only matched against the final path component. Using this flag, the pattern is matched against the full path. Example:
```
tmccombs, line 242: I'd be ok with chaning to to say just "absolute path" instead of "full (absolute) path" however,  I do think it is important that we specify that it is absolute. Otherwise "full" is potentially ambiguous, because it could either mean the absolute path, or the full relative path compared to the the relative path (relative to the working directory, or base directory). See also #839.

I also think that "last" is more clear than "final" when referring to the filename.

tmccombs (changes requested):

wyf027:
Friendly note: [#1997](https://github.com/sharkdp/fd/pull/1997) covers the same #1686 help-text updates with a broader pass (README, man, completion, regression test) and is CI-green. Consider closing this in favor of #1997 to reduce maintainer review load.

wyf027:
@fizyxbt Friendly follow-up: @tmccombs left `CHANGES_REQUESTED` here asking to say **absolute path** (not ambiguous "full"). [#1997](https://github.com/sharkdp/fd/pull/1997) applies that wording across README, man, completions, `-h`, plus a regression test — might be simplest to close #1996 and route review to #1997. Happy either way.

PR #1997, opened 2026-05-16
docs: clarify --full-path help text (fixes #1686)

wyf027:
## Summary

Clarifies `--full-path` / `-p` documentation so users understand:

- By default, patterns match the **last path component** (not "filename only" / ambiguous "full path").
- With `-p`, patterns match the **absolute path** (not a relative path from the search root).

Updates `src/cli.rs`, `doc/fd.1`, `README.md` (prose + `--help` snapshot), zsh completion, and an internal comment in `src/config.rs`.

Fixes #1686.

## Context

Follows feedback on #1996 from @tmccombs:

- Keep **absolute** in the description of what `-p` matches (see also #839).
- Prefer **last path component** over "final" / "filename only" for the default behavior.

## Checks

- [x] `cargo fmt -- --check`
- [x] `cargo clippy --all-targets --all-features -- -Dwarnings`
- [x] `cargo test --locked`

PR #1999, opened 2026-05-16
feat: warn when --full-path --glob pattern lacks path anchor (fixes #1650)

### LINKED ISSUE #1650: [BUG] Glob pattern that uses ** does not seem to work

wyf027:
## Summary

With `--glob` and `--full-path`, patterns match the **entire absolute path** (from `/`). Patterns like `foo.txt` or `bar/**/foo.txt` often match nothing because they do not anchor to the path root.

This adds a stderr warning when the pattern does not start with `/` or `*`, as suggested by @tavianator on #1650.

Does not change matching behavior (orthogonal to #1993).

## Test plan

- [x] `cargo test test_warn_when_full_path_glob_missing_leading_anchor`
- [x] `cargo clippy -- -D warnings`

Fixes #1650

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -203,22 +203,29 @@ fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
     }
 }
 
-/// With `--glob` and `--full-path`, patterns match the entire absolute path (from `/`).
-/// Warn when the pattern does not start with `/` or `*`, which often means no matches (#1650).
+/// With `--glob` and `--full-path`, patterns match the entire absolute path.
+/// Warn when the pattern does not start with a path root or glob wildcard, which often means no
+/// matches (#1650).
 fn warn_if_full_path_glob_missing_leading_anchor(opts: &Opts) {
     if !opts.glob || !opts.full_path || opts.pattern.is_empty() {
         return;
     }
-    let Some(first) = opts.pattern.chars().next() else {
-        return;
-    };
-    if first == '/' || first == '*' {
+    if full_path_glob_has_leading_anchor(&opts.pattern) {
         return;
     }
     print_warning(
         "With --glob and --full-path, the pattern matches the whole absolute path and \
          usually should start with '/' or '**/' (for example: '**/foo.txt').",
     );
+}
+
+fn full_path_glob_has_leading_anchor(pattern: &str) -> bool {
+    let bytes = pattern.as_bytes();
+    matches!(bytes.first(), Some(b'*' | b'/' | b'\\'))
+        || matches!(
+            bytes,
+            [drive, b':', b'/' | b'\\', ..] if drive.is_ascii_alphabetic()
+        )
 }
 
 fn build_pattern_regex(pattern: &str, opts: &Opts) -> Result<String> {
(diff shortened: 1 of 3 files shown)


wyf027:
CI fix: removed `test_full_path_help_text` that belonged to #1997 — it was accidentally on this branch without the matching `cli.rs` help change, which broke MSRV.

wyf027:
Friendly ping — CI green after dropping an accidentally included #1997 test. Small warning-only change for #1650. Thanks!

wyf027:
Friendly ping — warns when `--full-path --glob` patterns lack a `/` anchor (#1650). CI green; happy to adjust wording if needed.

src/main.rs
```diff
@@ -201,6 +203,24 @@ fn ensure_search_pattern_is_not_a_path(opts: &Opts) -> Result<()> {
...
+/// With `--glob` and `--full-path`, patterns match the entire absolute path (from `/`).
+/// Warn when the pattern does not start with `/` or `*`, which often means no matches (#1650).
+fn warn_if_full_path_glob_missing_leading_anchor(opts: &Opts) {
+    if !opts.glob || !opts.full_path || opts.pattern.is_empty() {
+        return;
+    }
+    let Some(first) = opts.pattern.chars().next() else {
+        return;
+    };
+    if first == '/' || first == '*' {
```
tmccombs: This is only correct on unix systems. On windows an absolute path could start with a letter (for example it could start with "c:\\").

tmccombs (changes requested):
I think this should be done in a way that works across operating systems. I'm not sure entirely sure what a robust way to determine if a glob pattern is absolute on windows, largely due to my lack of windows expertise, but I'm confident this isn't adequate.

Also, @leno23 it looks like you may be using an AI agent (perhaps OpenClaw or similar?) that is generating a lot of comments and pull requests, including duplicate pull requests. 

Although the use of AI isn't entirely prohibited, we ask that you have a human in the loop reviewing the changes before opening pull requests, and don't automatically add comments.  See our [contribution policy](https://github.com/sharkdp/fd/pull/1999#issuecomment-4468057290). 

In particular, comments like https://github.com/sharkdp/fd/pull/1999#issuecomment-4468057290 and https://github.com/sharkdp/fd/pull/1999#issuecomment-4468057290 and similar ones on related issues are not useful, and create noise and extra work for maintainers. Similarly with duplicate PRs.

wyf027:
Closing this since the review made clear that the current approach is not adequate across platforms, and the PR has not had a clear merge path for more than 14 days. Thanks for the review, and sorry for the extra noise from the earlier automated comments.

PR #2000, opened 2026-05-16
fix: apply --exclude globs under each --search-path root

### LINKED ISSUE #1919: [BUG] `--exclude` quirk, only seems to affect first `--search-path` in some cases (without glob)

wyf027:
## Summary

When multiple `--search-path` roots are given, `--exclude` patterns like `ghi/def` were only evaluated relative to the first root. Matching paths under other roots were not excluded unless the pattern was written as `**/ghi/def`.

This prefixes relative exclude globs with `**/` when more than one search path is configured, so the same pattern applies under every root.

Fixes #1919

## Test plan

- [x] Unit tests for `normalize_exclude_pattern_for_multi_root`
- [x] Integration test `test_exclude_with_multiple_search_paths`
- [x] Manual repro from the issue
- [x] `cargo test` / `cargo clippy -- -D warnings`

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -330,12 +330,18 @@ impl WorkerState {
     fn build_overrides(&self, paths: &[PathBuf]) -> Result<Override> {
         let first_path = &paths[0];
         let config = &self.config;
+        let multi_root = paths.len() > 1;
 
         let mut builder = OverrideBuilder::new(first_path);
 
         for pattern in &config.exclude_patterns {
+            let pattern = if multi_root {
+                Cow::Owned(normalize_exclude_pattern_for_multi_root(pattern))
+            } else {
+                Cow::Borrowed(pattern.as_str())
+            };
             builder
-                .add(pattern)
+                .add(pattern.as_ref())
                 .map_err(|e| anyhow!("Malformed exclude pattern: {}", e))?;
         }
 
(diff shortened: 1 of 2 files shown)


wyf027:
CI is green on the latest push (`03f91aa` — dropped an unrelated #1999 test that had slipped in). Ready for review when you have time.

wyf027:
@sharkdp Friendly ping when you have a moment — small fix for multi-root `--exclude`, CI green on all targets. Thanks!

tavianator:
@leno23 You _just_ opened all these pull requests, what's with the "friendly ping" comment 13 minutes later?

PR #2001, opened 2026-05-16
docs: clarify that -e/--extension matches all entry types

wyf027:
## Summary

Clarifies documentation for `-e` / `--extension` (#1693): the filter matches any entry whose name ends with the extension, including directories (e.g. `clojure.zip/`), not only regular files.

Follows maintainer feedback on #1922:
- Short help (`-h`): minimal — "Filter by extension"
- Long help (`--help`) and `fd.1`: note that `--type` can restrict entry types

## Test plan

- [x] `cargo clippy -- -D warnings`
- [x] Doc-only change (no runtime behavior change)

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -252,9 +252,9 @@ Examples:
 .BI "\-e, \-\-extension " ext
 Filter search results by extension
 .IR ext .
-By default, this applies to all entry types. Use \fB\-\-type\fR to restrict
-the search to specific types.
 This option can be used repeatedly to allow for multiple possible extensions.
+By default, extension filters apply to both files and directories. Use
+\fB\-\-type\fR to restrict the search to specific types.
 
 If you want to search for files without extension, you can use the regex '^[^.]+$'
 as a normal search pattern.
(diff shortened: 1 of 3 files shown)


wyf027:
CI is green on all targets. Doc-only change addressing #1693 / feedback on #1922 — ready for review when you have time.

wyf027:
@tmccombs Thanks for the review on #1922 — pushed `e60582f` to address it:

- **`-h`**: short help is only \`Filter by extension\` (details in \`--help\` / doc comment).
- **man**: avoids “not just files”; says the filter applies to all entry types and points to \`--type\`.
- **scope**: removed accidental \`--full-path\` doc edits from this branch (those are in #1997 only).

CI should stay green. Happy to close this if you prefer updating #1922 instead.

wyf027:
CI is green on the latest push (`9f938be`, extension-only). Thanks again for the review notes on #1922.

doc/fd.1
```diff
@@ -250,9 +250,11 @@ Examples:
 .RE
 .TP
 .BI "\-e, \-\-extension " ext
-Filter search results by file extension
+Filter search results by extension
 .IR ext .
-This option can be used repeatedly to allow for multiple possible file extensions.
+By default, this applies to all entry types. Use \fB\-\-type\fR to restrict
```
tmccombs: I think it would be more clear to say it applies to "both files and directories". 

Also, I'd prefer for this to come after teh comment about it being allowed to be used repeatedly.

Same applies to the help text.

tmccombs (changes requested):

tmccombs (approved):

PR #2002, opened 2026-05-16
fix: normalize --full-path matches for parent search paths (fixes #1513)

### LINKED ISSUE #1513: [BUG] Wrong result when --full-path and ..
### Checks

```
mkdir /tmp/1
cd /tmp/1
touch foo
mkdir /tmp/1/2
cd /tmp/1/2

fd --full-path -t f '/tmp/1/foo' ..
# no result here, but expected: ../foo

fd --full-path -t f '/tmp/1/.*/.*/foo' ..
# it seems to me that this result should not be here, but the result is: ../foo
```

_No response_

fd 9.0.0

```shell
$ uname -srm
Linux 6.5.9-t2-lunar x86_64

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 23.04
Release:	23.04
Codename:	lunar
```

wyf027:
## Summary

Fixes #1513: `--full-path` matching failed when the search path contained `..` (or `.`), because the match string was built as e.g. `/tmp/1/2/../foo` without normalization, while the pattern used the canonical `/tmp/1/foo`.

## Approach

When building the absolute path for `--full-path` matching, normalize via `normpath` only if the relative entry path contains `.` or `..` components; otherwise keep the existing join behavior.

## Testing

- New integration test `test_full_path_normalizes_relative_search_path`
- `cargo test full_path_normalizes`

## Note

Similar fix was proposed in #1969 (closed without merge); happy to credit @lawrence3699 if maintainers prefer that attribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ## Bugfixes
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
+- Normalize `--full-path` matching for relative search paths such as `..`, see #1513 (@leno23).
 - Fire the "search pattern contains a path separator" diagnostic for any pattern containing `/`, not just patterns that happen to name an existing directory. Preserves the legacy Windows behaviour that also flags native `\` separators when the pattern resolves to a real directory. See #1873.
 
 # 10.4.2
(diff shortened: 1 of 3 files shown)


wyf027:
Force-pushed a clean branch (33 lines) — the first push accidentally included unrelated #1919 changes from a dirty worktree. Sorry for the noise.

wyf027:
@sharkdp CI is green on all targets (including Windows after escaping the temp-dir path in the integration test). Fixes #1513 — normalizes `--full-path` match strings when the entry path contains `.` or `..`. Ready for review when you have time.

PR #2004, opened 2026-05-16
fix: stop walk on broken pipe with --glob (fixes #1479)

### LINKED ISSUE #1479: [BUG] Full-path search and globbing leads to `fd` not exit on pipeline closing
### Checks

The shortest example I came up with is this:

```shell
> (fd . -p / | (exit))
# closes immediately, as expected
> (fd -g '*' -p / | (exit))
# hangs until the search is complete
```

Where it matters practically though is when piping into fzf. It will not abort the search even when fzf has exited.

`fd` should exit the same way it does with regex search.

9.0.0

```shell
Linux 6.1.72 x86_64

No LSB modules are available.
Distributor ID:	NixOS
Description:	NixOS 24.05 (Uakari)
Release:	24.05
Codename:	uakari
```

wyf027:
## Summary

Fixes #1479: when a downstream process closes the pipe early (e.g. `fd -g '*' / | head`), fd kept walking until the search finished.

## Cause

`ReceiverBuffer::print` ignored `BrokenPipe` from `print_entry`, so the parallel walker never saw `quit_flag` set.

## Fix

On `BrokenPipe`, set `quit_flag` and stop the receiver loop (same as `flush()` on a closed pipe).

## Testing

- `cargo test test_broken_pipe_stops_glob_search` (Unix): spawns `fd -g '*'`, closes stdout, asserts exit within 3s on a 2000-file tree

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Add `--ignore-parent` option to override `--no-ignore-parent`, see #1958 (@tmchow)
 
 ## Bugfixes
+- Stop the directory walk when stdout hits a broken pipe (e.g. closed downstream reader with `--glob`), see #1479 (@leno23).
 - Handle invalid working directories gracefully when using `--full-path`, see #1900 (@Xavrir).
 - Fire the "search pattern contains a path separator" diagnostic for any pattern containing `/`, not just patterns that happen to name an existing directory. Preserves the legacy Windows behaviour that also flags native `\` separators when the pattern resolves to a real directory. See #1873.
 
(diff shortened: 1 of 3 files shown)


wyf027:
@sharkdp CI is green on all targets. Fixes #1479 — stops the parallel walk when `print_entry` returns `BrokenPipe` (e.g. `fd -g '*' … | head`). Ready for review when you have time.

tmccombs:
This doesn't actually solve the problem. If we get a pipe error on writing to stdout, we already terminate the process. The problem in #1479 is that we don't ever write to the pip, so don't ever get the pip error.

PR #1994, opened 2026-05-09
fix: hint shell builtins in command-not-found error

kimjune01:
Closes #1944.

When `fd -x cd` (or any shell builtin) fails with "command not found", the error now explains that the command is a shell builtin and shows how to invoke it through a shell:

```
[fd error]: Command not found: cd. Note: cd is a shell builtin, not a standalone
program. To run shell builtins, invoke a shell explicitly,
e.g. fd -x sh -c 'cd ... "$1"' sh {}
```

Non-builtins still get the plain "Command not found: foo" message.

The builtin list covers POSIX and common bash builtins that don't have standalone executables. Commands like `echo`, `printf`, and `test` are excluded since they typically exist as `/usr/bin/` binaries and work fine with `-x`.

Changelog entry added.

PR #2005, opened 2026-05-16
fix: hint shell builtins in command-not-found error (fixes #1944)

### LINKED ISSUE #1944: 'command not found' when running fd with `-x cd`
### Checks

Attempting to run fd with arguments including -x cd.
I get the output:
[fd error]: Command not found: cd

fd cds to the output (assuming only one directory is found)

fd 10.4.2

```shell
Linux 6.12.76_1 x86_64
LSB Version:	1.0
Distributor ID:	VoidLinux
Description:	Void Linux
Release:	rolling
Codename:	void
```

wyf027:
## Summary

Fixes #1944 with the same approach as #1994 by @kimjune01, plus a one-line **clippy** fix (`SHELL_BUILTINS.contains(&program)`).

## Note on #1994

[#1994](https://github.com/sharkdp/fd/pull/1994) has the same feature but CI fails on `clippy::manual_contains`. I’m happy to **close this PR** if #1994 is updated — opening this only so maintainers have a green CI option.

## Testing

- `cargo test builtin_tests`

### DIFF
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -98,18 +98,83 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
     ExitCode::Success
 }
 
+/// Common shell builtins that typically do not exist as standalone executables.
+/// When fd encounters a "command not found" error for one of these, it hints
+/// that the user may be trying to use a shell builtin.
+const SHELL_BUILTINS: &[&str] = &[
+    ".", "alias", "bg", "bind", "cd", "command", "declare", "dirs", "eval", "exec", "exit",
+    "export", "fg", "hash", "help", "history", "jobs", "let", "local", "logout", "popd", "pushd",
+    "read", "readonly", "return", "set", "shift", "shopt", "source", "suspend", "times", "trap",
+    "type", "typeset", "unalias", "unset", "wait",
+];
+
+fn is_shell_builtin(program: &str) -> bool {
+    SHELL_BUILTINS.iter().any(|&b| b == program)
+}
+
+fn command_not_found_message(program: &str) -> String {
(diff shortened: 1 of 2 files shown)


wyf027:
All CI checks are green and the branch is mergeable with `master`. Still happy to close this in favor of #1994 if that PR picks up the `contains()` clippy fix — just offering a ready-to-merge option for #1944.

src/exec/command.rs
```diff
@@ -98,13 +98,37 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
...
+fn is_shell_builtin(program: &str) -> bool {
+    SHELL_BUILTINS.contains(&program)
+}
+
+fn command_not_found_message(program: &str) -> String {
+    if is_shell_builtin(program) {
+        format!(
+            "Command not found: {program}. Note: {program} is a shell builtin, \
+             not a standalone program. To run shell builtins, invoke a shell explicitly, \
+             e.g. fd -x sh -c '{program} ... \"$1\"' sh {{}}",
```
ded-furby, line 120: This example is Unix-specific. On Windows, suggesting `sh -c` will usually fail, so the new diagnostic replaces one confusing error with another. Could we make the hint platform-neutral here, or choose the example based on the target shell/OS instead?

ded-furby (changes requested):
Thanks for getting a green version of this fix up. There is one cross-platform problem to address before merge: the new hint hardcodes `sh -c`, which is Unix-specific. `fd` supports Windows targets in CI, and on those builds the suggested command is likely to fail with another command-not-found error. I think this should either use a platform-neutral hint or switch examples per platform.

Issue #612, opened 2020-06-07
Discussion: show Git-ignored files by default?


sharkdp:
Since `fd` was first published, the feature to hide Git-ignored files by default has always been controversial. It's the number one pitfall for new users, as witnessed by the numerous issues that have been opened over time (even though this is the first point in the [Troubleshooting](https://github.com/sharkdp/fd#troubleshooting) section). Even experienced users will likely run into this from time to time.

We have had past discussions about this (see #179), but I'm not so sure anymore if this default is the best possible option for the "average user".

I thought it might make sense to discuss this again and see what others think. Whatever we choose as the default, it will always be easy for users to select a different default via an alias.

**Pro** "*ignore `.gitignore` entries by default*" (current default):
- Most searches are faster if we take `.gitignore` files into account. `.gitignore`d directories tend to contain huge amounts of automatically generated build artifacts or downloaded dependency files. Pruning these directories from the search tree typically results in a faster search overall. There are counterexamples to this where the parsing of long `.gitignore` files takes longer than actually traversing these directories.
- Most of the time, `.gitignore`d results are not "interesting" to the user (however, see counterpart below).
- When running `fd` without any arguments, I typically don't want to see `.gitignore`d files.

**Cons**:
- It can be very confusing to (new) users. If 10% of users go so far as to create a ticket on GitHub to ask about their problem, there must be hundreds of users that ran into this problem at some point.
- Even if you know about the default, it can be annoying to repeat the search because you forgot to add `-I` or `-u`. There are a lot of valid use cases where users are - in fact - interested in results from ignored directories or files. Personally, I would estimate that I use `-uu` or `-HI` in roughly 20% of my searches, which is quite high.

kpym:
I want to add that the nature of the files in `.gitignore` depends a lot on the nature of our projects. In my case for example most of the time the ignored files are files with sensitive information (not crap) that I want to be able to search with `fd`.

But I understand that for other people often the files in `.gitignore` have to be ignored by `fd` as well.

For this reason the desired default behavior is not the same for everyone.

In my opinion, the default behavior should be "search all", because it's easier to figure out why there are too many results than it is to figure out why there are missing results. 

But such a change in behavior will not be backward compatible, which is never good. To overcome this, people must be allowed to easily return to the old default behavior. Hence the need to be able to configure the default behavior of `fd` (#362).

sharkdp:
> because it's easier to figure out why there are too many results than it is to figure out why there are missing results.

[...]

PR #2011, opened 2026-05-17
docs: Update Help String

stevenwalton:
Updated help string to clarify that fd respects gitignore by default. Also added verbage to clarify that globbing and regex patterns are accepted without a flag.

This hopefully reduces confusion as seen in #612

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -22,7 +22,7 @@ use crate::filter::SizeFilter;
 #[command(
     name = "fd",
     version,
-    about = "A program to find entries in your filesystem with glob and regex based matching. By default, fd respects gitignore rules, ignores hidden directories, and is case insensitive.",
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

src/cli.rs
```diff
@@ -22,7 +22,7 @@ use crate::filter::SizeFilter;
 #[command(
     name = "fd",
     version,
-    about = "A program to find entries in your filesystem",
+    about = "A program to find entries in your filesystem with glob and regex based matching. By default, fd respects gitignore rules, ignores hidden directories, and is case insensitive.",
```
tmccombs: nit: It might be better to say 
```suggestion
    about = "A program to find entries in your filesystem with  regex and glob based matching. By default, fd respects gitignore rules, ignores hidden directories, and is case insensitive.",
```

since regex is the default.

stevenwalton: There's an extra space before 'regex' so I removed that, but otherwise I'm good with it.

PR #2016, opened 2026-05-26
feat: respect .gitignore in Jujutsu (.jj) repositories

### LINKED ISSUE #1817: Support for `.gitignore` with `.jj` (Jujutsu VCS) repositories
I'm currently using [Jujutsu VCS](https://jj-vcs.github.io/), a Git-compatible VCS that uses `.gitignore` files. I use it without its co-location feature (which places a `.git` directory in the project root). Instead, I only have a `.jj` directory in my project root. It would be nice if `fd` could look for both `.git` and `.jj` directories to determine if it should observe the `.gitignore` file.

The workaround, using `--no-require-git`, is a bit cumbersome.

wyf027:
## Summary

Fixes #1817.

Jujutsu repositories use a `.jj` directory and still rely on `.gitignore` files, but fd only applies gitignore rules without a `.git` directory when `--no-require-git` is passed.

When any search path is inside a Jujutsu repository (a `.jj` directory exists in an ancestor path), automatically treat the search like `--no-require-git` unless the user explicitly passes `--require-git`.

## Test plan

- [x] `cargo test --locked jujutsu`
- [x] New integration test `test_jujutsu_repo_respects_gitignore`
- [x] Unit test `should_require_git_in_jujutsu_repo`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,7 +15,7 @@ mod walk;
 
 use std::env;
 use std::io::IsTerminal;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow, bail};
@@ -98,7 +98,7 @@ fn run() -> Result<ExitCode> {
         .map(|pat| build_pattern_regex(pat, &opts))
         .collect::<Result<Vec<String>>>()?;
 
-    let config = construct_config(opts, &pattern_regexps)?;
+    let config = construct_config(opts, &pattern_regexps, &search_paths)?;
 
     ensure_use_hidden_option_for_leading_dot_pattern(&config, &pattern_regexps)?;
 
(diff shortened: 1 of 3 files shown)


tmccombs (changes requested):
I think it would be better to address this  upstream in the ignore crate. 

Also this doesn't handle if the search paths are relative, and if one search path has a .jj directory but others don't, it disables requiring a .git (or .jj) directory for all of them, which probably isn't desirable.

wyf027:
Addressed the review feedback and CI failure in commit 3267318. The .jj detection now resolves relative search paths against the current directory and only relaxes require_git when all search paths are inside a Jujutsu repo, so a mixed set of .jj and non-.jj search paths no longer disables the git requirement globally. Also ran cargo fmt to clear the formatter check.

Local verification:
- cargo fmt --check
- cargo test should_require_git
- cargo test detects_jujutsu_repo_for_relative_search_path
- cargo test test_jujutsu_repo_respects_gitignore
- cargo clippy --all-targets -- -D warnings

tmccombs:
This now has the opposite problem where it will only use gitignore if _all_ the base paths have .jj directories. And furthermore, doesn't handle cases where a subdirectory has a .jj dir, or has a .gitignore file, but not a .git OR .jj directory. 

I really think this would be better handled by looking for .jj directories in the ignore crate itself, and would suggest you instead open a PR with [ripgrep](https://github.com/burntsushi/ripgrep).

In lieu of that, along with my suspicion that  this is AI slop that hasn't had human review. I am closing this PR.

PR #2018, opened 2026-05-26
feat: add --links filter for hard link count (fixes #1844)

### LINKED ISSUE #1844: [Feature request] Support "-links" option from find
Hello there,

Thanks for your great tool.
Could consider adding the `-links` options of `find` to `fd` ?

```
-links n
          File has less than, more than or exactly n hard links.
```

It is really useful for certain cases, such as listing only leaf folders.

Cheers!

wyf027:
## Summary

- Adds `--links` to filter search results by hard link count (Unix only)
- Supports the same constraint syntax as `--size`: exact count, `+N` (at least), `-N` (at most)

## Motivation

Fixes #1844. Brings fd closer to find(1)'s `-links` option for use cases like finding singly-linked files or hard-linked duplicates.

## Example

```bash
# Files with exactly one hard link
fd --type f --links 1

# Files with more than one hard link
fd --type f --links +1

# Directories with at most one hard link (leaf directories on some filesystems)
fd --type d --links -1
```

## Test plan

- [x] Unit tests for `--links` parsing and matching logic
- [x] Integration test `test_links_filter` with hard-linked files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 - Add `--ignore-parent` option to override `--no-ignore-parent`, see #1958 (@tmchow)
 - Add `--exact` option to match the entire filename exactly (literal, non-substring).
+- Add `--links` filter to limit results by hard link count (Unix only), see #1844
 
 ## Bugfixes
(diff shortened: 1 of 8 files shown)


tmccombs:
As mentioned on the issue, this isn't how we want to approach this

PR #2019, opened 2026-05-26
feat: run --exec-batch batches in parallel when --threads is set (fixes #1760)

### LINKED ISSUE #1760: Allow `--exec-batch` to be used in parallel
The `--exec-batch` parameter runs the specified command multiple times over the biggest possible batches of file names. However, if a *lot* of files are to be found, and this parameter is used, it very easily becomes the bottleneck.

Batches always (appear to) run sequentially, ignoring the `--threads` parameter. If the command takes a long time (which tends to happen when you throw 10000 filenames at a singular program lol), then that singular invocation just sits there maxing out a single CPU core while 15 other cores and fd itself lay dormant.

It'd be nice to be able to get parallelism benefits of `--exec` with batching benefits of `--exec-batch`.

---

My particular use case right now is that I'm bored and I want to scan a lot of files with ClamAV. Its `clamscan` tool appears to run through files completely sequentially too, running on one CPU, and also always takes a few seconds on startup just to load its databases. Upon running, it seems to consume at most ~1.5GiB of RAM. I have 64GiB RAM and 16 cores, so it seems to be well within budget to run it in parallel on my machine.

In particular, I'm running this:

    sudo fd --type f --exec-batch clamscan -i --no-summary

As of right now, it seems my only choices is to either have only one core used, or write some weird script that would batch stuff up and dispatch instances of clamscan myself (I'm too lazy for that though lol), or use `--exec` instead and have all cores fully used at cost of ~10000x the "loading database" overhead, since there would be one invocation per one file.

I hope I'm not missing something lol

wyf027:
## Summary

Fixes #1760.

`--exec-batch` previously ran every batch invocation sequentially, even when `--threads` was set. This change enables parallel batch execution when the user **explicitly** passes `--threads` with a value greater than 1.

Default behavior is unchanged: without an explicit `--threads` flag, batches still run sequentially on the main thread. This preserves streaming/interactive use cases discussed in the issue (linters, editors, incremental output).

When parallel mode is active, command output is buffered and written under stdout/stderr locks, similar to `--exec` with multiple threads.

## Changes

- Track whether `--threads` was explicitly provided (`Config::explicit_threads`)
- Refactor batch building to collect `Command` instances before execution
- Execute batches concurrently (up to the thread limit) when explicitly requested
- Document the new behavior in `--exec-batch` help text
- Add integration test for `--batch-size=2 --threads=4 --exec-batch`

## Test plan

- [x] `cargo test test_exec_batch`
- [x] `cargo test` (full suite, 108 tests)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -940,6 +940,10 @@ impl clap::Args for Exec {
                        '{{':   literal '{' (for escaping)\n  \
                        '}}':   literal '}' (for escaping)\n\n\
                      If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
+                     By default, batch commands run sequentially on the main thread. If `--threads`\n\
+                     is explicitly set to a value greater than 1, multiple batches may run in\n\
+                     parallel (up to the given thread count). Command output is buffered in this\n\
+                     mode, similar to `--exec` with multiple threads.\n\n\
                      Examples:\n\n  \
                        - Find all test_*.py files and open them in your favorite editor:\n\n      \
                            fd -g 'test_*.py' -X vim\n\n  \
--- a/src/exec/command.rs
+++ b/src/exec/command.rs
@@ -1,7 +1,10 @@
 use std::io;
 use std::io::Write;
+use std::sync::Mutex;
+use std::thread;
 
 use argmax::Command;
+use crossbeam_channel::unbounded;
 
 use crate::error::print_error;
 use crate::exit_codes::ExitCode;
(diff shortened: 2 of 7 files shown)


src/exec/mod.rs
```diff
@@ -90,36 +90,51 @@ impl CommandSet {
...
     }
 
-    pub fn execute_batch<I>(&self, paths: I, limit: usize, path_separator: Option<&str>) -> ExitCode
+    pub fn execute_batch<I>(
+        &self,
+        paths: I,
+        limit: usize,
+        path_separator: Option<&str>,
+        parallel: bool,
+        threads: usize,
```
tmccombs, line 99: Why not pass an `Option<usize>` for the threads, instead of separate arguments for whether to use multiple threads, and the number of threads.

src/exec/mod.rs
```diff
@@ -90,36 +90,51 @@ impl CommandSet {
...
+    ) -> ExitCode
+    where
+        I: Iterator<Item = PathBuf>,
+    {
+        match self.collect_batches(paths, limit, path_separator) {
+            Ok(commands) => execute_batch_commands(commands, parallel, threads),
```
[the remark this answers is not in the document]
tmccombs: This is a legitimate issue, and should be addressed. Ideally, adding parallelism on the batches shouldn't result in waiting for all results before processing the first batch.

src/exec/command.rs
```diff
@@ -98,6 +101,114 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
...
+                }
+            });
+        }
+    });
+
+    drop(result_tx);
+
+    let mut exit_code = ExitCode::Success;
+
+    for result in result_rx {
```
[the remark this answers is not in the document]
tmccombs: Yeah, I'm concerned about this. This also means that the user won't see any output until all of the commands have completed.

src/exec/mod.rs
```diff
@@ -173,41 +186,39 @@ impl CommandBuilder {
...
+            self.finish(commands)?;
         }
 
         let arg = self.path_arg.generate(path, separator);
         if !self
             .cmd
             .args_would_fit(iter::once(&arg).chain(&self.post_args))
         {
-            self.finish()?;
+            self.finish(commands)?;
```
tmccombs, line 204: This actually made me realize an existing... oddity with the current implementation. If we have multiple batch commands, and we overflow the argument limit for the second comand before the first, we'll end up running the second command before the first one.  Which is unexpected, and probably a bug.

src/exec/command.rs
```diff
@@ -98,6 +101,114 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
...
+    }
+
+    exit_code
+}
+
+fn execute_batch_commands_parallel(commands: Vec<Command>, threads: usize) -> ExitCode {
+    let num_workers = threads.min(commands.len());
+    let (work_tx, work_rx) = unbounded::<Command>();
+    for cmd in commands {
+        let _ = work_tx.send(cmd);
```
tmccombs, line 138: This breaks the behavior listed in the man page

> Like --exec, this can be used multiple times, in which case each command  will  be  run in the order given.

There is no longer any guarantee the commands will be run in the order given. 

Granted, there is an existing bug around that, but this makes it worse. Probably what should be done, assuming that we do add support for this, which I'm still not completely convinced we should, is run each batch command in parallel for all paths before proceeding to the next. 

Or possibly, ensure that all batches are the same size, somehow (difficult with the command line limit), and run all the commands for each batch on the same thread, in series.  Although in that case, I think  the behavior should probably be better documented.

src/exec/command.rs
```diff
@@ -98,6 +101,114 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
...
+    }
+    drop(work_tx);
+
+    let (result_tx, result_rx) = unbounded::<BatchCommandResult>();
+    let stdout_lock = Mutex::new(());
+    let stderr_lock = Mutex::new(());
```
tmccombs, line 144: I'm pretty sure these locks are unnecessary and redundant with locking stdout and stderr.

And we only ever acquire them in a single-threaded context.

src/exec/command.rs
```diff
@@ -98,6 +101,114 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
...
+
+    thread::scope(|scope| {
+        for _ in 0..num_workers {
+            let work_rx = work_rx.clone();
+            let result_tx = result_tx.clone();
+
+            scope.spawn(move || {
```
tmccombs, line 151: we don't necessarily need separate threads to get parallelism. We could just spawn the commands asynchronously and buffer the output, and the subprocesses would provide the parallelism.

src/cli.rs
```diff
@@ -940,6 +940,10 @@ impl clap::Args for Exec {
                        '{{':   literal '{' (for escaping)\n  \
                        '}}':   literal '}' (for escaping)\n\n\
                      If no placeholder is present, an implicit \"{}\" at the end is assumed.\n\n\
+                     By default, batch commands run sequentially on the main thread. If `--threads`\n\
+                     is explicitly set to a value greater than 1, multiple batches may run in\n\
+                     parallel (up to the given thread count). Command output is buffered in this\n\
+                     mode, similar to `--exec` with multiple threads.\n\n\
```
tmccombs, line 946: I don't love tying this to using `--threads`. There are several reasons why running the `--exec-batch` commands in parallel may be problematic:

1.  The program may be interactive (for example, vim) in which case buffering the output, and running multiple instances would be problematic.
2. Buffering all the output may use a large amount of memory
3. Buffering all the output may mean it takes longer for the user to get any feedback.

But with this, if the user  wants to avoid any of those issues, they now can't specify the number of threads.

tmccombs (changes requested):
In addition to the other comments I leff, I think this needs more design work, and decisions on what behavior is desired here, if this feature is even worth adding,  what flag(s) would enable it, etc.

src/exec/mod.rs
tavianator: FWIW I'm aware of that but I don't really consider it unexpected

Issue #1845, opened 2025-11-23
Implement --list-details natively


tmccombs:
Currently, the --list-details option is implemented by calling `ls -lhd` (possibly with color). 

This has a number of downsides, including poor performance, inability to work if `ls` is not installed (common on windows), etc.

I propose that instead of using an external tool we implement `--list-details` natively, probably using a similar flow to [jsonl output](https://github.com/sharkdp/fd/pull/1800). 

### Potential Challenges

- We would need code to format the permissions to a rwx string (for example with something like https://docs.rs/unix_mode/latest/unix_mode/fn.to_string.html)
- We would need something to print the file size in a human-readable way (https://docs.rs/humansize/latest/humansize/ ?)
- Is the format of `ls -lhd` the same on platforms with non-GNU ls?
- What does the permissions field even look like for windows?

### Related Issues

- https://github.com/sharkdp/fd/issues/1775
-  https://github.com/sharkdp/fd/issues/1595
- https://github.com/sharkdp/fd/issues/1482
- https://github.com/sharkdp/fd/pull/1842

PR #2022, opened 2026-05-27
Fix --list-details failing with "Argument list too long" (fixes #1775)

### LINKED ISSUE #1775: [BUG] --list-details / -X ls -ld gives argument list too long
### Checks

```sh
fd-v10.2.0-x86_64-unknown-linux-gnu/fd --full-path <some path with many files> --unrestricted --list-details
```

gives

```out
[fd error]: Problem while executing command: Argument list too long (os error 7)
```

This only happens when `-X` or `--list-details` is used.

```sh
fd-v10.2.0-x86_64-unknown-linux-gnu/fd --full-path <some path with many files> --unrestricted | wc -l
```

gives

```out
890439
```

Fast listing details such as owner / group / size / ... similar to `ls -ld` or `find <> -printf '%u %g %T@ %M %s "%p"\n'`.

As soon as I use `--list-details` with `--batch-size` to get rid of above error `fd` we are comparable if not slower than `find`. (I didn't do serious speed tests, just `time` on a couple of trails.)

fd 10.2.0

```shell
Linux 4.18.0-372.16.1.el8_6.x86_64 x86_64
```

wyf027:
## Summary

Fixes #1775.

When `--list-details` matches a very large number of files, fd can fail with:

```
[fd error]: Problem while executing command: Argument list too long (os error 7)
```

This happens when argmax's size estimate allows another path argument but `try_arg` still returns `E2BIG`. The batch executor now flushes the current `ls` invocation and retries instead of aborting.

Also removes the clap restriction that prevented using `--batch-size` together with `--list-details`.

## Test plan

- [x] `cargo test test_list_details_many_files` (40k files, ~15s)
- [x] `cargo test exec_batch` (existing batch tests)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -498,7 +498,6 @@ pub struct Opts {
         long,
         value_name = "size",
         hide_short_help = true,
-        requires("exec_batch"),
         value_parser = value_parser!(usize),
         default_value_t,
         help = "Max number of arguments to run as a batch size with -X",
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -17,6 +17,20 @@ use crate::fmt::{FormatTemplate, Token};
 use self::command::{execute_commands, handle_cmd_error};
 pub use self::job::{batch, job};
 
+/// Returns true if an I/O error indicates that a command line exceeded OS limits.
+fn is_argument_list_too_long(err: &io::Error) -> bool {
+    #[cfg(unix)]
+    {
+        // E2BIG: argmax uses this when no more arguments can be added.
+        err.raw_os_error() == Some(7)
+    }
+    #[cfg(not(unix))]
+    {
+        let _ = err;
+        false
+    }
+}
+
 /// Execution mode of the command
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ExecutionMode {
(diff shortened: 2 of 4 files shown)


src/cli.rs
```diff
@@ -498,7 +498,6 @@ pub struct Opts {
         long,
         value_name = "size",
         hide_short_help = true,
-        requires("exec_batch"),
         value_parser = value_parser!(usize),
         default_value_t,
         help = "Max number of arguments to run as a batch size with -X",
```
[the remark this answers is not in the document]
tmccombs: Yeah. This isn't right. It might be ok to allow batch-size with list-details. Maybe. But it definitely shouldn't remove the requires altogether, and if https://github.com/sharkdp/fd/issues/1845 is done, then batch size would no longer apply to list-details, so I'm inclined against it.

tests/tests.rs
```diff
@@ -2618,6 +2618,22 @@ fn test_list_details() {
     te.assert_success_and_get_output(".", &["--list-details"]);
 }
 
+/// `--list-details` runs `ls` in batch mode; make sure large result sets are split
+/// across multiple invocations instead of failing with "Argument list too long".
+#[test]
+#[cfg(unix)]
+fn test_list_details_many_files() {
```
tmccombs, line 2625: Does this test actually fail with the current code?

src/exec/mod.rs
```diff
@@ -186,7 +200,18 @@ impl CommandBuilder {
             self.finish()?;
         }
 
-        self.cmd.try_arg(arg)?;
+        match self.cmd.try_arg(&arg) {
+            Ok(_) => {}
+            Err(err) if is_argument_list_too_long(&err) => {
+                if self.count > 0 {
```
tmccombs, line 206: If this happened that means there is a bug in args_would_fix that should be fixed there.

tmccombs (commented):
I'm not sure this is actually an issue, and   if it is, I think it should be fixed in argmax. 

Im going to close this.

PR #2042, opened 2026-06-29
[codex] Reject overflowing size filters

kmg0308:
## Summary
- Use checked multiplication while parsing --size values so overflowing inputs are rejected.

## Validation
- size parser tests, CLI check, cargo fmt check, and diff check passed in Docker.

## Notes
- Opened as a draft PR for maintainer review.


tmccombs:
Duplicate of #2035

PR #2043, opened 2026-06-29
refactor(walk): extract entry filtering and normalization logic

parneetsingh022:
This PR refactors the directory traversal logic in WorkerState to improve readability and maintainability. Previously, the evaluation of filtering constraints (depth, patterns, extensions, size, time) was implemented as a large block of inline conditionals within the core walk loop.

This change extracts that logic into a dedicated EntryFilter struct, cleanly separating the traversal mechanics from the filtering rules. All existing traversal behavior and the execution order of the filter checks are strictly preserved.

## Changes

This PR cleans up the walk entry processing so the main loop is easier to read and reason about.

* Moved file filtering logic into a new `EntryFilter` helper, with separate methods for pattern, size, owner, and other checks.

* Added `normalize_walk_entry` to handle `ignore::DirEntry` conversion, broken symlink recovery, and walker errors outside of the main loop.

* Moved Unix-specific owner-check logic into `check_owner`, keeping the main path less cluttered with platform-specific conditionals.

### DIFF
--- a/src/walk.rs
+++ b/src/walk.rs
@@ -302,6 +302,254 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
     }
 }
 
+struct EntryFilter<'a> {
+    config: &'a Config,
+    patterns: &'a [Regex],
+}
+
+impl<'a> EntryFilter<'a> {
+    pub fn new(config: &'a Config, patterns: &'a [Regex]) -> Self {
+        Self { config, patterns }
+    }
+
+    /// Fast-path checks that operate on the raw ignore::DirEntry
+    fn evaluate_before_normalization(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
+        if let Some(state) = self.check_ignore_file(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_root_dir(entry) {
+            return Some(state);
+        }
+        None
+    }
+
+    /// Evaluates a normalized entry against all configured constraints to determine its walk state.
+    fn evaluate(&self, entry: &DirEntry) -> Option<WalkState> {
+        if let Some(state) = self.check_min_depth(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_patterns(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_extensions(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_file_types(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_owner(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_size(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_modification_time(entry) {
+            return Some(state);
+        }
+        None
+    }
+
+    fn check_ignore_file(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
+        // If the entry is a directory that contains a
+        // "ignore contain" file, we want to skip this
+        // directory.
+        // Check the filetype first to avoid unnecessary
+        // syscalls.
+        if entry.file_type().is_some_and(|t| t.is_dir()) {
+            let entry_path = entry.path();
+            if self
+                .config
+                .ignore_contain
+                .iter()
+                .any(|ic| entry_path.join(ic).exists())
+            {
+                return Some(WalkState::Skip);
+            }
+        }
+
+        None
+    }
+
+    fn check_root_dir(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
+        if entry.depth() == 0 {
+            // Skip the root directory entry.
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_min_depth(&self, entry: &DirEntry) -> Option<WalkState> {
+        let Some(min_depth) = self.config.min_depth else {
+            return None;
+        };
+
+        if entry.depth().is_none_or(|depth| depth < min_depth) {
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_patterns(&self, entry: &DirEntry) -> Option<WalkState> {
+        let entry_path = entry.path();
+
+        let search_str = search_str_for_entry(entry_path, self.config.full_path_base.as_deref());
+
+        if !self
+            .patterns
+            .iter()
+            .all(|pat| pat.is_match(&filesystem::osstr_to_bytes(search_str.as_ref())))
+        {
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_extensions(&self, entry: &DirEntry) -> Option<WalkState> {
+        let entry_path = entry.path();
+        if let Some(ref exts_regex) = self.config.extensions {
+            if let Some(path_str) = entry_path.file_name() {
+                if !exts_regex.is_match(&filesystem::osstr_to_bytes(path_str)) {
+                    return Some(WalkState::Continue);
+                }
+            } else {
+                return Some(WalkState::Continue);
+            }
+        }
+        None
+    }
+
+    fn check_file_types(&self, entry: &DirEntry) -> Option<WalkState> {
+        if let Some(ref file_types) = self.config.file_types
+            && file_types.should_ignore(&entry)
+        {
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_size(&self, entry: &DirEntry) -> Option<WalkState> {
+        let entry_path = entry.path();
+        if !self.config.size_constraints.is_empty() {
+            if entry_path.is_file() {
+                if let Some(metadata) = entry.metadata() {
+                    let file_size = metadata.len();
+                    if self
+                        .config
+                        .size_constraints
+                        .iter()
+                        .any(|sc| !sc.is_within(file_size))
+                    {
+                        return Some(WalkState::Continue);
+                    }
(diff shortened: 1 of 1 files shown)


src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+        if let Some(state) = self.check_size(entry) {
+            return Some(state);
+        }
+        if let Some(state) = self.check_modification_time(entry) {
+            return Some(state);
+        }
+        None
+    }
+
+    fn check_ignore_file(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
```
tmccombs: This should probably be named `check_ignore_contain` or maybe `check_contains_sentinel_file`? 

The current name suggests that it is checking ignore files like `.ignore` and `.gitignore`, but this is actually checking if the directory contains a file passed with `--ignore-contain`. 

Also, please add a doc-comment for this.

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+            return Some(state);
+        }
+        if let Some(state) = self.check_root_dir(entry) {
+            return Some(state);
+        }
+        None
+    }
+
+    /// Evaluates a normalized entry against all configured constraints to determine its walk state.
+    fn evaluate(&self, entry: &DirEntry) -> Option<WalkState> {
```
tmccombs: I wonder if it would be worth using a macro like

```rust

macro_rules! entry_filter {
  ($e:expr) => {
    let result = $e;
    if result.is_some() {
      return result;
    }
  }
}
```

Then each of these checks becomes `entry_filter!(self.check_min_depth(entry))`.

Or alternatively, have this return `Result<(), WalkState>`, and then this becomes

```rust

self.check_min_depth(entry)?;
self.check_patterns(entry)?;
//...
```

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+                .any(|ic| entry_path.join(ic).exists())
+            {
+                return Some(WalkState::Skip);
+            }
+        }
+
+        None
+    }
+
+    fn check_root_dir(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
```
tmccombs:
```suggestion
    /// Check if the entry is a root directory, and skip it if so,
    /// so that we don't include the root search paths in the output.
    fn check_root_dir(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
```

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+    patterns: &'a [Regex],
+}
+
+impl<'a> EntryFilter<'a> {
+    pub fn new(config: &'a Config, patterns: &'a [Regex]) -> Self {
+        Self { config, patterns }
+    }
+
+    /// Fast-path checks that operate on the raw ignore::DirEntry
+    fn evaluate_before_normalization(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
```
tmccombs: I might call this something like `evaluate_dir` , since these are checks we need to do on directories before traversing them. Or possibly `evaluate_raw` since this is evaluating the raw ignore DirEntry.

src/walk.rs
tmccombs: Or maybe `skip_ignored_dir`  to make it clear this is skipping over directorories containing an `--ignore-contain` file (I am kind of regretting the name of that flag...)

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+    fn check_root_dir(&self, entry: &ignore::DirEntry) -> Option<WalkState> {
+        if entry.depth() == 0 {
+            // Skip the root directory entry.
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_min_depth(&self, entry: &DirEntry) -> Option<WalkState> {
```
tmccombs: Most of these functions are actually just checking a condition, and return WalkState::Continue. I wonder if i instead of these returning Option, it woudl be simpler to just have them return `bool`, and do the `return Some(WalkState::Continue`)` in the evaluate method

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+            .iter()
+            .all(|pat| pat.is_match(&filesystem::osstr_to_bytes(search_str.as_ref())))
+        {
+            return Some(WalkState::Continue);
+        }
+
+        None
+    }
+
+    fn check_extensions(&self, entry: &DirEntry) -> Option<WalkState> {
```
tmccombs:
```suggestion
    /// Filter out unwanted extensions
    fn check_extensions(&self, entry: &DirEntry) -> Option<WalkState> {
```

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+            if !matched {
+                return Some(WalkState::Continue);
+            }
+        }
+
+        None
+    }
+
+    #[cfg(unix)]
+    fn check_owner(&self, entry: &DirEntry) -> Option<WalkState> {
```
tmccombs: nit: Most of the methods are in the same order they are invoked, except `check_owner` is defined after check_size and check_modification_time, despite the fact it is called before them

src/walk.rs
```diff
@@ -302,6 +302,238 @@ impl<'a, W: Write> ReceiverBuffer<'a, W> {
...
+            } else {
+                return Some(WalkState::Continue);
+            }
+        }
+
+        None
+    }
+
+    #[cfg(not(unix))]
+    #[inline]
```
tmccombs: I would hope this inline annotation isn't necessary

tmccombs: but it might be better to just not have this defined for non-unix, and have a `#[cfg(unix)]` on the call as well.

tmccombs (changes requested):
Have you done any benchmarking or comparison of assembly on the release build to see if this has any negative impact on performance?

I would hope that it the wrapping and unwrapping of options can mostly be optimized away, but I'm not sure, and since this is in the hot path, I am slightly worried about it.

parneetsingh022:
> Have you done any benchmarking or comparison of assembly on the release build to see if this has any negative impact on performance?
> 
> I would hope that it the wrapping and unwrapping of options can mostly be optimized away, but I'm not sure, and since this is in the hot path, I am slightly worried about it.

I’ve tried to address the concerns mentioned in the PR comments/review.

Regarding the performance concern: I benchmarked this PR against master using release builds and hyperfine.

Benchmark setup:

* OS: Windows
* Benchmark tool: hyperfine
* Baseline: master
* PR branch: this PR
* Benchmark directory: official Rust repository **checkout**

I used a local checkout of the official Rust repository as the benchmark directory because it is large enough to make traversal/filtering costs measurable and should give a more useful signal than benchmarking only on the fd repository itself.

I do not see any measurable performance impact from this PR. All benchmark results are within noise, with master and this PR performing effectively the same across the tested cases.

That matches my expectation, since this PR does not change the underlying logic. It is primarily a refactor that splits the existing logic into smaller functions to improve readability and maintainability.

I ran local benchmarks on Windows using `hyperfine --warmup 10 --runs 50`.

| Benchmark | Master | PR | Result |
|---|---:|---:|---|
| Broad traversal | 498.4 ms ± 27.9 ms | 497.4 ms ± 30.2 ms | PR 1.00 ± 0.08× faster |
| Name pattern | 479.6 ms ± 19.8 ms | 481.6 ms ± 21.3 ms | Master 1.00 ± 0.06× faster |
| Extension filter | 491.5 ms ± 18.0 ms | 491.0 ms ± 21.9 ms | PR 1.00 ± 0.06× faster |
| Type filter | 493.2 ms ± 17.2 ms | 490.4 ms ± 17.3 ms | PR 1.01 ± 0.05× faster |
| Size filter | 1.720 s ± 0.051 s | 1.728 s ± 0.040 s | Master 1.00 ± 0.04× faster |

Overall, the results look effectively within benchmark noise rather than showing a clear regression.

src/walk.rs
parneetsingh022: Internal functions now return a `bool`, I guess this might be better implementation than using a macro (which would have been effective on previous implementation).

```rust
if self.is_below_min_depth(entry)
    || self.fails_pattern_filter(entry)
    || self.fails_extension_filter(entry)
    || self.matches_ignored_file_type(entry)
    || fails_owner_constraints
    || self.fails_size_constraints(entry)
    || self.fails_modification_time_constraints(entry)
{
    return Err(WalkState::Continue);
}
```

src/walk.rs
parneetsingh022: I guess you are right. There is no point in returning `Option<WalkState>` on every function. I've refactored the code to return `bool`, and now only `evaluate` and `evaluate_raw` functions use `Result<(), WalkState>`.

src/walk.rs
parneetsingh022: Thanks, that makes sense. I removed the extra non-Unix helper implementation.

Instead, the owner constraint check is now only called on `Unix`, and `non-Unix` builds just set `fails_owner_constraints` to false inside evaluate:

```rust
#[cfg(unix)]
let fails_owner_constraints = self.fails_owner_constraints(entry);

#[cfg(not(unix))]
let fails_owner_constraints = false;
```

PR #2054, opened 2026-07-04
Include the underlying reason in invalid --changed-before/-within date errors

### LINKED ISSUE #2053: [BUG] `fd` should print more specific error messages for `--changed-` date parsing
### Checks

This, is rather unintuitive:

```
fd -u --changed-before=2025-11-31                                                                                                                                                     1:02:18
[fd error]: '2025-11-31' is not a valid date or duration. See 'fd --help'.
```

If `fd` fails to parse the date, it should unwrap the inner error message and actually tell the user _why_ this is not a valid date according to `fd`.

The fact that it might not be a valid day on that calendar year should be presented to the user. The natural inclination is to assume it's a formatting issue.

`#fb1486f`

```shell
Darwin 23.6.0 arm64
```

momomuchu:
`fd --changed-before=2025-11-31` (an invalid calendar date) prints a generic "'...' is not a valid date or duration" and hides the actual reason.

`TimeFilter::from_str` returned `Option<SystemTime>` via `.ok()?`, discarding the underlying parse errors from all four format attempts (span, timestamp, datetime, `@timestamp`). It now returns `Result<_, String>` and, on total failure, surfaces the `DateTime` parser's error, which is the most relevant one for calendar-date-shaped input.

Before: `'2025-11-31' is not a valid date or duration. See 'fd --help'.`
After: `'2025-11-31' is not a valid date or duration: parsed date is not valid: parameter 'day' for '2025-11' is invalid, must be in range 1..=30. See 'fd --help'.`

Added a unit test asserting the error for an out-of-range calendar date includes the reason. `cargo test`, clippy (`-Dwarnings`) and fmt clean. CHANGELOG entry added.

Note on non-calendar input: the surfaced reason is always the `DateTime` (calendar) parser's error. That is precise for calendar-shaped mistakes (`2025-13-01`, `2025-02-30`, ...), but for input that is not date-shaped at all (e.g. a mistyped duration like `5xyz`) the reason reads as `failed to parse year in date`, which is a bit off-topic. It is still strictly more informative than the old generic message; if you'd prefer, I can use a neutral fallback wording for the non-calendar case instead.

Closes #2053

### DIFF
--- a/src/filter/time.rs
+++ b/src/filter/time.rs
@@ -216,35 +216,85 @@ mod tests {
         );
     }
 
+    /// Length of the longest contiguous substring shared by `a` and `b`.
+    ///
+    /// Used to check that two error messages share substantial content
+    /// without hardcoding what that content actually says.
+    fn longest_common_substring_len(a: &str, b: &str) -> usize {
+        let a: Vec<char> = a.chars().collect();
+        let b: Vec<char> = b.chars().collect();
+        let mut prev = vec![0usize; b.len() + 1];
+        let mut best = 0;
+        for i in 1..=a.len() {
+            let mut cur = vec![0usize; b.len() + 1];
+            for j in 1..=b.len() {
+                if a[i - 1] == b[j - 1] {
+                    cur[j] = prev[j - 1] + 1;
+                    best = best.max(cur[j]);
+                }
+            }
+            prev = cur;
+        }
+        best
+    }
+
     #[test]
     fn invalid_calendar_date_error_includes_inner_reason() {
-        // November only has 30 days, so this is not a valid calendar date.
-        // The error should surface the actual underlying parse failure
-        // instead of a generic rejection. We deliberately don't assert on
-        // jiff's exact wording (e.g. that it mentions "day"), since that
-        // would break on an unrelated jiff version bump. Instead, check
-        // that the message is non-empty and differs from the message for a
-        // differently-malformed input, which shows the inner reason is
-        // actually being propagated. See
+        // The error must surface *why* parsing failed, not merely echo the
+        // raw input back under a generic wrapper. We deliberately don't
+        // assert on jiff's exact wording (e.g. that it mentions "day"),
+        // since that would break on an unrelated jiff version bump. See
         // https://github.com/sharkdp/fd/issues/2053
-        let day_err = TimeFilter::before("2025-11-31").unwrap_err().to_string();
-        let garbage_err = TimeFilter::before("not-a-real-date")
-            .unwrap_err()
-            .to_string();
-        assert!(!day_err.is_empty());
-        assert_ne!(
-            day_err, garbage_err,
-            "distinct invalid inputs should surface distinct underlying reasons, not a shared generic message"
-        );
+        //
+        // A naive check that two different invalid inputs produce two
+        // different messages is gameable: a regression that dropped the
+        // real parse reason but still echoed the input (e.g.
+        // `format!("could not parse '{s}' as a date")`) would still make
+        // the two messages differ, purely because the inputs differ, while
+        // never actually surfacing the reason.
+        //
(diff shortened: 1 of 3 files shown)


src/filter/time.rs
```diff
@@ -26,31 +26,47 @@ fn now() -> Zoned {
...
 impl TimeFilter {
-    fn from_str(s: &str) -> Option<SystemTime> {
+    /// Parses a duration/timestamp/date/`@`-prefixed unix-timestamp string.
+    ///
+    /// On failure, returns the underlying parse error message from the
+    /// `DateTime` parser (the calendar-date format, and the most common
+    /// source of confusing failures, e.g. `2025-11-31` which is not a valid
+    /// calendar date) instead of silently discarding it, so callers can tell
+    /// the user *why* their input was rejected.
+    fn from_str(s: &str) -> Result<SystemTime, String> {
```
tmccombs: Why is the error type a string, and not an anyhow error or jiff::Error?

src/filter/time.rs
```diff
@@ -199,4 +215,23 @@ mod tests {
                 .applies_to(&t1s_later)
         );
     }
+
+    #[test]
+    fn invalid_calendar_date_error_includes_inner_reason() {
```
tmccombs, line 285: This test seems a little brittle. If the jiff error messages change, then this test could fail on a version update.

src/filter/time.rs
```diff
@@ -26,31 +26,47 @@ fn now() -> Zoned {
...
+                if let Some(timestamp_secs) = s.strip_prefix('@')
+                    && let Ok(timestamp_secs) = timestamp_secs.parse()
+                {
+                    return Ok(UNIX_EPOCH + Duration::from_secs(timestamp_secs));
+                }
+                // None of the supported formats matched. The `DateTime`
+                // parser gives the most useful reason for calendar-date-
+                // shaped input (the common case for this kind of mistake),
+                // so surface that instead of a generic message.
+                Err(datetime_err.to_string())
```
tmccombs: Is this error message always user-friendly, or are there cases where this could expose implementation details.

src/filter/time.rs
```diff
@@ -26,31 +26,47 @@ fn now() -> Zoned {
...
+                .later()
+                .map(Into::into)
+                .map_err(|e| e.to_string()),
+            Err(datetime_err) => {
+                if let Some(timestamp_secs) = s.strip_prefix('@')
+                    && let Ok(timestamp_secs) = timestamp_secs.parse()
+                {
+                    return Ok(UNIX_EPOCH + Duration::from_secs(timestamp_secs));
+                }
+                // None of the supported formats matched. The `DateTime`
```
tmccombs, line 65: This could however be more confusing if the user meant to use a span, or timestamp.

momomuchu:
Thanks, good points. 1. Error type: switched from_str/before/after to anyhow::Result - the rest of the crate already uses anyhow (main.rs, walk.rs), String was the outlier. main.rs needed no change since anyhow::Error's Display gives the same message; fd is bin-only so no API concern. 2. Test brittleness: reworked it so it no longer matches jiff's literal wording - it checks that two dates invalid for the same reason share more of the message than a date invalid for a different reason (computed common-substring length), so it survives a jiff reword but still proves the real parse reason is surfaced, not just the echoed input. 3. Error message exposure: no sensitive data - the appended text only echoes the user's own input plus generic range wording (must be in range 1..=30), no paths or secrets. 4. Span/timestamp confusion: fair - a malformed span/timestamp currently shows the calendar-date parser reason, which can read off-topic. I can gate the appended reason to only show when the input looks date-shaped, with a neutral fallback otherwise. Want that in this PR or a follow-up?

momomuchu:
Went with in this PR since it stayed small. If you'd rather land the original fix alone I can pull the last commit into a follow-up.

PR #2055, opened 2026-07-05
Improve changed date parse errors

KSHITIZ6341:
## Summary

- Improve error reporting for invalid `--changed-*` date arguments.
- Keep the parsed time filter context available so parse failures can mention the specific option.
- Add a regression test for invalid modified-date input.

## Validation

- `cargo test`


tmccombs:
How does this compare to #2054 ?

KSHITIZ6341:
> How does this compare to #2054 ?

Thanks for taking a look. I’m going to close this PR since #2054 is the better place to continue: it was opened first, already has maintainer feedback, and covers the same underlying issue.

The only pieces from this PR that may still be useful are the date-shaped-input guard . I’ll follow along on #2054 instead.

PR #2060, opened 2026-07-07
fix: expand format placeholder followed by a literal '}'

upuddu:
A format placeholder immediately followed by a literal `}` wasn't expanded: `fd --format '{}}'` emitted a literal `{}` and dropped the trailing brace instead of producing `<path>}`. The parser fell through to a literal branch whenever a placeholder was followed by `}`; removing it lets the `}` be treated as ordinary literal text like everywhere else. Since `--exec` shares the parser, `fd -x echo '{.}}'` was affected too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fire the "search pattern contains a path separator" diagnostic for any pattern containing `/`, not just patterns that happen to name an existing directory. Preserves the legacy Windows behaviour that also flags native `\` separators when the pattern resolves to a real directory. See #1873.
 - Also fire the "search pattern contains a path separator" diagnostic for `--and` patterns, not only the primary positional pattern. `--and` patterns are matched against the file name just like the primary pattern, so a path separator in them silently returned zero results. See #1873.
 - Fix bug where passing "-" as a directory argument didn't actually search that directory, see #849 (@Sean-Kenneth-Doherty).
+- Fix `--format`/`--exec` placeholders that are immediately followed by a literal `}` (e.g. `{}}` or `{.}}`). The placeholder is now expanded with the `}` kept as trailing literal text, instead of the whole sequence being treated as fixed text.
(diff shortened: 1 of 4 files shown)


tmccombs:
What you describe is expected behavior. Currently, "}}" is the way to escape a "}" character, so the "}}" is replaced witha literal "}", and the "{" is unmatched, so not replaced. 

I haven't looked at this deeply, but I am guessing this is effectively changing the eacape syntax, which would be a breaking change.  Such a change should probably have more discussion in an issue before it is merged.

upuddu:
Thanks for taking a look, and fair point, I don't want to quietly change the escaping rules either.

Digging into the parser, I think the change is narrower than "changing the escape syntax." `{{` → `{` and `}}` → `}` are handled exactly as before, and the documented way to emit a literal `{}` (`{{}}`) is unchanged. The only strings whose output changes are a placeholder immediately followed by a *single* `}`: `{}}`, `{/}}`, `{.}}`, etc.

The reason I filed it as a bug is consistency of the lone `}`. A single, un-doubled `}` is already treated as ordinary literal text everywhere else in a format string (e.g. `--format 'a}b'` emits `a}b`). The one exception was directly after a placeholder, where the trailing `}` silently swallowed the placeholder, so `{}}` produced the literal text `{}` instead of `<path>}`. Removing that special case is what makes a lone `}` behave the same in that position as everywhere else.

That said, you're right that it's technically a behavior change for `{}}` specifically, and I'm happy to follow your process on it. Want me to open an issue laying this out (bug-vs-breaking-change, the consistency argument, and the fact that the `{{}}` escape is unaffected) so it can be decided there before merging? Happy to hold the PR until then.

tavianator:
I think the sane thing to do here is maximum munch tokenization, same as most lexers/parsers do.  Although IMO `{.}}` should be a syntax error, you should be forced to write `{.}}}`

PR #2068, opened 2026-07-23
fix: New clippy warnings

tmccombs:


hexbinoct (approved):
This looks right to me, and it would be good to get it in.

`master` is currently red on it. The CICD runs for ca51233, 41532d1 and 5864181 each fail one job, `Ensure 'cargo clippy' has no warnings`, on the 14 `useless_borrows_in_formatting` errors in `tests/tests.rs` that clippy 1.97 started reporting. Every open PR inherits that red job, so it reads as though the PR under review broke something when it did not.

The 14 changed lines are all named arguments in `format!` macros, 13 of them `abs_path = &abs_path` and one `dir = &parent_parent`. `format_args!` already takes its arguments by reference, so dropping the `&` is not a move and changes nothing at runtime. CI here is green.

PR #2071, opened 2026-07-24
tests: Remove redundant references in format! arguments

hexbinoct:
## Summary

The `Ensure 'cargo clippy' has no warnings` CI step is currently failing on master. Rust 1.97 promoted `clippy::useless_borrows_in_formatting` to flag a reference passed as a `format!` argument (the formatting machinery already borrows the value), and the test suite has fourteen `abs_path = &abs_path` style arguments that now trip it under `-D warnings`.

This drops the redundant `&` in those fourteen arguments. It is exactly what `cargo clippy --fix` produces, and it does not change behavior (the value is borrowed either way).

## Testing

Verified on Linux with the same toolchain CI uses (rustc 1.97.1):

- `cargo clippy --all-targets --all-features -- -Dwarnings` now passes
- `cargo test --all` passes
- `cargo fmt --all -- --check` passes

I used an AI assistant to help identify and mechanically apply the fix; I reviewed each change and ran the checks above.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1274,7 +1274,7 @@ fn test_absolute_path() {
             {abs_path}/one/two/three/d.foo
             {abs_path}/one/two/three/directory_foo/
             {abs_path}/symlink",
-            abs_path = &abs_path
+            abs_path = abs_path
         ),
     );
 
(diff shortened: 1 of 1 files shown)


tmccombs:
Duplicate of #2068

PR #2074, opened 2026-07-25
fix(glob): support path separators in glob patterns on Windows

### LINKED ISSUE #2067: --glob patterns with a path separator never match in --full-path mode on native Windows
On native Windows (PowerShell or cmd, no MSYS involved), glob patterns that contain a path separator never match anything in --full-path mode.

Repro on fd 10.4.2, Windows 11:

mkdir C:\t\fixture\src\foo
echo x > C:\t\fixture\src\foo\a.spec.ts
fd --glob --full-path "**/src/**/*.spec.ts" C:\t\fixture      -> no output, exit 0
fd --glob --full-path "**\src\**\*.spec.ts" C:\t\fixture      -> no output
fd --glob --full-path "**\\src\\**\\*.spec.ts" C:\t\fixture   -> no output
fd --full-path "src[/\\].*\.spec\.ts$" C:\t\fixture           -> matches (regex mode)
The regex mode line shows the candidate paths still contain backslashes at match time, while the glob derived regex only accepts /.

Cause: in glob mode build_pattern_regex (src/main.rs) extracts the regex from globset (GlobBuilder::new(pattern).literal_separator(true).build()?.regex()), where / is the literal separator. walk.rs then matches that regex against the raw native path bytes (pat.is_match(&filesystem::osstr_to_bytes(...))). globset's own matcher normalizes \ to / in Candidate::new before matching, but that step is skipped when the extracted regex is used directly.

Possibly related: [#1227](https://github.com/Hotragn/pi/issues/1227), but that report is from MSYS and mixes in shell path mangling. This repro has no MSYS in the loop.

Found while debugging [earendil-works/pi#6817](https://github.com/earendil-works/pi/issues/6817), where the find tool inherited this via fd --glob --full-path.

arimu1:
### Summary

Fixes #2067

On Windows, glob patterns containing a path separator never match in `--full-path` mode:

1. In glob mode the regex is taken straight out of `globset` (`GlobBuilder::new(pattern).literal_separator(true).build()?.regex()`), where `/` is the literal separator.
2. `walk.rs` matches that regex against the raw native path bytes, which on Windows still contain `\`. `globset`'s own matcher normalizes `\` to `/` in `Candidate::new`, but that step is skipped when the extracted regex is used directly.
3. Users on Windows also naturally write `\` in the pattern itself, where `globset` treats it as an escape character.

This change normalizes `\` to `/` in the pattern on Windows, and rewrites the generated regex so either separator is accepted at match time.

### Why the rewrite is class-aware

A naive `regex_str.replace('/', "[/\\\\]")` looks sufficient but corrupts the character classes `globset` emits for `*` and `?`:

```
pattern  **/src/**/*.spec.ts
globset  (?-u)^(?:/?|.*/)src(?:/|/.*/)[^/]*\.spec\.ts$
naive    (?-u)^(?:[/\\]?|.*[/\\])src(?:[/\\]|[/\\].*[/\\])[^[/\\]]*\.spec\.ts$
                                                          ^^^^^^^^^
```

`[^/]` becomes `[^[/\\]]`, i.e. the class `[^[/\]` followed by a *literal* `]` — so the regex only matches paths with a `]` in them. A user-supplied class such as `[/x]` is mangled the same way.

`accept_either_separator` therefore tracks whether it is inside a class and picks the right form: inside, the backslash joins the existing members (`[^/]` → `[^/\\]`); outside, the separator becomes a class of its own (`a/b` → `a[/\\]b`). Escape sequences are passed through so an escaped `\[` doesn't open a class.

### Test Plan

`accept_either_separator` is compiled on all platforms (its use is gated by `cfg!(windows)`, which still type-checks both branches), so the unit tests run everywhere rather than only on Windows runners:

- `accept_either_separator_rewrites_literal_separators`
- `accept_either_separator_rewrites_separators_inside_a_class`
- `accept_either_separator_leaves_escaped_brackets_alone`
- `rewritten_glob_regex_matches_backslash_paths` — compiles the `**/src/**/*.spec.ts` pattern from the issue and asserts the rewritten regex matches `fixture\src\foo\a.spec.ts` and `fixture/src/foo/a.spec.ts`, but not `fixture\lib\foo\a.spec.ts`.

`cargo test` (160 unit + 108 integration), `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` all pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fire the "search pattern contains a path separator" diagnostic for any pattern containing `/`, not just patterns that happen to name an existing directory. Preserves the legacy Windows behaviour that also flags native `\` separators when the pattern resolves to a real directory. See #1873.
 - Also fire the "search pattern contains a path separator" diagnostic for `--and` patterns, not only the primary positional pattern. `--and` patterns are matched against the file name just like the primary pattern, so a path separator in them silently returned zero results. See #1873.
 - Fix bug where passing "-" as a directory argument didn't actually search that directory, see #849 (@Sean-Kenneth-Doherty).
+- Fix `--glob` patterns containing a path separator never matching on Windows, where candidate paths keep their backslashes at match time, see #2067.
 
 # 10.4.2
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -215,10 +215,60 @@ fn ensure_single_search_pattern_is_not_a_path(pattern: &str) -> Result<()> {
     }
 }
 
+/// Rewrite a globset-generated regex so that either `/` or `\` is accepted wherever
+/// globset emitted a path separator. Windows candidate paths keep their backslashes
+/// at match time, so a regex that only accepts `/` never matches.
+///
+/// Separators inside a character class have to be handled differently from the ones
+/// outside it: inside, the backslash joins the existing members (`[^/]` -> `[^/\\]`),
+/// outside, the separator becomes a class of its own (`a/b` -> `a[/\\]b`).
+fn accept_either_separator(regex: &str) -> String {
+    let mut out = String::with_capacity(regex.len());
+    let mut chars = regex.chars();
+    let mut in_class = false;
+
+    while let Some(c) = chars.next() {
+        match c {
+            // Keep escape sequences intact so an escaped `[` doesn't open a class.
+            '\\' => {
+                out.push(c);
+                if let Some(escaped) = chars.next() {
+                    out.push(escaped);
+                }
+            }
+            '[' if !in_class => {
+                in_class = true;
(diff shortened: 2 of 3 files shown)


arimu1:
Heads up on the red check: `Ensure 'cargo clippy' has no warnings` is failing on `master` too ([run 30123809942](https://github.com/sharkdp/fd/actions/runs/30123809942)), not from anything in this PR. All the warnings are `clippy::useless_borrows_in_formatting` in `tests/tests.rs` (e.g. `abs_path = &abs_path` at line 1277), which a recent stable clippy started flagging. Every other check here is green, and `cargo fmt --check` plus the full test suite pass locally.

PR #2079, opened 2026-07-28
fix: reject over-large @timestamps that overflow SystemTime

### LINKED ISSUE #2077: fd panics (`overflow when adding duration to instant`) on `--changed-within/@` with a large timestamp
## Steps to reproduce

```console
$ fd --changed-within @18446744073709551615 pat
thread 'main' panicked at library/std/src/time.rs:746:31:
overflow when adding duration to instant
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ echo $?
101
```

`--changed-before` uses the same parser and crashes identically:

```console
$ fd --changed-before @18446744073709551615 .   # exit 101, same panic
```

## Root cause 

`src/filter/time.rs:44`
```rust
let timestamp_secs: u64 = s.strip_prefix('@')?.parse().ok()?;
Some(UNIX_EPOCH + Duration::from_secs(timestamp_secs))   // overflows SystemTime for large seconds
```

`fd --changed-within @<seconds>` (and `--changed-before @<seconds>`) accept a raw UNIX timestamp after `@` and compute `UNIX_EPOCH + Duration::from_secs(seconds)`. So a large value overflows `SystemTime` and fd aborts with `overflow when adding duration to instant` (exit 101).

MeGaurav4:
Fixes #2077. When a `@timestamp` value exceeds `i64::MAX` seconds, `UNIX_EPOCH + Duration::from_secs(timestamp_secs)` overflows the underlying `SystemTime` representation and panics. Adding an explicit range check returns a clean parse error instead.

AI disclosure: This PR was created with the assistance of an AI coding tool. The fix was reviewed and verified by a human.

## Verification

- `cargo test` passes (111/111)

### DIFF
--- a/src/filter/time.rs
+++ b/src/filter/time.rs
@@ -42,6 +42,9 @@ impl TimeFilter {
             )
         } else {
             let timestamp_secs: u64 = s.strip_prefix('@')?.parse().ok()?;
+            if timestamp_secs > i64::MAX as u64 {
+                return None;
+            }
             Some(UNIX_EPOCH + Duration::from_secs(timestamp_secs))
         }
     }


tmccombs:
#2082 is a better solution for this.

PR #2090, opened 2026-08-05
Validate maximum thread count

### LINKED ISSUE #2078: `--threads` accepts unsatisfiable values and panics or aborts during channel allocation and thread creation

```rust
```

On a 64-bit release build:

```console
$ fd -j 9223372036854775807 .
capacity overflow

$ echo $?
101
```

```console
$ fd -j 99999999999999 .

$ echo $?
134
```

```rust
for _ in 0..threads {
    // ...
}
```

```console
$ fd -j 200000 -x echo .
$ echo $?
134
```

Sushanth012:
## Summary

Validate explicit `--threads` values during CLI argument parsing and reject values outside `1..=65536`.

Previously, excessively large values could be accepted and later trigger panics or aborts during channel allocation or thread creation. The implicit default remains capped at 64 threads to limit startup overhead, while users on high-core-count systems can explicitly request larger values.

## Changes

- use Clap built-in ranged integer validation for `--threads`
- accept explicit values from 1 through 65536
- retain the separate 64-thread implicit default cap
- test both accepted and rejected range boundaries
- document the fix in the changelog

## Testing

- `cargo fmt --check`
- `cargo test --bin fd` (147 passed)
- `cargo clippy --bin fd -- -D warnings`
- the focused Windows integration test is blocked during test-environment setup by OS error 1314 because symlink privileges are unavailable; it fails before reaching the parser assertions

## Fixes

Closes #2078

## AI usage

OpenCode was used to inspect the review feedback and assist with the implementation and validation of the follow-up change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Add `--exact` option to match the entire filename exactly (literal, non-substring).
 
 ## Bugfixes
+- Reject `--threads` values above 64 instead of panicking during resource
+  allocation, see #2078 (@Sushanth012).
 - Sanitize control characters and bidirectional override characters in filenames
   when output goes to a terminal, to prevent terminal escape-sequence injection.
   Also reject a placeholder as the executable for `--exec-batch`, while still
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -18,6 +18,8 @@ use crate::filesystem;
 use crate::filter::OwnerFilter;
 use crate::filter::SizeFilter;
 
+const MAX_NUM_THREADS: usize = 64;
+
 #[derive(Parser)]
 #[command(
     name = "fd",
@@ -553,7 +555,7 @@ pub struct Opts {
 
     /// Set number of threads to use for searching & executing (default: number
     /// of available CPU cores)
-    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = str::parse::<NonZeroUsize>)]
+    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = parse_num_threads)]
     pub threads: Option<NonZeroUsize>,
 
     /// Milliseconds to buffer before streaming search results to console
@@ -792,11 +794,23 @@ fn default_num_threads() -> NonZeroUsize {
     let fallback = NonZeroUsize::MIN;
     // To limit startup overhead on massively parallel machines, don't use more
     // than 64 threads.
-    let limit = NonZeroUsize::new(64).unwrap();
+    let limit = NonZeroUsize::new(MAX_NUM_THREADS).unwrap();
 
     std::thread::available_parallelism()
         .unwrap_or(fallback)
         .min(limit)
+}
+
+fn parse_num_threads(arg: &str) -> Result<NonZeroUsize, String> {
+    let threads = arg
+        .parse::<NonZeroUsize>()
+        .map_err(|error| error.to_string())?;
+    if threads.get() > MAX_NUM_THREADS {
+        return Err(format!(
+            "the number of threads cannot exceed {MAX_NUM_THREADS}"
+        ));
+    }
+    Ok(threads)
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, ValueEnum)]
@@ -967,5 +981,17 @@ fn ensure_current_directory_exists(current_directory: &Path) -> anyhow::Result<(
         Err(anyhow!(
             "Could not retrieve current directory (has it been deleted?)."
         ))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse_num_threads;
+
+    #[test]
+    fn number_of_threads_is_bounded() {
+        assert_eq!(parse_num_threads("64").unwrap().get(), 64);
+        assert!(parse_num_threads("65").is_err());
+        assert!(parse_num_threads("9223372036854775807").is_err());
     }
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -4,8 +4,8 @@ use std::time::Duration;
 
 use anyhow::anyhow;
 use clap::{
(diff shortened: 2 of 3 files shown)


src/cli.rs
```diff
@@ -18,6 +18,8 @@ use crate::filesystem;
 use crate::filter::OwnerFilter;
 use crate::filter::SizeFilter;
 
+const MAX_NUM_THREADS: usize = 64;
```
tmccombs: Why 64? 

The number of threads can be significantly higher than that. Especially on hosts with a large number of cpu cores.

src/cli.rs
```diff
@@ -553,7 +555,7 @@ pub struct Opts {
 
     /// Set number of threads to use for searching & executing (default: number
     /// of available CPU cores)
-    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = str::parse::<NonZeroUsize>)]
+    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = parse_num_threads)]
```
tmccombs: why not use clap's built-in support for ranges?

tmccombs (changes requested):
While values below 64 are probably safe, I think that is probably too low for a hard-coded limit. 

A fundamental problem here is that the maximum number of threads that won't result in an error is basically impossible to know. 

It depends on a variety of factors, including OS, amount of available memory, ulimit settings for the current process, etc.  I think something like say 2^16 is probably reasonable to use as a limit, but 64 is definitely too low. That means that on, say, a 128 core computer, you can't even use all of your cores.

Sushanth012:
Thanks for the review @tmccombs — that's a fair point, 64 is way too low as a hard cap and would block legitimate use on high-core-count machines. I'll raise \MAX_NUM_THREADS\ to 2^16 (65536) so it still catches absurd values that would panic during channel/thread allocation without limiting real hardware, and update the tests accordingly.

Sushanth012:
Pushed b001153 addressing your feedback @tmccombs:

- \MAX_NUM_THREADS\ is now 2^16 (65536), so the validation only rejects values that would obviously fail later during channel/thread allocation — a 128-core machine can now use all of its cores.
- The implicit default (when \--threads\ is not given) still caps at 64 via a separate \DEFAULT_NUM_THREADS_LIMIT\ constant, preserving the startup-overhead behavior on massively parallel machines. Happy to drop that cap too if you'd prefer defaults to follow \vailable_parallelism\ unbounded.

Tests updated: the parser unit tests now cover 65/128/65536 (ok) and 65537+ (rejected), and the integration test asserts \--threads=65\ succeeds and \--threads=65537\ fails.

Note: one integration test (\	est_number_parsing_errors\) fails locally on my Windows machine during test-env setup with OS error 1314 (symlink privilege) — this reproduces identically without my changes, so it's environmental; CI should be authoritative here.

Sushanth012:
Thanks for the review. I addressed the hard-cap value earlier but missed the parser simplification and left the PR description stale. I will replace the custom `parse_num_threads` path with Clap built-in range validation while preserving the nonzero invariant, update the affected tests and description, then push and report the validation results for re-review.

src/cli.rs
Sushanth012: Addressed in 8755068. Explicit thread counts are accepted through 65536, while only the implicit default remains capped at 64. The full CI matrix passes.

src/cli.rs
Sushanth012: Addressed in 8755068. The custom parse_num_threads function is removed and --threads now uses Clap RangedU64ValueParser with the 1..=65536 range. Parser boundary tests, formatting, Clippy, the 147 unit tests, and the full CI matrix pass. Ready for re-review.

src/cli.rs
```diff
@@ -753,7 +772,9 @@ impl Opts {
     }
 
     pub fn threads(&self) -> NonZeroUsize {
-        self.threads.unwrap_or_else(default_num_threads)
+        self.threads.map_or_else(default_num_threads, |threads| {
+            NonZeroUsize::new(threads).expect("--threads is validated as nonzero")
```
tmccombs, line 776: I think "--threads cannot be zero" would be a better message here

src/cli.rs
```diff
@@ -969,3 +990,21 @@ fn ensure_current_directory_exists(current_directory: &Path) -> anyhow::Result<(
...
+    use super::{MAX_NUM_THREADS, Opts};
+    use clap::Parser;
+
+    #[test]
+    fn number_of_threads_is_bounded() {
+        assert_eq!(
+            Opts::try_parse_from(["fd", "--threads", "65536"])
+                .unwrap()
+                .threads,
+            Some(MAX_NUM_THREADS)
```
tmccombs, line 1005: I think it would be more clear if we either used the same number here
```suggestion
            Some(65536)
```

or we specifically used `MAX_NUM_THREADS.to_string()` as the argument above. 

Actually, the latter might be better, since then the test wouldn't need to be updated if `MAX_NUM_THREADS` was changed later. (and maybe use `(MAX_NUM_THREADS + 1).to_string()` on line 1008.)

tests/tests.rs
```diff
@@ -2659,6 +2659,8 @@ fn test_number_parsing_errors() {
     te.assert_failure(&["--threads=a"]);
     te.assert_failure(&["-j", ""]);
     te.assert_failure(&["--threads=0"]);
+    let _ = te.assert_success_and_get_output(".", &["--threads=65"]);
```
tmccombs, line 2662: I don't think this belongs in this test, since it isn't an error. I'd be ok just removing it altogether.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - Add `--exact` option to match the entire filename exactly (literal, non-substring).
 
 ## Bugfixes
+- Reject `--threads` values above 64 instead of panicking during resource
```
tmccombs, line 8: This is no longer correct.

tmccombs (changes requested):

Issue #1985, opened 2026-04-26
[BUG] `fd` cannot handle arbitrary deep directories like GNU `find`


collinfunk:
### Checks

- [x] I have read the troubleshooting section and still think this is a bug.

### Describe the bug you encountered:

One of the nice features of GNU `find` and many other GNU programs is that they avoid [arbitrary limits](https://www.gnu.org/prep/standards/standards.html#Writing-Robust-Programs) whenever possible. As a result, GNU `mkdir` can create arbitrarily deep directories which `find` can then visit:

```
$ mkdir -p $(yes a/ | head -n  32768 | tr -d '\n')
$ find a | wc -l
32768
```

This is not the case with `fd`:

```
$ ./target/release/fd a | wc -l
2062
```

Using `strace`, we can see that `fd` constructs the file names in full each time and calls `statx(2)` instead of operating on file descriptors. This means `fd` reaches `PATH_MAX` limitations:

```
$ strace -ff -e trace=statx ./target/release/fd -j 1 '' a  2>&1 | grep '= -1'
[...]
[pid 344540] statx(AT_FDCWD, "[long file name trimmed]"..., AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7fdb3c3fbcf0) = -1 ENAMETOOLONG (File name too long)
[...]
```

It also gives a misleading non-zero exit status after failing to list all of the file names:

```
$ ./target/release/fd -j 1 '' a > /dev/null
$ echo $?
0
```

### What version of `fd` are you using?

fd 10.4.2

### Which operating system / distribution are you on?

```shell
$ uname -srm
Linux 6.18.4-200.fc43.x86_64 x86_64
$ lsb_release -a
LSB Version:	n/a
Distributor ID:	Fedora
Description:	Fedora Linux 43 (Workstation Edition)
Release:	43
Codename:	n/a
```

tmccombs:
This would be good to address, but it may be quite difficult to do so, since we currently use `walkdir` for traversing the filesystem (via the `ignore` crate), which [can't currently handle long paths](https://github.com/BurntSushi/walkdir/issues/23). 

Possibly there is a way to work around that, but it probably woudn't be easy, and may add a lot of complexity.

hiSandog:
[...]

PR #2095, opened 2026-08-15
Exit non-zero on filesystem traversal errors

ezekiel06:
fixes the exit-code part of #1985; used Claude to investigate and write the fix, reviewed and tested it myself

### DIFF
--- a/src/exec/job.rs
+++ b/src/exec/job.rs
@@ -26,6 +26,9 @@ pub fn job(
                 if config.show_filesystem_errors {
                     print_error(err.to_string());
                 }
+                // A traversal error means the search was incomplete; reflect that in
+                // the exit code even if it isn't printed (i.e. without --show-errors).
+                ret = merge_exitcodes([ret, ExitCode::GeneralError]);
                 continue;
             }
         };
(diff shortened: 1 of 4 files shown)


src/exec/job.rs
```diff
@@ -26,6 +26,9 @@ pub fn job(
                 if config.show_filesystem_errors {
                     print_error(err.to_string());
                 }
+                // A traversal error means the search was incomplete; reflect that in
+                // the exit code even if it isn't printed (i.e. without --show-errors).
+                ret = merge_exitcodes([ret, ExitCode::GeneralError]);
```
tmccombs: It might be worth having a separate exit code for this.

tests/tests.rs
```diff
@@ -2491,6 +2491,50 @@ fn test_owner_root() {
...
+    // A directory search should complete successfully and report success when there
+    // are no filesystem errors.
+    te.assert_success_and_get_output(".", &[]);
+
+    // Remove read+execute permissions so that `fd` can't list this directory's
+    // contents, which surfaces as a traversal (`WorkerResult::Error`) during the walk.
+    fs::set_permissions(&unreadable_dir, fs::Permissions::from_mode(0o000))
+        .expect("could not change permissions for test");
+
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
tmccombs: What is the purpose of the catch_unwind here?

Issue #2096, opened 2026-08-15
Windows: negated attribute-hidden files still leak without --hidden


SomSamantray:
On Windows, fd's hidden contract also covers files with `FILE_ATTRIBUTE_HIDDEN` whose names lack a leading dot (see existing `test_hidden_file_attribute`). The hidden-entry guard added for #1266 in the walker closure (`src/walk.rs`) is name-based only: it checks for a leading dot. A `.gitignore` negation pattern can therefore still leak an attribute-hidden file without `--hidden` on Windows — the ignore crate's hidden filter is bypassed by the negation (the root cause of #1266) and the new guard does not fire for a non-dot name.

Suggested fix: on Windows, extend the guard to also skip entries whose metadata carries `FILE_ATTRIBUTE_HIDDEN` (e.g. `#[cfg(windows)]` + `std::os::windows::fs::MetadataExt::file_attributes()`), using the entry's cached metadata to avoid extra syscalls, mirroring the ignore crate's `is_hidden_entry` semantics.

Evidence:
- `src/walk.rs:515-519`: `if config.ignore_hidden && file_name.is_some_and(|name| name.as_encoded_bytes().first() == Some(&b'.')) { return WalkState::Skip; }`
- `tests/tests.rs` (`test_hidden_file_attribute`): Windows-only test proving fd's hidden handling includes `FILE_ATTRIBUTE_HIDDEN` files
- ignore crate 0.4.x `is_hidden_entry`: leading dot plus, on Windows, the hidden attribute

Source: branch `fix/hidden-files-negated-gitignore`, head `3bd8980` (PR link to be back-filled).

Metadata: Severity: P3 | Confidence: 100 | Reviewer(s): correctness, testing | Finding ID: src-walk-rs 512-518 hidden check only covers dot-prefix names

SomSamantray:
PR carrying this change: https://github.com/sharkdp/fd/pull/2097

tmccombs:
This is essentially a duplicate of #1266 specifically for windows, assuming a PR was merged, that was not merged. 

I'm going to close this for now. When (and if) that bug is addressed it should be done in a way that works on windows as well.

PR #2097, opened 2026-08-15
fix: do not list hidden files negated in .gitignore without --hidden

### LINKED ISSUE #1266: [BUG] Hidden files are wrongly listed if negated in .gitignore
### Checks

Hidden files that are negated / listed as exclusions in a `.gitignore` file are shown, even when not using `--hidden`:
```
> mkdir fd-issue
> cd fd-issue
> git init
> echo '!.gitignore' > .gitignore
> fd
.gitignore
```

The `.gitignore` file is not listed. This is a problem in the `ignore` crate, as it can be reproduced with `rg --files` as well.

8.7.0

```shell
Linux 5.4.0-139-generic x86_64
```

SomSamantray:
Without `--hidden`, `fd` now never lists dot-prefixed hidden entries, even when a `.gitignore` pattern negates them. Previously, a negation such as `!.gitignore` bypassed the ignore crate's hidden filter and the file was listed; negated hidden directories were traversed as well. Hidden directories are now pruned, so their contents are not searched. With `--hidden`, behavior is unchanged.

The check is name-based and runs before the `--min-depth` guard, so the hidden contract holds regardless of depth configuration.

Fixes #1266

One known limitation: on Windows, files hidden via `FILE_ATTRIBUTE_HIDDEN` whose names lack a leading dot are not covered by the name-based check. The follow-up is tracked separately (see Related).

Related: #2096

Validation: new regression test `test_hidden_negated_gitignore` fails on master and passes here. It covers self-negated hidden files, negated hidden directories, and `--min-depth` interplay. Full suite: 157 unit + 109 integration tests pass; `cargo fmt` and `cargo clippy --all-targets --all-features -- -Dwarnings` are clean. Manual repro of the issue verified against a local build.

Session-settled decisions carried from planning: no third-party branding on this PR (user-directed); fd contribution guidelines followed (user-directed).

AI disclosure: This PR was produced with the assistance of an AI coding assistant (fix implementation, regression test, changelog entry, and PR scaffolding) and was reviewed by the human author before submission.

### DIFF
--- a/docs/residual-review-findings/3bd8980b5250d0e969adf36b45b64d784ca868f8.md
+++ b/docs/residual-review-findings/3bd8980b5250d0e969adf36b45b64d784ca868f8.md
@@ -1,0 +1,12 @@
+# Residual Review Findings
+
+Run context: LFG pipeline for sharkdp/fd issue #1266 (hidden files wrongly listed when negated in `.gitignore`). Plan: `docs/plans/2026-08-15-001-fix-hidden-gitignore-negation-plan.md`. Review run: `ce-code-review mode:agent`, run id `20260815-212237-5e098adb`, artifact at `/tmp/compound-engineering-501/ce-code-review/20260815-212237-5e098adb/`.
+
+## Residual Review Findings
+
+- P3 — `src/walk.rs:515` — Hidden check only covers dot-prefix names, so attribute-hidden entries with non-dot names still leak on Windows when un-ignored by a gitignore negation — tracker ticket: https://github.com/sharkdp/fd/issues/2096
+
+## Not applied in step 5
+
+- #1 (above): fix is a behavior change (per-entry attribute metadata checks on Windows) conflicting with the plan's documented name-based, no-syscall decision (KTD1); documented as a limitation in the plan's Assumptions and in the PR description. Filed as issue #2096.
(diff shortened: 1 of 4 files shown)


docs/residual-review-findings/3bd8980b5250d0e969adf36b45b64d784ca868f8.md
tmccombs, line 1: This file should not be checked in

tmccombs (commented):
The current behavior is intentional. 

Hidden files are not supposed to be shown unless `--hidden` is used.  Negation in ignore files is intended to negate the ignore mechanism, not force the file to be shown even if it otherwise wouldn't be.

tmccombs:
Oh, I'm sorry, I misread the description. 

However, I still think this needs work. 

I'll reopen and submit a needs-work review.

tmccombs (changes requested):
First of all, I think it would be better if this was fixed in the ignore crate itself, or at least if the ignore create provided an option to do this.  See https://github.com/BurntSushi/ripgrep/issues/3514.  Although if that is unlikely to happen, or would take to long, I'm fine with something like this. 

Secondly, I think we should do this in a way that works for windows as well.

PR #2098, opened 2026-08-17
docs: update Guix package link

JiuYue0820:
## Problem

The Guix install section linked to a version-pinned page
`https://guix.gnu.org/en/packages/fd-8.1.1/`. That URL no longer serves the
package page; GET follows to the packages index (`https://packages.guix.gnu.org/`)
instead of the fd package.

The current package page is `https://packages.guix.gnu.org/packages/fd/`.

Verified:

    GET https://guix.gnu.org/en/packages/fd-8.1.1/ -> 200, final URL is https://packages.guix.gnu.org/
    GET https://packages.guix.gnu.org/packages/fd/ -> 200, stays on the fd package page

## Solution

Link the unversioned Guix package page so the README does not go stale
when fd's Guix version changes.

## Verification

GET `https://packages.guix.gnu.org/packages/fd/` returns 200.

Documentation-only README change. I did not add a CHANGELOG entry because
this does not change fd's runtime behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -676,7 +676,7 @@ winget install sharkdp.fd
 
 ### On GuixOS
 
-You can install [the fd package](https://packages.guix.gnu.org/packages/fd/) from the official repo:
+You can install [the fd package](https://packages.guix.gnu.org/packages/fd/10.4.2/) from the official repo:
 ```
 guix install fd
 ```
--- a/README.md
+++ b/README.md
@@ -676,7 +676,7 @@ winget install sharkdp.fd
 
 ### On GuixOS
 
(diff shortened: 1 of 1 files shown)


tmccombs (changes requested):
That link gives me a page saying that fd is an umknown package

FirmaSpring:
The unversioned https://packages.guix.gnu.org/packages/fd/ page currently shows Unknown package, even though Guix still ships fd.

The package is still defined as `fd` 10.4.2 in rust-apps.scm (crate name `fd-find`). The versioned page https://packages.guix.gnu.org/packages/fd/10.4.2/ loads, so the README now points there. `guix install fd` is unchanged.

PR #2099, opened 2026-08-19
Clamp --threads to a sane maximum to avoid panic/abort

### LINKED ISSUE #2078: `--threads` accepts unsatisfiable values and panics or aborts during channel allocation and thread creation

nagendramohan:
fd -j N parses N into an unbounded thread count. Pathological values then either overflow the work-channel capacity (bounded(2 * threads), 'capacity overflow', exit 101 on every run) or exhaust thread creation with --exec ('failed to spawn thread', abort).

Clamp the resolved thread count to max(available_parallelism, 64) — more worker threads than the machine can run give fd no benefit, and this never clamps below the hardware parallelism nor below the existing 64-thread default cap, so realistic values are unaffected.

Adds regression tests (verified failing before the fix) and a CHANGELOG entry. Closes #2078.

Closes #2078.

fd -j N parses N into an unbounded thread count, so pathological values crash:
- fd -j 9223372036854775807 . → capacity overflow (exit 101) — the work channel is bounded(2 * threads). Happens on every run.
- fd -j 200000 -x echo . → failed to spawn thread → abort (exit 134) with --exec.

Fix: clamp the resolved thread count to max(available_parallelism(), 64). More worker threads than the machine can run provide no benefit for fd's workload, and this never clamps below the hardware parallelism nor below the existing 64-thread default cap (default_num_threads), so realistic values are unaffected. Happy to switch to a hard error or a different bound if you'd prefer — flagging the policy choice.

Adds regression tests (verified failing before the fix) and a CHANGELOG entry. cargo fmt --check, cargo clippy --all-targets, and the full test suite pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Also fire the "search pattern contains a path separator" diagnostic for `--and` patterns, not only the primary positional pattern. `--and` patterns are matched against the file name just like the primary pattern, so a path separator in them silently returned zero results. See #1873.
 - Fix bug where passing "-" as a directory argument didn't actually search that directory, see #849 (@Sean-Kenneth-Doherty).
 - Fix panic when `--changed-before`/`--changed-within` is given an out-of-range `@` Unix timestamp; the value is now rejected gracefully, see #2081 (@nikolauspschuetz).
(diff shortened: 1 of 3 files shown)


nagendramohan:
Closing as a duplicate of #2090.

PR #2120, opened 2026-09-04
fix: escape control characters in error messages

### LINKED ISSUE #2119: [BUG] Sanitize gap for error messages
### Checks

10.5.0

```shell
Arch
```

curious-rabbit:
Fixes #2119

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,13 @@
+use std::io::IsTerminal;
+
+use crate::sanitize::sanitize_message;
+
 pub fn print_error(msg: impl std::fmt::Display) {
-    eprintln!("[fd error]: {msg}");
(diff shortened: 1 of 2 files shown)


tmccombs (commented):
Would it be better to just always allow newlines? 

Newlines aren't dangerous in the sense of sending control commands to the terminal. Although then can potentially be confusing.

curious-rabbit:
Newlines are already allowed in messages, with sanitize_message. They stay escaped in printed paths on purpose, since one file with a newline would show as two result lines. 
in sanitize_for_terminal this is kept to prevent spoofing.

tmccombs:
Fixed by https://github.com/sharkdp/fd/pull/2135

PR #2123, opened 2026-09-05
Include files modified exactly at the starting timestamp for --changed-after and its aliases

### LINKED ISSUE #2122: [BUG] file not found when using --changed-after / --changed-before
### Checks

Hi guys, 

and thanks for the great fd.

However, I believe there's an issue when looking for files created a certain day, when at midnight. Or hour, at start of hour, etc.

The compare conditions, now, I believe are

--changed-after .....    >mtime
--changed-before .... <mtime

or something along those line. That means that a file created at midnight, eg with touch file -d <date> is missed when you search: what files are created <date>?

Same goes for: what files are created 07:xx, 08:xx?,  the files with time 07:00, 08:00 are not found.

The lower bound ought to be >= mtime.

BR/ Ingemar

_No response_

10.5.0

```shell
Linux, CachyOS
```

pederbe:
## Summary

Include files modified exactly at the starting timestamp for `--changed-after` and its aliases. Starting timestamp is now included; end remains excluded.

Added boundary tests and updated help, man page, and changelog.

Fixes #2122

## Testing

- All 148 unit tests and direct CLI checks passed.
- Formatting passed. Clippy passed with existing unused-code warnings suppressed.
- All 95 integration tests passed, including the new timestamp-boundary regression.

## AI assistance

I used Codex to help develop and test this change. I personally reviewed the work and wrote this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 -
 
 ## Bugfixes
+- Include files modified exactly at the starting timestamp for `--changed-after` and its aliases, see #2122 (@pederbe).
 - Don't incorrectly escape newlines in error messages, see #2104
 - Restore jemalloc as default allocator on supported systems
 
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -355,7 +355,7 @@ tebibytes
 .TP
 .BI "\-\-changed-within " date|duration
 Filter results based on the file modification time.
-Files with modification timestamps greater than the argument will be returned.
+Files with modification timestamps greater than or equal to the argument will be returned.
 The argument can be provided as a duration (\fI10h, 1d, 35min\fR) or as a specific point
 in time as full RFC3339 format with time zone, as a date or datetime in the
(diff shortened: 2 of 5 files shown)


tmccombs (changes requested):
We need to make a decision on whether we actually want to make any changes for #2122, and if so, what they should be, before this can be merged.

pederbe:
> We need to make a decision on whether we actually want to make any changes for #2122, and if so, what they should be, before this can be merged.

Yeah, I'll leave it open but won't push anything else until #2122 is settled.

One thing now that I've been in the code again: I think tavianator is right that strict after/before is the normal reading, and this PR ignores that by making the start bound inclusive in every case. So even if you do want a change here, this probably isn't it.

The precision idea seems like the only version worth building. Treat `--changed-after 2026-09-01` as the whole day, so >= midnight, but keep a full timestamp strict. Durations and exact timestamps stay as they are, and only date-only arguments change, which is the case in the issue. The cost is that after and before stop being exact complements for coarse arguments.

PR #2132, opened 2026-09-09
fix: sanitize control characters in error messages again

### LINKED ISSUE #2119: [BUG] Sanitize gap for error messages

jabrailkhalil:
## Summary

Error output lost its terminal-escape filtering in a3942db: while fixing newline rendering, the whole sanitization filter was dropped. File names and command placeholders in `[fd error]:` lines can carry control characters, so an injected filename reached the terminal unchanged again.

Changes:

- `src/sanitize.rs`: refactored the escaping loop into `sanitize_inner` and added `sanitize_for_terminal_except_newline`, which escapes every control/format character except `\n`. The strict `sanitize_for_terminal` (which also escapes newlines) is unchanged and remains the tool for file names.
- `src/error.rs`: `print_error` now routes the message through the new variant, so control characters are escaped while multi-line error chains keep their line breaks (the original purpose of a3942db).

## Tests

- `src/sanitize.rs`: new tests for the except-newline variant (newlines kept, other controls still escaped, borrow passthrough for safe input).
- `src/error.rs`: new `format_error` tests covering an OSC-injection payload mixed with a newline and plain-text passthrough.

## Verification

```bash
cargo test --bin fd           # 151 passed
cargo test                    # 151 unit + 92 integration passed;
                              # 2 pre-existing Windows-only failures:
                              # test_list_details needs GNU ls,
                              # test_exec_with_separator needs a POSIX echo
cargo fmt -- --check          # clean
cargo clippy --all-targets    # clean (2 pre-existing dead-code warnings)
```

Fixes #2119.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,28 @@
+use crate::sanitize::sanitize_for_terminal_except_newline;
+
 pub fn print_error(msg: impl std::fmt::Display) {
-    eprintln!("[fd error]: {msg}");
+    eprintln!("{}", format_error(&msg.to_string()));
+}
+
+/// Build the `[fd error]: ...` line, escaping terminal control characters in
+/// `msg` while preserving newlines so multi-line errors stay readable.
+fn format_error(msg: &str) -> String {
+    format!("[fd error]: {}", sanitize_for_terminal_except_newline(msg))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn format_error_escapes_control_chars_but_keeps_newlines() {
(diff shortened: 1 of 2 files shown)


jabrailkhalil:
Merged current `master` in `4097282` and carried the error-message newline behavior into the upstream sanitization refactor. The new regression test verifies that terminal control characters are escaped while line breaks remain intact.

`cargo fmt --check` passed. The focused `cargo test` could not link locally because this Windows host lacks the MSVC `link.exe`; project CI will provide the authoritative test result.

jabrailkhalil:
Fixed the only CI finding in cfbd450: the new regression test now uses io::Error::other, as required by the repository's clippy configuration.

Issue #1761, opened 2025-08-01
Allow to separate -x outputs with the match name


Porkepix:
Currently when using `fd -x` to execute a command on every match of what we look for with `fd`, if it's a command where the output and where does it come from matters, we should be able to know which result is associated with this output with a separator between them (ideally with a toggle on/off, probably defaulting to off to keep the default as it currently is).

dogsc729:
Hi! 👋 I’m currently working on this feature.
Here’s the separator format I came up with — I’d love to hear your suggestions or preferences:
```
---- fd: cat /path/to/file/test1.txt ----
hello
---- fd: cat /path/to/file/test2.txt ----
world
```

This behavior is controlled by a new flag, `--exec-print-separator`. The separator will only be printed when this flag is explicitly set.

This is also my first time contributing to an open-source project, and I’m really excited to get involved! 🚀

Porkepix:
> Hi! 👋 I’m currently working on this feature. Here’s the separator format I came up with — I’d love to hear your suggestions or preferences:
> 
> ```
> ---- fd: cat /path/to/file/test1.txt ----
> hello
> ---- fd: cat /path/to/file/test2.txt ----
> world
> ```
> 
> This behavior is controlled by a new flag, `--exec-print-separator`. The separator will only be printed when this flag is explicitly set.
> 
> This is also my first time contributing to an open-source project, and I’m really excited to get involved! 🚀

First of all, thanks for tackling on the issue!

Format is fine to me, even just the file name followed by a semicolon would have been fine to me! I feel like the flag wording could be a little bit confusing as I read it in a way that suggest you could pick what the separator will be. Take this with a grain of salt though: English isn't my native language and it's completely fine to me this way (but @sharkdp's input would be welcome for flags choice and homogeneity).

I just thought some kind of whitelist of commands that would toggle the flag automatically by default (cat, ls, …) would be welcome but have a relevant list would prove to be difficult I think, so maybe not that much a good idea. But maybe get this always on or behind a flag to toggle this off would be useful?

dogsc729:
@Porkepix Thanks a lot for the feedback! 🙏

I see what you mean about the flag name — something like `--exec-enable-separator` might make the intent clearer. I also agree that having the option the other way around (separator on by default, with a flag to disable) could work too. Maybe best to hold off until we hear what @sharkdp thinks about naming and defaults for consistency.

tmccombs:
FWIW, you can accomplish something like this today with:

```
[...]

PR #2133, opened 2026-09-10
docs(exec): clarify command ordering and output grouping for -x/--exec

pederbe:
## Summary

The previous manpage explained the order of repeated -x commands for a single search result, but then added a broader note that command execution order isn't guaranteed. That could read as contradicting the per-result sequencing it had just described. The --help text had a similar broad caveat without explaining repeated -x behavior at all.

The updated wording clarifies three points:

- For a single search result, commands supplied via repeated -x run sequentially in the order given.
- When running in parallel, fd buffers command output for each result and prints it together, without interleaving it with command output from other results.
- The existing non-determinism warning is about the order in which different search results are processed, not about commands within one result, and it still holds even with --threads=1.

The long help text also now explains repeated -x usage, including the need to terminate each command except the last with a `;` argument (escaped as `\;` in most shells).

This is a documentation-only change. No execution logic, flags, or behavior changed. It does not add the separator flag requested in the issue.

## Related issue

https://github.com/sharkdp/fd/issues/1761

## Testing

- `cargo test --locked --all-features`: 147 unit and 94 integration tests   passed on Windows, with no failures. Unix-only tests are excluded from this Windows run.
- `cargo fmt --all --check` and `cargo clippy --locked --all-targets --all-features`: both passed; clippy showed two pre-existing dead-code warnings in unrelated Windows test helpers.
- `git diff --check`: passed.
- A local probe exercised 8 files, 2 commands, at 1 and 4 threads, 3 runs each, checking per-result command order (via a marker dependency) and output grouping independently for stdout and stderr.
- `fd.exe --help` inspected manually for correct wrapping/spacing of the new paragraph.
- Updated manpage section rendered via `groff -Kutf8 -man -Tutf8` (WSL), passed with no stderr warnings.

## AI disclosure

I used Codex to assist coding and testing. I personally reviewed all work and verified testing.

### DIFF
--- a/doc/fd.1
+++ b/doc/fd.1
@@ -448,13 +448,13 @@ Note that all subsequent positional arguments are considered to be arguments to 
 It is therefore recommended to place the \-x/\-\-exec option last. Alternatively, you can supply
 a ';' argument to end the argument list and continue with more fd options.
 Most shells require ';' to be escaped: '\\;'.
-This option can be specified multiple times, in which case all commands are run for each
-file found, in the order they are provided. In that case, you must supply a ';' argument for
-all but the last commands.
+This option can be specified multiple times. Commands for the same search result run sequentially
+in the order they are given. In that case, you must supply a ';' argument for all but the last command.
+When running in parallel, fd buffers command output and prints it together for each search result,
+without interleaving it with command output for other results.
 
-If parallelism is enabled, the order commands will be executed in is non-deterministic. And even with
---threads=1, the order is determined by the operating system and may not be what you expect. Thus, it is
-recommended that you don't rely on any ordering of the results.
+The order in which different search results are processed and their output is printed is not
+guaranteed, even with --threads=1.
 
 Before executing the command, any placeholder patterns in the command are replaced with the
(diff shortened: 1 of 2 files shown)


doc/fd.1
```diff
@@ -448,13 +448,13 @@ Note that all subsequent positional arguments are considered to be arguments to
 It is therefore recommended to place the \-x/\-\-exec option last. Alternatively, you can supply
 a ';' argument to end the argument list and continue with more fd options.
 Most shells require ';' to be escaped: '\\;'.
-This option can be specified multiple times, in which case all commands are run for each
-file found, in the order they are provided. In that case, you must supply a ';' argument for
-all but the last commands.
+This option can be specified multiple times. Commands for the same search result run sequentially
```
tmccombs: I think "Commands for the same search result" is confusing.  I think I know what you mean, but it isn't very clear language.

tmccombs (changes requested):

doc/fd.1
pederbe: Fair point. I revised this section in my latest commit. It should be good to go.

tmccombs (approved):

PR #2134, opened 2026-09-13
fix: validate --threads value to prevent channel-capacity overflow panic

### LINKED ISSUE #2078: `--threads` accepts unsatisfiable values and panics or aborts during channel allocation and thread creation

koopatroopa787:
Fixes #2078.

## Problem

`fd -j N` accepted any `NonZeroUsize` without bounds checking. The value was then used directly in `walk.rs`:

```rust
```

For values where `2 * N` overflows `usize` (e.g. `fd -j 9223372036854775807 .`) this produced a panic with exit 101 on release builds rather than a user-facing error.

## Fix

Add a custom value-parser `parse_thread_count` that checks `N.checked_mul(2).is_some()` before accepting the value. Oversized values are now rejected at argument parsing with a clear error message:

```
error: invalid value '9223372036854775807' for '--threads <num>': 9223372036854775807 threads would overflow the channel capacity
```

## Changes

- `src/cli.rs`: Replace `value_parser = str::parse::<NonZeroUsize>` with a new `parse_thread_count` function that validates the overflow guard.
- `tests/tests.rs`: Extend `test_number_parsing_errors` to cover `--threads=usize::MAX`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -554,7 +554,7 @@ pub struct Opts {
 
     /// Set number of threads to use for searching & executing (default: number
     /// of available CPU cores)
-    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = str::parse::<NonZeroUsize>)]
+    #[arg(long, short = 'j', value_name = "num", hide_short_help = true, value_parser = parse_thread_count)]
     pub threads: Option<NonZeroUsize>,
 
     /// Milliseconds to buffer before streaming search results to console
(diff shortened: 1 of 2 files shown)


tmccombs (changes requested):
all the copilot comments are relevant

PR #2135, opened 2026-09-14
fix: Better sanitizing

### LINKED ISSUE #2119: [BUG] Sanitize gap for error messages

tmccombs:
This refactors the sanitization code somewhat, to provide a couple benefits:

- Ensure we sanitize output in error messages (#2119), while not escaping newlines from error messages.
- Avoids some unnecessary allocations and string copies
-  Makes `print_error` a little nicer to use. As a macro, it avoids
      needing a separate call to `format!`.

The main downside is added complexity.

Fixes: #2119


src/exec/command.rs
```diff
@@ -101,14 +100,14 @@ pub fn execute_commands<I: Iterator<Item = io::Result<Command>>>(
 pub fn handle_cmd_error(cmd: Option<&Command>, err: io::Error) -> ExitCode {
     match (cmd, err) {
         (Some(cmd), err) if err.kind() == io::ErrorKind::NotFound => {
-            print_error(format!(
-                "Command not found: {}",
-                cmd.get_program().to_string_lossy()
-            ));
+            print_error!(
+                "Command not found: {:?}",
```
tmccombs, line 104: Using the debug format here is convenient, and I think safe (at least with the current implementation, although I think it is unlikely that they will change it not to escape non-printable characters in the future). However, one downside is that the format is inconsistent with our other sanitize function. Also, if we used our Display wrapper, we could potentially get away without allocating a string even if it isn't valid utf-8, by escaping the raw non-ascii bytes from `as_encoded_bytes()` .

src/cli.rs
```diff
@@ -711,10 +710,12 @@ impl Opts {
...
                     Some(self.normalize_path(path))
                 } else {
-                    print_error(format!(
-                        "Search path '{}' is not a directory.",
-                        path.to_string_lossy()
-                    ));
+                    // We use debug for the path to make it more readable if it has special characters
+                    // Should we use a more reliable escape?
+                    print_error!(
+                        "Search path {:?} is not a directory.",
```
tmccombs, line 716: Same thing here about using the debug format

tavianator (approved):

PR #2137, opened 2026-09-17
fix: escape control characters in argument error messages

curious-rabbit:
When a command-line argument is invalid, fd wrote the argument into the error message and printed it without escaping control characters and a crafted argument could run terminal escape sequences.

Example: a script runs fd "$INPUT" with an untrusted pattern. An invalid regex or glob holding an escape sequence is echoed in the error, and the terminal executes it.


src/main.rs
```diff
@@ -78,7 +78,15 @@ fn main() {
 }
 
 fn run() -> Result<ExitCode> {
-    let opts = Opts::parse();
+    let opts = match Opts::try_parse() {
+        Ok(opts) => opts,
+        Err(err) if err.use_stderr() => {
+            // escape control characters in the clap error before printing
+            eprint!("{}", sanitize::sanitized_error(&err.render().to_string()));
```
tmccombs, line 85: This will break ansi colors in the error message

src/sanitize.rs
```diff
@@ -70,6 +70,14 @@ pub fn sanitize_for_term(raw: &str, is_terminal: bool) -> SanitizedStr<'_> {
     }
 }
 
+/// Escape control characters in an error message but keep its line breaks.
+pub fn sanitized_error(raw: &str) -> String {
+    raw.split('\n')
```
tmccombs, line 75: I think it would be better to change the existing sanitize function to have a way to exclude newlines than to do this, which will require several allocations.

tmccombs (changes requested):
Do you have a usecase where this is actually a concern? 

I'm not sure if fd should actually be concerned with treating command line arguments as untrusted. My inclination is that that is more the responsibility of the caller. Are there examples of other CLI programs that treat command line arguments this way? In particular, does ripgrep, which is a cousin of this project?  If this is something that CLI's should be concerned with, that seems like maybe it would be better for clap-rs to handle that directly in it's error messages, rather than fd working around it.

curious-rabbit:
> Do you have a usecase where this is actually a concern?
> 
> I'm not sure if fd should actually be concerned with treating command line arguments as untrusted.

I have been wavering to report this for a while questioning the same. There are some cases where I used fd in scripts that the input to fd was untrusted, though I solved that differently now.
Im not sure this is worth fixing either but I generally consider `find` to be a suitable template to look at.

`fd <eveilString>` is affected but
`find <evilString>` is not vulnerable because it doesnt use regex by default, `find -regex <evilString>` is also affected though.


> Are there examples of other CLI programs that treat command line arguments this way?

I have been looking through coreutils and many other traditional tools and the truth is that the entire topic is a mess. coreutils does a pretty decent job at addressing these issues and they did take a log of effort doing it, but there are still gaps and some cli tools in coreutils do it better then others, so even they arent consistant.

Maybe someone else knows valid use cases for this case, or other reasons why this should or should not be addressed

PR #2138, opened 2026-09-18
Add more placeholders for format

rusker86:
I have added new placeholders for "format" and included the "format-metadata" flag, which allows customizing the view with the desired placeholders.


rusker86:
Some bugs cropped up that only occurred on Windows; I had to turn to AI to fix them since I couldn't see them myself (this is my first contribution to another project, and I'm still learning).

PR #2139, opened 2026-09-21
fix: honor --exclude for broken symlinks

### LINKED ISSUE #746: Broken links don't respect `--exclude` when using `--follow`
```
$ ln -s nonexistent_file broken_link
$ fd '' .  --exclude broken_link --follow --type symlink
broken_link
```
Despite the `--exclude` argument, `broken_link` is always printed. The same thing happens when using `--ignore-file`, `.fdignore`, etc. instead of `--exclude`.

`broken_link` should not be shown in the output.

```
$ fd --version
fd 8.2.1
```

```
$ uname -srm
Linux 5.10.19-hardened1-1-hardened x86_64
$ lsb_release -a
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```

sb123sb123:
## Summary

- Apply the existing `--exclude` override matcher to broken symlink entries reported as `ignore::Error::WithPath` errors.
- Keep broken symlinks visible when they are not excluded, and add regression coverage for `--follow --type symlink --exclude`.
- Update the upcoming-release changelog.

Closes #746.

## Testing

- `rustup run nightly cargo fmt -- --check`
- `rustup run nightly cargo test --offline --locked --all-features`
- `rustup run nightly cargo test --offline --locked --test tests test_follow_broken_symlink -- --nocapture`
- `rustup run nightly cargo clippy --offline --locked --all-targets --all-features`
- `git diff --check`

The manual reproduction also passes on the Windows Git Bash host: excluding `broken_link` produces no output, while the same search without `--exclude` still finds it.

## Limitations

- The remote stable Rust installation was incomplete, so verification used the available nightly Rust 1.100.0 toolchain on Windows Git Bash.
- The strict CI clippy command with `-Dwarnings` remains blocked by two pre-existing unused test helpers in `tests/testenv/mod.rs`; the documented non-deny clippy command passes with only those baseline warnings.
- No dependencies were changed.

## AI-Assistance

An autonomous AI coding agent was used to inspect the repository, reproduce issue #746, implement the focused fix, add the regression test, and prepare this description. The resulting diff and test evidence were reviewed before submission, and the implementation is limited to reusing the existing override matcher for the broken-symlink error path.


tmccombs:
It would probably be better to fix this in the ignore crate directly.

